PR #133, opened 2013-10-24
Added support for 'sources' construct, so multiple target packages generated.

jewzaam:
The project I am working on has lots of json schemas and I wanted the generated source to land in many packages to help organize things.  This change adds support for a configuration construct called "sources" to support this.  Example configuration has been added to the test pom.xml.


joelittlejohn:
Closing as no longer needed. We now have `sourceDirectory` and `sourcePaths`, plus recursive reading of schemas when a directory is supplied.

Issue #164, opened 2014-03-24
Wrong class name when using $ref in json schema.


razem-io:
We have the following issue:

alliance.json

``` json
{
  "type": "object",
  "properties": {
    "tag": {
      "type": "string"
    },
    "name": {
      "type": "string"
    },
    "points": {
      "type": "integer"
    },
    "diplomacy": {
      "type": "object",
      "properties": {
        "treaties": {
          "type": "array",
          "items": {
               "$ref": "alliance_treaty.json"
          }
        },
        "ownOffers": {
          "type": "array",
          "items": {
               "$ref": "alliance_treaty.json"
          }
        },
        "foreignOffers": {
          "type": "array",
          "items": {
               "$ref": "alliance_treaty.json"
          }
        }
      }
    }
  }
}
```

alliance_treaty.json

``` json
{
  "type": "object",
  "properties": {
    "allianceId": {
      "type": "integer"
    },
    "otherAllianceId": {
      "type": "integer"
    },
    "type": {
      "type": "integer"
    },
    "status": {
      "type": "integer"
    } 
  }
}
```

We expect the output to be like this:
- Alliance.java
- AllianceTreaty.java

Actual result is:
- Alliance.java
- Treaty.java

This only happens when using "$ref" otherwise the class gets generated correctly. Is this a bug or are we missing something?

Thanks for your time!
Julian

dadrus:
This is related to #481. Please check the commit linked to it.

Issue #125, opened 2013-09-18
Generate inner public static classes for object properties.


dant3:
If property of type object does not have any ref or common id, I think is't better to create inner public static class Instead of package level classes. This way multiple schemas won't clash in class names if you have to create classes for object based on property name (like it works nowadays).

Example:

one.json

```
{
    "type" : "object",
    "properties" : {
        "data" : {
            "type": "object",
            "properties" : {
                "something" : {
                    "type" : "number"
                }
            }
        }
    }
}
```

two.json

```
{
    "type" : "object",
    "properties" : {
        "data" : {
            "type": "object",
            "properties" : {
                "other" : {
                    "type" : "string"
                }
            }
        }
    }
}
```

That I'm getting is two classes: Data and Data_ in a package scope. This does not makes much sence, and is impossible to use. You can use `javaType` workaround, but it's not so good either - there is not much type names you can think of.
IMHO the best solution for this problem is to generate inner public static classes inside class for parent schema, getting One.Data class and Two.Data class. This way it is readable/understandable and does not clashes.

Possibly duplicate of https://github.com/joelittlejohn/jsonschema2pojo/issues/22 as well.

zhouzhipeng:
I've implemented the "inner class " feature in my fork.   here is the pull request:  https://github.com/joelittlejohn/jsonschema2pojo/pull/1008

tviegut:
@zhouzhipeng Thanks for your efforts toward providing "inner class" support.

@joelittlejohn Any thoughts on if/when you may accept the request?

joelittlejohn:
@tviegut Thanks for the nudge. I've had a quick look through that PR and added some comments on what's left to do.

tviegut:
@joelittlejohn No problem and thanks for following up.

saberya:
@joelittlejohn is there a plan to merge this feature in the next release?  nested classes is important when i generate  a single java file from front request body.

wigbam:
Is there any update on this? Any plans to get the PR merged?

wigbam:
I have taken a shot at it:
https://github.com/joelittlejohn/jsonschema2pojo/pull/1380

Please can somebody have a look?

UnasZole:
Hi @joelittlejohn ,
Can you please review the PR #1380 that is pending to add this feature ?
It's been sitting here for more than a year, and it's definitely a very important feature.

[...]

PR #180, opened 2014-05-08
Schema Discovery

ctrimble:
The purpose of this PR is to rework the schema discovery process, so that:

1. schema names are more predictable, especially when "$ref" entries are present in the schemas.  #164
2. nested schemas can be inner classes or use the current naming structure. #125
3. the POJO generation process can be configured in a way native to the tool being used.

This PR is a work in progress.  Currently, the following changes are included:

1. Some m2e lifecycle mappings have been added, to clean up eclipse warnings.  This could be moved into another PR or removed.
2. The current schema discovery code has been moved from the CLI module to PojoGenerator in the core module.  This is a verbatim move at this point.
3. The Maven, Ant, and Gradle modules have been updated to depend on the core module.

Planned next steps:

1. Convert PojoGenerator into an instance based class with fluid api.
2. Allow clients of the generator to specify the schema discovery process.
3. Update the rules, so that generation can happen in 2 phases.

### DIFF
--- a/jsonschema2pojo-core/pom.xml
+++ b/jsonschema2pojo-core/pom.xml
@@ -30,6 +30,10 @@
             <artifactId>commons-lang</artifactId>
         </dependency>
         <dependency>
+            <groupId>commons-io</groupId>
+            <artifactId>commons-io</artifactId>
+        </dependency>
+        <dependency>
             <groupId>javax.validation</groupId>
             <artifactId>validation-api</artifactId>
         </dependency>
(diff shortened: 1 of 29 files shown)


joelittlejohn:
@ctrimble I'm going to close this pull request as I think things have moved on a lot since you were working on this branch. Clearly it's going to be hard to run major refactorings in parallel to development on master - I'd suggest that we bite off some of the smaller things in isolation if possible. It looks like you learned a good deal about where the pain points are.

ctrimble:
Totally agree.  I think that '0.5.X' may benefit from a change to the rule signatures, so that an immutable context is passed around.  That may help get the naming/ref problems resolved without causing merge conflicts while the changes go in.

PR #204, opened 2014-06-12
Adds options to generate immutable types.

jburnim:
The purpose of this PR is to add options for generating immutable Java types.  Please let me know if you think something like this would be useful to add to jsonschem2pojo.  (We're using these modifications internally at Sift Science.)

For the address.json example, the generated code with `immutable=true` and `generateBuilderClasses=true` looks roughly like:

``` java
@JsonDeserialize(builder = Address.Builder.class)
public class Address {
    @JsonProperty("post-office-box")
    private final String postOfficeBox;
    ...
    @JsonIgnore
    private final Map<String, Object> additionalProperties;

    private Address(Address.Builder builder) {
        this.postOfficeBox = builder.postOfficeBox;
        ...
        this.additionalProperties = ImmutableMap.copyOf(builder.additionalProperties);
    }

    // no setter methods

    public static Address.Builder newBuilder() {
        return new Address.Builder();
    }

    @JsonProperty("post-office-box")
    public String getPostOfficeBox() {
        return postOfficeBox;
    }
    ...
    @JsonAnyGetter
    public Map<String, Object> getAdditionalProperties() {
        return this.additionalProperties;
    }

    public final static class Builder {
        private String postOfficeBox;
        ...
        private Map<String, Object> additionalProperties = new HashMap<String, Object>();

        private Builder() { }

        public Address build() {
            return new Address(this);
        }

        @JsonProperty("post-office-box")
        public Address.Builder withPostOfficeBox(String postOfficeBox) {
            this.postOfficeBox = postOfficeBox;
            return this;
        }

        public Address.Builder withAdditionalProperties(Map<String, Object> additionalProperties) {
            this.additionalProperties = additionalProperties;
            return this;
        }

        @JsonAnySetter
        public Address.Builder withAdditionalProperty(String name, Object value) {
            this.additionalProperties.put(name, value);
            return this;
        }
    }
}
```

And generated code with `immutable=true`, `generateBuilderClasses=true`, and `usePublicFields=true` looks roughly like:

``` java
@JsonDeserialize(builder = Address.Builder.class)
public class Address {
    @JsonProperty("post-office-box")
    public final String postOfficeBox;
    ...
    @JsonIgnore
    public final Map<String, Object> additionalProperties;

    private Address(Address.Builder builder) {
        this.postOfficeBox = builder.postOfficeBox;
        ...
        this.additionalProperties = ImmutableMap.copyOf(builder.additionalProperties);
    }

    // no setter or getter methods

    public static Address.Builder newBuilder() {
        return new Address.Builder();
    }

    public final static class Builder {
        // Builder is identical to above
        ...
    }
}
```

At the moment, the options `immutable=true` and `generateBuilderClasses=false` don't work together.  (For these options, it is necessary to generate a constructor that takes one argument per property and initializes all property fields, but I haven't written code to do this.)  I can merge these options into one or maybe take a stab at generating constructors for immutable Java types.

### DIFF
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/Jackson2Annotator.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/Jackson2Annotator.java
@@ -36,9 +36,11 @@ import com.sun.codemodel.JBlock;
 import com.sun.codemodel.JClass;
 import com.sun.codemodel.JDefinedClass;
 import com.sun.codemodel.JExpr;
+import com.sun.codemodel.JExpression;
 import com.sun.codemodel.JFieldVar;
 import com.sun.codemodel.JMethod;
 import com.sun.codemodel.JMod;
+import com.sun.codemodel.JOp;
 import com.sun.codemodel.JStringLiteral;
 import com.sun.codemodel.JVar;
 
@@ -46,7 +48,7 @@ import static org.jsonschema2pojo.rules.ObjectRule.immutableCopy;
 
 /**
  * Annotates generated Java types using the Jackson 2.x mapping annotations.
- * 
+ *
  * @see <a
  *      href="https://github.com/FasterXML/jackson-annotations">https://github.com/FasterXML/jackson-annotations</a>
  */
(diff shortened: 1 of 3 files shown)


grant:
@joelittlejohn What would be the next steps to have this feature merged?

holger-s:
+1 I'd really like to use this :-)

holger-s:
So what is blocking this?

joelittlejohn:
@diversit I take your point, but the complexity and cost of your solution in Java is very large. Java doesn't have a notion of structural sharing, so doing a full, deep clone when one property changes is hugely expensive. Your short example is neat, however as the number of properties grows there's an explosion of code required to achieve the model you're proposing. I completely agree that it's a better model, but it's just impractical in Java [_hint: use Clojure for a sane approach to immutability and effortless JSON_].

It's time to get this in. I see a bunch of further changes have happened on this branch that aren't associated with the original PR - I'll pick through what's here and see how easy it is to merge only the builder parts.

EngineerBetter:
+1 for immutability

dhalbrook:
Another +1

PR #235, opened 2014-08-07
Support for @JsonTypeInfo in generated classes using deserializationClassProperty

jayashe:
Jackson 2 has a `@JsonTypeInfo` annotation available so that (de)serialization of classes that extend abstract classes works correctly. 

Adding this annotation to an abstract POJO will (during serialization) add a top-level property to the schema that specifies a fully qualified class name that specifies the type that a particular json object should be deserialized to. 

This PR allows for an "abstractJavaTypeProperty" that specifies the name of the property that will contain the fully qualified name in the serialized object. 

For example, if your schema contains node with:

```
"abstractJavaTypeProperty" : "className"
```

The generated POJO will be an abstract class, and will have this annotation:

```
@JsonTypeInfo(use = JsonTypeInfo.Id.CLASS, include = JsonTypeInfo.As.PROPERTY, property = "className")
```

### DIFF
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/ObjectRule.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/ObjectRule.java
@@ -27,10 +27,12 @@ import static org.apache.commons.lang3.ArrayUtils.*;
 import javax.annotation.Generated;
 
 import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
 
 import org.jsonschema2pojo.Schema;
 import org.jsonschema2pojo.SchemaMapper;
 import org.jsonschema2pojo.exception.ClassAlreadyExistsException;
+import org.jsonschema2pojo.AnnotationStyle;
 
 import com.sun.codemodel.JAnnotationUse;
 import com.sun.codemodel.JBlock;
@@ -45,6 +47,7 @@ import com.sun.codemodel.JMod;
 import com.sun.codemodel.JPackage;
 import com.sun.codemodel.JType;
 import com.sun.codemodel.JVar;
+import com.sun.codemodel.ClassType;
 
 /**
  * Applies the generation steps required for schemas of type "object".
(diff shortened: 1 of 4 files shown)


MonkeyLeeT:
This seems a very good feature, yet I could not see it being mentioned in the release logs (using latest version). I'm working with a POJO A that contains POJO B as its field and B is abstract (or having multiple child), so that I wish to deserialize A using different children class definition of B in run time.

Is there any ETA for actual merges of this? Thanks!

joelittlejohn:
@MonkeyLeeT Yes, this _has_ been merged and is present in 0.4.6 and later releases. Thanks for mentioning this though as the issue hadn't been tagged correctly and hence didn't show in the release notes.

MonkeyLeeT:
@joelittlejohn thanks for the update. I did try it out yesterday yet I could not get the abstract keyword added to POJO. I'm gonna try it again.

What I'm trying to achieve is to be able to define like 20 fields in one base json, then make some of them required in different json by referencing the base one.

I tried to use ref or extends to base one and add required list in the new one, but to no avail since fields are not brought to the new class, only the builder method.

Do you have some idea if this is supported in any PR? Thanks a lot!

joelittlejohn:
I'm not sure I understand what you mean. The the fields _are_ inherited by the children since the child classes inherit getters & setters from the parent. There's no need for the fields to be 'brought to the new class' (?).

What are you expecting `required` to actually _do_ to the child classes?

The builders are brought into the children because Java allows covariant return, hence the child can return a more narrow type than the parent (the child type instead of the parent type). This makes the builder method more useful, as you can chain calls and still keep operating on the child type (see #455).

MonkeyLeeT:
Sorry for any confusions.

This is the base.json
`{
  "$schema": "http://json-schema.org/draft-04/schema",
  "type": "object",
  "properties": {
    "a": {
      "type": "integer"
    },
    "b": {
      "type": "integer"
    }
  }
}`

And this is the child.json
`{
  "$schema": "http://json-schema.org/draft-04/schema",
  "type": "object",
  "extends": {
    "$ref": "base.json"
  },
  "required": ["a"]
}`

As you can see, I wish to be able to do so for other children who might want different fields inherited to be required, yet @Notnull will only be added on the actual field declaration, which is in parent class base.json.

MonkeyLeeT:
@joelittlejohn Do you have more wisdom to share with me on these? Thanks a lot!

PR #283, opened 2015-02-03
Issue231/constructors

andrewmains12:
I was looking for the ability to generate constructors, and saw the preexisting issue for it, so I decided to take a stab at implementing them. As requested in the ticket, parameters are only added to the constructor for required fields.

In terms of implementation, I made the constructor generation part of `ObjectRule`, my reasoning being that the constructor is something affected by all fields of an object, in a manner similar to `hashCode` and `equals`. This leads to some mildly ugly inspection of the `properties` node in the schema, in order to get the required properties. If it seems cleaner, I could move the generation of constructors to `PropertyRule`, with each property adding a param to the constructor (though it's a bit weirder to decide who creates the constructor in that case).

Let me know if there ought to be more test coverage; I added an integration test which I think covers the code paths.

### DIFF
--- a/jsonschema2pojo-ant/src/main/java/org/jsonschema2pojo/ant/Jsonschema2PojoTask.java
+++ b/jsonschema2pojo-ant/src/main/java/org/jsonschema2pojo/ant/Jsonschema2PojoTask.java
@@ -107,6 +107,8 @@ public class Jsonschema2PojoTask extends Task implements GenerationConfig {
 
     private String classNameSuffix = "";
 
+    private boolean constructorsRequiredPropertiesOnly = false;
+
     /**
      * Execute this task (it's expected that all relevant setters will have been
      * called by Ant to provide task configuration <em>before</em> this method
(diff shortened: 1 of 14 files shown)


joelittlejohn:
Great, thanks! I assume for your use-case you needed the 'required properties only' type of constructor? I'm a little unsure about whether this is the most useful use-case (I think many users may prefer to include all properties in the constructor, as they don't use the 'required' rule a lot) but we can always add another config option at some point.

joelittlejohn:
Regarding the ObjectRule and properties node, this isn't a problem so don't be concerned. In jsonschema draft04 the required rule has changed. It's no longer a boolean value inside the property definition, it's now an array of property names on the object schema, so I think in future this code would have to be in the ObjectRule anyway.

andrewmains12:
>  I assume for your use-case you needed the 'required properties only' type of constructor? I'm a little unsure about whether this is the most useful use-case (I think many users may prefer to include all properties in the constructor, as they don't use the 'required' rule a lot)

Actually funnily enough, for our use-case it'd probably be better to include all properties in the constructor, so I'm totally good with making the default behavior to include all properties. I mostly went with only required properties because others who commented on the ticket seemed to prefer it that way. I can change it such that the default behavior is to include all properties in the constructor, and users can specify "constructorsRequiredFieldsOnly" (or whatever better name) to use only the required fields. Seem reasonable?

joelittlejohn:
Yes, I think that would be a better (and less surprising) default.

joelittlejohn:
If you're not so interested in the `constructorsRequiredFieldsOnly` option, don't feel the need to implement this now. I'd happily release `includeConstructors` and see how well this feature works and how people use it. We can add `constructorsRequiredFieldsOnly` at a later date.

andrewmains12:
I actually had some time last night, so I went ahead and added the flag (along with some more tests for it).

joelittlejohn:
This looks like a very thorough change. You matched the conventions of the project and included some comprehensive tests. Thanks again for taking the time to work on this :+1:

andrewmains12:
Thanks, and no problem! This was actually a lot of fun to work on; the project/object model is laid out in such a way that it was really easy to get started on. A+, would use again.

Issue #223, opened 2014-07-18
Support "javaType" for things that are not "type": "object"


dflemstr:
There is nothing that says that only `"type": "object"` things can have `"javaType"`. For example, I would want a Joda `DateTime` to be serialized as an integer (letting Jackson handle the actual conversion) but this is not possible to express right now.

joelittlejohn:
I've generalised the implementation in #287 to allow for any schema type to be overridden using `javaType` (not just integers and numbers).

Issue #41, opened 2013-06-23
Add support for primitive types via the javaType property


joelittlejohn:
_Original author: christop...@smule.com (February 24, 2012 20:57:23)_

I want to generate some pojos with 64bit integer (i.e. long) fields. I can't use type=number (java type Double) because of the loss of precision.

Refer to this json schema:

{
    &quot;type&quot;:&quot;object&quot;,
    &quot;properties&quot;:{
        &quot;prop1&quot;:{&quot;type&quot;:&quot;object&quot;, &quot;javaType&quot;:&quot;java.lang.Long&quot;},
        &quot;prop2&quot;:{&quot;type&quot;:&quot;object&quot;, &quot;javaType&quot;:&quot;long&quot;},
        &quot;prop3&quot;:{&quot;type&quot;:&quot;integer&quot;, &quot;javaType&quot;:&quot;java.lang.Long&quot;},
        &quot;prop4&quot;:{&quot;type&quot;:&quot;integer&quot;, &quot;javaType&quot;:&quot;long&quot;}
    }
}

prop1 works as expected (but is limited to non-primitive).

prop2 works partly. The generated pojo is as expected, but a &quot;long.java&quot; class is also generated that I have to manually delete. This is because jsonschema2pojo uses ClassLoader.loadClass to check for existing classes, but this method doesn't work for primitives. A common way to handle primitive class lookup is to just make your own Map&lt;String,Class&gt; lookup. See http://stackoverflow.com/questions/180097/dynamically-find-the-class-that-represents-a-primitive-java-type

prop3 and prop4 don't work (the pojo has int/Integer) because javaType is only used for type:object. But I wish they would work because my field is really a (64-bit) integer, not an object.

_Original issue: http://code.google.com/p/jsonschema2pojo/issues/detail?id=41_

PR #287, opened 2015-02-16
Added support for javaType on integer and number properties.

ctrimble:
This PR adds support for the `javaType` on `integer` and `number` properties.

`integer` supports the following `javaType` values:

-  `int`
- `java.lang.Integer`
- `long`
- `java.lang.Long`

`number` supports the following `javaType` values:

- `float`
- `java.lang.Float`
- `double`
- `java.lang.Double`

When `javaType` is present, the configuration settings are ignored.

### DIFF
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/TypeRule.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/TypeRule.java
@@ -16,6 +16,8 @@
 
 package org.jsonschema2pojo.rules;
 
+import static org.jsonschema2pojo.rules.PrimitiveTypes.*;
+
 import org.jsonschema2pojo.GenerationConfig;
 import org.jsonschema2pojo.Schema;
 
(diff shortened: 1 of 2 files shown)


ctrimble:
This relates to #223 and seems to be a regression of #41.

joelittlejohn:
Ha, so it is. A bit of archeology needed :)

Thanks for submitting this! I'll give it a look over and merge tomorrow.

ctrimble:
Thanks @joelittlejohn !

joelittlejohn:
Took a quick look at #41, and no this is not a regression of that change. That change relates to being able to specify primitive types when you use the javaType property inside 'object'.

I'm interested to know what prompted you to make this change. Did you need to change the type of specific properties and not want to change the configuration globally with useLongIntegers/useDoubleNumbers/usePrimitives?

ctrimble:
@joelittlejohn I needed most of my `integer` fields to be `java.lang.Integer`, but a few required `java.lang.Long`.  We were double configuring the maven plugin to do this, but that was a lot of config to get things to work.  This seems much more clear.

My current project is polyglot and we are using JSON Schema as a way of keeping our interfaces sane.  I am starting to have a lot of schemas in my build, so fine grained control over things like is coming up more frequently.

ctrimble:
Totally willing to work on this some, if you think the implementation needs to be different.

joelittlejohn:
Hi Christian. Looks great to me. Lots of tests too so I'm perfectly happy to merge.

To be perfectly honest, I probably would have left this as an unconstrained option and simply used whatever type is found in javaType. E.g. If I want to use BigInteger/BigDecimal I could use that, or maybe my own custom type (as long as I give Jackson a serializer/deserializer for my custom type, it'll work). I'm happy to go with these constraints though as they make sense in most cases. If anyone feels limited by them, we can relax it in a later version.

ctrimble:
Do you already have a method around that operates like Class.forName(String), but handles primitives like `int`, `long`, etc.?  If so, I will take a moment and update the PR.

joelittlejohn:
This class may help:

https://github.com/joelittlejohn/jsonschema2pojo/blob/master/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/PrimitiveTypes.java

You could use a check like `if (isPrimitive(javaType)) {...} else {owner.ref(javaType)}` or maybe `try { return primitiveType(javaType); } catch (GenerationException e) {owner.ref(javaType)}`. Or something else along those lines :)

ctrimble:
Cool.  Let me take a look at this real quick and add a commit for it.  If it looks good, we can squash.  If it sucks, we can drop the commit.  Would like this to be done right.

ctrimble:
@joelittlejohn the implementation is generalized.  Let me know what you think and I can clean up the history.

joelittlejohn:
Looks good. Squash awaysony and I'll merge tomorrow. Thanks again for all your contributions!

ctrimble:
Squashed and ready to merge.

PR #292, opened 2015-02-20
Android support in gradle plugin (fix issue 266)

ben-manes:
Previously the `java` plugin was automatically applied when configuring the task dependencies. This was not compatible with the `android` plugin. Now this plugin will inspect the build to determine the right configuration to target.

Added the example provided by @samskiter that demonstrated the bug and now shows a working sample.


joelittlejohn:
@ben-manes feel free to merge this when you're ready.

samskiter:
Hi @ben-manes  I checkout out the repo and opened the android project up in android studio and still got the same error. I assume that's because the jsonschema2pojo dependency is still being pulled in from maven, so I'll have to trust you that this fixes it...

ben-manes:
@samskiter, you have to perform `mvn install` first, which would make the dependency available in a local repository.

joelittlejohn:
Thanks @ben-manes and @samskiter!

Issue #298, opened 2015-03-03
Add config options to generate Joda LocalDate and LocalTime


brentsowers1:
There is a really useful feature currently in place where the generated class property will be LocalDateTime when the JSON schema type is string with format of date-time (implemented in https://github.com/joelittlejohn/jsonschema2pojo/blob/a0797e0878dc46f174b64643bc09655b22983afb/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/FormatRule.java).  I'd like to propose generating a LocalDate when the format is date, and LocalTime when the format is time.  I see date and time listed in that file explicitly being set to a type of string, so I am guessing that this was thought about at at least some point?

If you think this is a good idea I will take care of implementing it and issuing a pull request, I just wanted to propose it here to make sure it makes sense before starting work.

For the implementation, I would follow the pattern that has been implemented for date-time. I'd add a new  config flag for whether or not to use Joda LocalDate and LocalTime (separate from the current flag for whether to use Joda DateTime), defaulting to false.

joelittlejohn:
Hi Brent. This sounds like a good addition, thanks for raising this. Happy to accept a pull request along the lines you have suggested (useJodaLocalDates).

I assume the Joda serializer for Jackson can support LocalDate and LocalDateTime?

joelittlejohn:
Closed by #300

PR #300, opened 2015-03-06
Added options to generate Joda LocalDate and LocalTime for date and time formatted strings

brentsowers1:
https://github.com/joelittlejohn/jsonschema2pojo/issues/298

I'm pretty sure I covered everything here. I added 2 new config flags - useJodaLocalDates and useJodaLocalTimes. If those are set to true, and the schema is type string with format date or format time, the generated class properties will have types of LocalDate and LocalTime, respectively.

I've run the integration tests, and manually packaged a test version, using the maven plugin in my application to use this version, and verified that the class gets generated correctly.


joelittlejohn:
Thanks!! I'll take a look through and we'll get this merged for 0.4.9.

Issue #305, opened 2015-03-11
Add support for android library plugin


yomik:
0.4.8 added support for Android gradle plugin (com.android.application).

However, it lacks support for Android libraries (com.android.library). It would be nice to have support for these as well.

joelittlejohn:
Could you elaborate a bit on what you can and can't do with the current android support? I don't fully understand why this plugin needs to explicitly support these libraries. Can you not simply add this dependency yourself? Or do you need this plugin to recognise com.android.library as something that should be treated as an android plugin?

Hopefully @ben-manes and @samskiter could offer an opinion on this too :)

samskiter:
@joelittlejohn this is probably the same basic issue as https://github.com/joelittlejohn/jsonschema2pojo/issues/266

@yomik i guess you are receiving an error about the java plugin being incompatible with the android plugin.

I imagine that this should be an easy fix for someone who knows more about the system.

Looks as though it would just require an addition here:

https://github.com/joelittlejohn/jsonschema2pojo/blob/master/jsonschema2pojo-gradle-plugin/src/main/groovy/org/jsonschema2pojo/gradle/GenerateJsonSchemaTask.groovy#L44

joelittlejohn:
Is this something that every Gradle plugin has to go through? Something
doesn't feel right about this but I don't know enough Gradle projects to
understand this :)

samskiter:
im probably less familiar. are there any other ways to integrate other than as a plugin?

ben-manes:
I didn't realize that the Android stack has two plugins (library and application). As Sam said, its probably as simple as adding an OR condition to inspect for that plugin as well. I'd like to have a sample project like last time to verify with.

Most Gradle plugins don't have this complexity with Android as they don't need to integrate into a compilation workflow. This is one of the few cases where Maven's phase model is easier than a task graph, because the code generation would always occur prior to the compiler plugin's execution. In Gradle, we need to setup the dependencies which due to Android's non-standard workflow adds complexity. We could of course make our users configure the DAG themselves, but that would annoy them and spam you with a lot of basic setup questions.

If @yomik provided a sample project, perhaps by extending @samskiter's, this should be an easy task.

joelittlejohn:
Fantastic, thanks for the detailed explanation Ben.

yomik:
Sorry, I should have given more details.

[...]

PR #311, opened 2015-03-16
#305 : Add android lib sample + Add support for android library plugin

yomik:
(Untested)

- Android gradle plugin comes into two flavors : com.android.application for normal applications, and com.android.library for libraries. Added support for the latter.


joelittlejohn:
Thanks!

Issue #321, opened 2015-03-20
$ref doesn't support local definitions


tylerprete:
I have this file (saved as entityType.json):
```
{
  "$schema": "http://json-schema.org/draft-04/schema#",
  "$ref": "#/definitions/entityType",
  "id": "resource:///json_schema/match/entityType.json#",
  "definitions": {
    "entityType": {
      "enum": [ "LEAD", "CONTACT", "ACCOUNT", "OPPORTUNITY", "ROW" ]
    }
  }
}
```
This is valid jsonschema, and a fairly common pattern, but jsonschema2pojo fails compiling this file, with the error below. I have a change that solved this for my use-cases, though I'm not sure it's entirely safe. I'll link to it (https://github.com/joelittlejohn/jsonschema2pojo/pull/322).

```
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 0.816 s
[INFO] Finished at: 2015-03-19T22:19:36-07:00
[INFO] Final Memory: 7M/245M
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.jsonschema2pojo:jsonschema2pojo-maven-plugin:0.4.10:generate (default) on project schema: Execution default of goal org.jsonschema2pojo:jsonschema2pojo-maven-plugin:0.4.10:generate failed: Path not present: definitions -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.jsonschema2pojo:jsonschema2pojo-maven-plugin:0.4.10:generate (default) on project schema: Execution default of goal org.jsonschema2pojo:jsonschema2pojo-maven-plugin:0.4.10:generate failed: Path not present: definitions
	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:224)
	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:153)
	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:145)
	at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject(LifecycleModuleBuilder.java:116)
	at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject(LifecycleModuleBuilder.java:80)
[...]

PR #322, opened 2015-03-20
Use refschema rather than parent to resolvRef's in PropertyRule.

tylerprete:
This change resolves an issue I've been having locally, where using a $ref to local definitions does not work.

The issue: https://github.com/joelittlejohn/jsonschema2pojo/issues/321

### DIFF
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/PropertyRule.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/PropertyRule.java
@@ -129,7 +129,7 @@ public class PropertyRule implements Rule<JDefinedClass, JDefinedClass> {
         if (node.has("$ref")) {
             Schema refSchema = ruleFactory.getSchemaStore().create(parent, node.get("$ref").asText());
             JsonNode refNode = refSchema.getContent();
-            return resolveRefs(refNode, parent);
+            return resolveRefs(refNode, refSchema);
         } else {
             return node;
         }


joelittlejohn:
Hi Tyler, thanks for submitting this. I'm a little confused about why this would be failing as we have an integration test to cover this feature (ref to a fragment within the same schema):

https://github.com/joelittlejohn/jsonschema2pojo/blob/jsonschema2pojo-0.4.10/jsonschema2pojo-integration-tests/src/test/java/org/jsonschema2pojo/integration/ref/FragmentRefIT.java#L49

which uses this schema:

https://github.com/joelittlejohn/jsonschema2pojo/blob/jsonschema2pojo-0.4.10/jsonschema2pojo-integration-tests/src/test/resources/schema/ref/fragmentRefs.json

What's even more strange though is that this test passes both before your change and after, so something is not right here :) I'm wondering if this is a regression caused by changes in 0.4.9 to the way that files are resolved. 

Could you try this on some earlier versions (like 0.4.7 and 0.4.0) and tell me if you see the same error?

joelittlejohn:
This seems to work correctly at jsonschema2pojo.org (for instance) which is using 0.4.6, so I think may be a recent regression. Admittedly jsonschema2pojo.org uses a completely different way of reading files, since they're in memory only.

tylerprete:
Interesting. the website does generate it just fine. I'll give some other versions a try and let you know.

tylerprete:
So I made a small project with just that file, tried several versions of jsonschema2pojo, and all of them are working. I'm thinking it's an artifact of having many json files in different directories. I'll slowly keep adding files from my full project to my test one to see if I can figure out where it breaks.

tylerprete:
So it turns out this was actually user error :(.

In a different file, I was referencing entityType with just:
```
"$ref": "entityType.json#"
```

rather than:
```
"$ref": "entityType.json#/definitions/entityType"
```

I'll go ahead and close this.

joelittlejohn:
Great. Thanks for reporting back!

tylerprete:
Actually, I think this is still a bug.

As entityType has a "$ref" top-level pointing at it's own definitions, other files wanting to reference entityType should only need to use "$ref": "entityType.json#", right?

They shouldn't need to dig down into definitions.

tylerprete:
Check out this project:
https://github.com/tylerprete/jsonschema-test

It's the minimal setup to cause the failure.

Issue #323, opened 2015-03-20
Feature Request: Support required array, vs required on each property


tylerprete:
In json_schema v4, the idiomatic way to do required properties is with a required array at the end of the object:
```
{
  "$schema": "http://json-schema.org/draft-04/schema#",
  "$ref": "#/definitions/abc",
  "id": "resource:///json_schema/match/abc.json#",
  "definitions": {
    "abc": {
      "type": "object",
      "properties": {
        "name": {
          "type": "string"
        },
        "age": {
          "type": "string"
        }
      },
      "required": ["name", "age"]
    }
  }
}
```

jsonschema2pojo supports required on each property, but ignores the required array at the end.

We have a lot of files using this, so we don't really want to refactor them all. I'd be happy to try my hand at it, especially if you could point me in the right direction.

PR #325, opened 2015-03-23
Adding support for required array from http://tools.ietf.org/html/draft-...

dmcfarland:
Adding support for required array from http://tools.ietf.org/html/draft-fge-json-schema-validation-00#section-5.4.3

As described under https://github.com/joelittlejohn/jsonschema2pojo/issues/323 

In json_schema v4, the idiomatic way to do required properties is with a required array at the end of the object:

```javascript
{
  "$ref": "#/definitions/abc",
  "id": "resource:///json_schema/match/abc.json#",
  "definitions": {
    "abc": {
      "type": "object",
      "properties": {
        "name": {
          "type": "string"
        },
        "age": {
          "type": "string"
        }
      },
      "required": ["name", "age"]
    }
  }
}
```

### DIFF
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/NameHelper.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/NameHelper.java
@@ -20,6 +20,7 @@ import static java.lang.Character.*;
 import static javax.lang.model.SourceVersion.isKeyword;
 import static org.apache.commons.lang3.StringUtils.*;
 
+import com.sun.codemodel.JType;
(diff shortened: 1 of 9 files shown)


joelittlejohn:
Hi David. Thanks a lot for tackling this, a lot of people will find it useful. 

The one thing we need before I can merge this is an integration test to cover this feature. Hopefully you can use the current `RequiredIT` for an example of how to write these. There are helpers to generate and compile code during the test so they're quite straight forward to write.

dmcfarland:
Thanks for pointing out the missing integration test. Hopefully should be committed now.

tylerprete:
@dmcfarland Thanks for implementing this!

Issue #331, opened 2015-04-01
Using extends breaks equals hashcode


brentryan:
If you use the extends keyword in your json schema and you use equals() or hashcode() on your subtype then it can mean that equals() will evaluate to true even though your parent object might have different values in your properties.

I think we just need to add a fix to call super.equals() super.hashcode() to evaluate this all the way up the chain.

joelittlejohn:
Closed by #333

PR #333, opened 2015-04-08
Add super() to the equals/hashcode impl

brentryan:
This is to address this issue: https://github.com/joelittlejohn/jsonschema2pojo/issues/331


brentryan:
@joelittlejohn Do you see any issues with being able to merge in this PR?  I'd rather avoid having to maintain a fork for this.

joelittlejohn:
Hi there. I haven't had much time to dedicate to the project recently. I have no objection to the changes in this PR, I just haven't had a chance to review.

I've held of merging this immediately because I want to check back through the assumptions made when #241 was implemented and understand _exactly_ what the trade-off is between calling super or not here. I believe #241 was implemented with the recommendations of Effective Java in mind, and I want to be sure that we're not creating another trap here by calling super.

I think your [reasoning is sound](http://www.angelikalanger.com/Articles/JavaSolutions/SecretsOfEquals/Equals.html), I just wanted to verify this assumption before merging.

brentryan:
Awesome! Thanks!

PR #338, opened 2015-04-18
Added the feature to generate addXXX method for array

GuuJiang:
For the array field, generate a addXXX method which make user to construct the object more fluent with the builder pattern.


joelittlejohn:
Hi there. Very nice addition, thank you!

Before I merge this could you add an integration test please? We need these
tests to make sure this feature isn't lost or broken in a future release.
Also your indentation width seems to be different than the rest of the
project, could you make sure this is 4 spaces?

Cheers

joelittlejohn:
Also, it looks like this change has caused some tests to fail. Could you take a quick look at:

https://travis-ci.org/joelittlejohn/jsonschema2pojo/jobs/58994293

Thanks!

To run the integration tests yourself, just use `mvn clean verify`.

joelittlejohn:
I'm going to close this pull request as it's not in a mergeable state. If you are able to fix the broken tests and rebase against master then I can merge this. Thanks.

Issue #332, opened 2015-04-08
Exclude/Includes breaks recursive folder reading


balinju:
I have the following configuration:

```
<groupId>org.jsonschema2pojo</groupId>
<artifactId>jsonschema2pojo-maven-plugin</artifactId>
<version>0.4.10</version>
<configuration>
	<sourceDirectory>${basedir}/src/main/api/schemas</sourceDirectory>
	<targetPackage>com.xxx.domain</targetPackage>
	<outputDirectory>${project.build.directory}/generated/jsonschema</outputDirectory>
	<removeOldOutput>true</removeOldOutput>
	<annotationStyle>jackson1</annotationStyle>
</configuration>
```

This breaks on Linux due to the .svn folders

```
[ERROR] Failed to execute goal org.jsonschema2pojo:jsonschema2pojo-maven-plugin:0.4.10:generate (default) on project p1: Execution default of goal org.jsonschema2pojo:jsonschema2pojo-maven-plugin:0.4.10:generate failed: Error parsing document: file:/opt2/jenkins-data/jobs/Build-p1/workspace/src/main/api/schemas/.svn/all-wcprops: Unexpected character ('K' (code 75)): expected a valid value (number, String, array, object, 'true', 'false' or 'null')
```

Temporary removing .svn folder to test works brilliantly.

Adding Exclude/Include to only process *.json as shown below uncovers another issue

```
<includes>
	<include>*.json</include> 
</includes>
```

When the above is added, only the schemas at the top level of the directory are read. The directory is not recursively read. This was working without the include. I also tried the below, but did not work either.

```
<includes>
	<include>**/*.json</include> 
</includes>
```

balinju:
Hi,
I have checked this issue out and I have discovered the source of the problem. When no includes/excludes are configured, the AllFileFilter is used which returns true on every check. However when include/excludes are configured, the MatchPatternsFileFilter is used instead. The problem here is with the directory check against include/excludes.

In the accept() method, !excludePatterns.matchesPatternStart(path, caseSensitive) is always returning false if the exclude list contains at least one item. By default the createFilter() method adds the default exclude list so this list is never empty.

This might be a misuse of the method matchesPatternStart().

joelittlejohn:
[...]

PR #340, opened 2015-04-27
Corrected Exclude Pattern Behavior with Maven.

ctrimble:
The default set of exclude directories was causing all files to be excluded.  This corrects that issue and adds tests related to the default exclude list.  This is related to #332.

This PR also adds the default set of maven excludes anytime the sourceDirectory property is supplied without the sourcePaths property.


ctrimble:
@joelittlejohn I have finished my changes for this PR and it is ready for review.  I left this as 2 commits, since there are 2 distinct changes coming in.

PR #347, opened 2015-05-06
Allow different enum name mapping for EnumRule

masgari:
Allow different constant name mapping by moving `getConstantName()` to another interface.

Default `EnumRule` constructor is using `SplitCamelCaseNameMapper`. 
Current behaviour has not changed.

### DIFF
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/EnumRule.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/EnumRule.java
@@ -54,7 +54,7 @@ import com.sun.codemodel.JVar;
 
 /**
  * Applies the "enum" schema rule.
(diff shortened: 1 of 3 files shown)


joelittlejohn:
Hi, I must admit this seems a little complicated for what you need to do here. I assume you want to create your own mapper, but couldn't you do this by simply making `EnumRule#getConstantName` protected instead of private? Then you could extend EnumRule and provide your own implementation of this method.

What do you think?

masgari:
I think `EnumRule` is doing multiple jobs, so I moved `getConstantName()` to another class, but it can be done by making it protected, as you said.

joelittlejohn:
I think you're right about it doing multiple jobs, but the complexity of introducing new types to achieve what you need here seems a little over complicated. I'll change the accessor on this method.

PR #348, opened 2015-05-07
Allow empty package names

ctrimble:
This PR corrects an issue where the `targetPackage` is empty and source files were being generated with a `package .;` statement.

This update has the following known limitation:  If the source directory contains a directory that is a java keyword, like `/default`, then the generated source will be invalid.  Not sure if handling key words should be part of this PR or another.


joelittlejohn:
:+1:

PR #349, opened 2015-05-08
Multiple enum array properties with same name cause JClassAlreadyExistsException

tmkarthi:
The following Json causes JClassAlreadyExistsException
```javascript
{
    "type": "object",
    "properties": {
        "foo": {
            "type": "object",
            "properties": {
                "statuses": {
                    "type": "array",
                    "items": {
                        "type": "string",
                        "enum": [
                            "open",
                            "closed"
                        ]
                    }
                }
            }
        },
        "bar": {
            "type": "object",
            "properties": {
                "statuses": {
                    "type": "array",
                    "items": {
                        "type": "string",
                        "enum": [
                            "open",
                            "closed"
                        ]
                    }
                }
            }
        }
    }
}
```
In this case, as the enum item is inside arrays, the enum class is generated at the package level. If there are multiple enum properties with same name, there is a name collision. Changed the EnumRule code to append _ to the Enum name, if a class by that name is already found.

### DIFF
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/EnumRule.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/EnumRule.java
@@ -28,29 +28,13 @@ import java.util.Map;
 
 import javax.annotation.Generated;
 
+import com.sun.codemodel.*;
 import org.jsonschema2pojo.Schema;
 import org.jsonschema2pojo.SchemaMapper;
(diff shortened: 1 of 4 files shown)


joelittlejohn:
Thanks for providing this pull request. Before I can merge however we need an integration test that will demonstrate that the fix is working. This is mainly to avoid a regression that breaks or loses this behaviour in future. Hopefully the existing enum integration tests are easy to understand.

tmkarthi:
Thanks for the feedback. I've added a test case for verifying the fix. Please let me know, if something more needs to be done.

PR #371, opened 2015-06-05
add string util class

simpleton:
follow your opinion `Can we use StringUtils method for this?`


joelittlejohn:
I'm really sorry I wasn't clear, I meant using `org.apache.commons.lang3.StringUtils` instead of implementing this yourself.

joelittlejohn:
I'm going to close this without merging, we can look at using org.apache.commons.lang3.StringUtils to replace some of the string parsing logic.

Issue #377, opened 2015-06-22
RequiredArrayRule not working for properties with _ or -.


josseyj:
RequiredArrayRule is not transforming the field name, and so if the field name has underscores in it, it wont lookup the right field.
https://github.com/joelittlejohn/jsonschema2pojo/blob/61ce3bfd16af27dac8e4ebd814670c435f64fe94/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/RequiredArrayRule.java#L51-L52

josseyj:
Sample Request
```json
{
  "type":"object",
  "properties": {
    "foo_bar": {
      "type": "string"
    },
    "foobar": {
      "type": "string"
    }
  },
  "required": [
     "foo_bar",
     "foobar"
    ]
}
```

Output:
```java
public class Test {

    @JsonProperty("foo_bar")
    private String fooBar;

    /**
     * (Required)
     */
    @JsonProperty("foobar")
    @NotNull
    private String foobar;

    @JsonIgnore
    private Map<String, Object> additionalProperties = new HashMap<String, Object>();

}
```
`@NotNull` not placed on `fooBar` property.

nstoddar:
Submitted PR for this.
https://github.com/joelittlejohn/jsonschema2pojo/pull/378

nstoddar:
When's the next release supposed to be?  Anything stopping a 0.4.13 release now?

joelittlejohn:
@nstoddar sounds like you really need that annotation :)

I've just cut 0.4.13, it should appear in maven central in 4 hours.

PR #378, opened 2015-06-22
fix for properties with underscore not having @NotNull applied if pro…

nstoddar:
This is a fix for https://github.com/joelittlejohn/jsonschema2pojo/issues/377 -- which we discovered today.  Any chance this can get accepted in the next couple of minutes?  :)


joelittlejohn:
Thanks :+1:

PR #379, opened 2015-06-23
Corrected behaviour of constructorsRequiredPropertiesOnly

UrbaneChimp:
When the constructorsRequiredPropertiesOnly configuration option is true, the "required" array in an object's schema is now also consulted when determining the set of constructor properties.


joelittlejohn:
Thanks for this. We'll need an integration test to make sure this remains
fixed in future. Could you add one?

Cheers

PR #380, opened 2015-06-25
Add @JsonFormat annotation if date-pattern specified

### LINKED ISSUE #247: Property with type string and format date-time should output ISO8601 Date instead of Timestamp
If we have a property of type string with format date-time defined like this

``` json
"properties": {
        "generatedAt": {
            "id": "http://jsonschema/generatedAt#",
            "type": "string",
            "format": "date-time",
            "minLength": 1
        },
```

the code generated for that property will be

``` java
@JsonProperty("generatedAt")
private Date generatedAt;
```

When the property gets serialised, the value produced is a timestamp in milliseconds..(default Jackson behaviour)
Since the type specified is a string with a date-time format I would expect this to produce an ISO8601 Date string

This can be solved by annotating the generated property with:

``` java
@JsonProperty("generatedAt")
@JsonFormat(shape=JsonFormat.Shape.STRING, pattern="yyyy-MM-dd'T'HH:mm:ss'Z'", timezone="GMT")
private Date generatedAt;
```

tszpinda:
resolves #247

### DIFF
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/JsonFormatRule.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/JsonFormatRule.java
@@ -1,0 +1,43 @@
+/**
+ * Copyright © 2010-2014 Nokia
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jsonschema2pojo.rules;
+
+import org.jsonschema2pojo.Schema;
+
+import com.fasterxml.jackson.annotation.JsonFormat;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.sun.codemodel.JAnnotationUse;
+import com.sun.codemodel.JFieldVar;
+
+public class JsonFormatRule implements Rule<JFieldVar, JFieldVar> {
+
+    protected JsonFormatRule(RuleFactory ruleFactory) {
+    }
+
+    @Override
+    public JFieldVar apply(String nodeName, JsonNode node, JFieldVar field, Schema currentSchema) {
+
+        if (node.has("date-pattern")) {
+            JAnnotationUse annotation = field.annotate(JsonFormat.class);
+            annotation.param("pattern", node.get("date-pattern").asText());
+            annotation.param("shape", JsonFormat.Shape.STRING);
+        }
+
+        return field;
+    }
+
+}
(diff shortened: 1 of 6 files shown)


tszpinda:
Hi do you know why java 7 branch would fail with:

Tests in error: 
  org.jsonschema2pojo.integration.ref.HttpRefIT: Unrecognised URI, can't resolve this: https://jsonschema2pojo.googlecode.com/git-history/jsonschema2pojo-0.3.4/jsonschema2pojo-integration-tests/src/test/resources/schema/ref/address.json

?

joelittlejohn:
I think there was simply a temporary connectivity issue that stopped the schema being downloaded (e.g. a timeout). I kicked off the build again and it's successful :)

tszpinda:
@joelittlejohn have you got any plans to merge this one?

joelittlejohn:
Hi, apologies for being slow on this. I haven't had much time to dedicate to the project recently.

I must admit I'm a little concerned about merging this at it feels like this new property that you have conceived (`date-pattern`) conflicts with what JSON Schema has to say about date formats. You're choosing a bespoke format when in fact the spec says this in draft03:

```
date-time  This SHOULD be a date in ISO 8601 format of YYYY-MM-
      DDThh:mm:ssZ in UTC time.  This is the recommended form of date/
      timestamp.

   date  This SHOULD be a date in the format of YYYY-MM-DD.  It is
      recommended that you use the "date-time" format instead of "date"
      unless you need to transfer only the date part.
```

and this in the newer validation spec (draft04+):

```
7.3.1.  date-time

7.3.1.1.  Applicability

   This attribute applies to string instances.

7.3.1.2.  Validation

   A string instance is valid against this attribute if it is a valid
   date representation as defined by RFC 3339, section 5.6 [RFC3339].
```

Are you currently validating the data with your schema or simply using the schema with this library to generate code?

I guess I'd be happy to merge this with a few changes:
* Could you rename `date-pattern` to `customDatePattern`. The convention for these rules is lower camel case
* Could you apply this in the annotator instead of in a new rule, you can call the annotator from within org.jsonschema2pojo.rules.FormatRule (if the format is `date` or `date-time`) and remove JsonFormatRule entirely.

OverlyExcessive:
@joelittlejohn Any chance that you could complete the outstanding tasks on this PR if the OP is no longer maintaining it? It would be a very nice thing to have, as it stands right now the out-of-box behaviour for POJOs serialised using Jackson will *not* validate with the supplied schema if they have `date-time` properties.

joelittlejohn:
Hi @OverlyExcessive. Is this something that would be useful for you on your projects?

OverlyExcessive:
@joelittlejohn Yes it would be very useful for me, and many others I'm sure.

joelittlejohn:
@OverlyExcessive Is it something it would be worth your while to pick up?

OverlyExcessive:
@joelittlejohn Yes possibly but I would like to change it so it better conforms to the specs in that case. From what I can gather there is no `date-pattern` field, rather specifying the `format` property to of value `date-time` means that the string value must be a valid ISO-8601/RFC3339 date time. So if I were to pick this up I would like it to work based on that instead.

Would it be best if I created a separate PR for this?

joelittlejohn:
Yes, I'd definitely prefer that you take the route of using the spec features rather than a new custom rule.

Feel free to create a fresh PR and we will close this one.

OverlyExcessive:
@joelittlejohn Are you sure that it would be possible to call the annotator directly from within `org.jsonschema2pojo.rules.FormatRule`? It seems to me that `FormatRule` is a `TypeRule` and is only passed a reference to a `JType` not a `JFieldVar`.

shravanspurohit:
I needed the same feature and I started working on this a while ago and didn't check the progress on this thread. I have implemented the feature based on the initial feedback from @joelittlejohn 
Here is the PR: https://github.com/joelittlejohn/jsonschema2pojo/pull/643

joelittlejohn:
Superseded by #643.

PR #386, opened 2015-07-06
Add Dynamic Accessors

ctrimble:
This PR adds methods for getting and setting properties by name, either in one of the declared fields or in the additional properties map.  This code generates getter methods like this:

```
protected final static Object NOT_FOUND_VALUE = new Object();

protected Object declaredPropertyOrNotFound(String name, Object notFoundValue) {
    switch (name) {
        case "someField":
            return getSomeField();
        // more cases here for fields on this schema.
        default:
            // call to super.getField$(name, notFoundValue) if extending another schema.
            return notFoundValue;
    }
}

public<T >T get(String name) {
    Object value = declaredPropertyOrNotFound(name, NOT_FOUND_VALUE);
    if (NOT_FOUND_VALUE!= value) {
        return ((T) value);
    } else {
        // IllegalArgumentException here if no additional properties.
        return ((T) getAdditionalProperties().get(name));
    }
}
```

and setter methods like this:

```
protected boolean declaredProperty(String name, Object value) {
    switch (name) {
        case "someField":
            if (value instanceof String) {
                setSomeField(((String) value));
            } else {
                throw new IllegalArgumentException(("property \"someField\" is of type \"java.lang.String\", but got "+ value.getClass().toString()));
            }
            return true;
        // more cases here for fields in this schema.
        default:
            // call to super.setField$(name, value) if extending another schema.
            return false; 
    }
}
public void set(String name, Object value) {
   if (!declaredProperty(name, value)) {
        // IllegalArgumentException here if no additional properties.
        getAdditionalProperties().put(name, ((Object) value));
    }
}
```

when `generateBuilders` == true, then a builder method is also generated, with the signature

```
public THIS_TYPE with( String property, Object value )
```

when the `target` == `1.6`, `if/elseif` blocks are used, instead of String switch statements.

This feature can be turned off by setting `includeDynamicAccessors` to false.


jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/ObjectRule.java
```diff
@@ -325,6 +327,7 @@ private void addHashCode(JDefinedClass jclass) {
         }
         
         for (JFieldVar fieldVar : fields.values()) {
+            if( (fieldVar.mods().getValue() & JMod.STATIC) == JMod.STATIC) continue;
```
ctrimble: This change, and the others related to filtering out static fields, should probably come in as a separate PR, since this is technically broken on the 0.4.X line of development.

pom.xml
```diff
@@ -161,8 +161,8 @@
                 <version>2.3.2</version>
                 <configuration>
                     <encoding>UTF-8</encoding>
-                    <source>1.6</source>
-                    <target>1.6</target>
+                    <source>1.7</source>
+                    <target>1.7</target>
```
ctrimble: This change to the compiler target isn't really needed for the plugin itself, but the code generated does require 1.7, since String based switch statements are being employed.

ctrimble:
My thoughts on this change:

1. Not sure about the naming.  This will conflict with the keys `$` and `field$`, so that is not great.
2. This adds a new linkage with the parent class when using `extends`.  The alternative isn't great either, where `$set` and `$get` call ancestor setter and getter methods directly.
3. Moving to 1.7 for this change isn't great either.  This could just be an if/else if construction for now.

ronnyroeller:
+1

ctrimble:
@cminardi I had not planned on getting into pathing with this change, since that would probably lead to `jsonpath` support.  That seems outside the bounds for this project and would probably be better facilitated by an external tool.  It looks like https://github.com/nebhale/JsonPath might do what you want.

ctrimble:
@joelittlejohn Is the intention of `includeAccessors` to not generate any methods at all and only output fields?  If so, it seems like this PR should respect that flag.

joelittlejohn:
Yes exactly that, public fields.

ctrimble:
Great.  I will add in checks for that flag and some more testing, since I am generating methods now and the test suite is passing.

ctrimble:
The `$` character has been removed from the setter and getter methods.

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/PropertiesByNameRule.java
```diff
@@ -0,0 +1,367 @@
...
+import static org.jsonschema2pojo.util.LanguageFeatures.*;
+
+public class PropertiesByNameRule implements Rule<JDefinedClass, JDefinedClass> {
+    
+    public static final String NOT_FOUND_VALUE_FIELD = "NOT_FOUND_VALUE";
+    public static final String SETTER_NAME = "set";
+    public static final String GETTER_NAME = "get";
+    public static final String BUILDER_NAME = "with";
+    public static final String DEFINED_SETTER_NAME = "declaredProperty";
+    public static final String DEFINED_GETTER_NAME = "declaredPropertyOrDefault";
```
ctrimble: This should probably be `declaredPropertyOrNotFound`.  The current name implies that the provided value is returned for null.

jsonschema2pojo-core/src/test/java/org/jsonschema2pojo/rules/PropertiesByNameRuleTest.java
```diff
@@ -0,0 +1,76 @@
...
+    public void shouldAddNotFoundField() {
+        JFieldRef var = rule.getOrAddNotFoundVar(type);
+        assertThat(var, notNullValue());
+    }
+    
+    @Test
+    public void shouldReferenceNotFoundFieldInParent() {
+        JFieldRef parentRef = rule.getOrAddNotFoundVar(type2);
+        type._extends(type2);
+        JFieldRef ref = rule.getOrAddNotFoundVar(type);
```
ctrimble: This test doesn't assert anything.

ctrimble:
Once #454 is on master, this change this change should be rebased and squashed.  Otherwise, it is complete.

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/ObjectRule.java
```diff
@@ -337,6 +339,7 @@ private void addHashCode(JDefinedClass jclass) {
         }
 
         for (JFieldVar fieldVar : fields.values()) {
+            if( (fieldVar.mods().getValue() & JMod.STATIC) == JMod.STATIC) continue;
```
ctrimble: When squashing this PR, bring all of the changes like this into a single commit, before the actual feature.

joelittlejohn:
Nice work, thanks for continuing with this.

Is 1.6 support something you need for your own project @ctrimble? It's nearly three years since public updates stopped for that version of Java so I'm not sure we still need to support it. I'll leave that up to you though (it seems to add a lot of code and complexity to this PR). The population of users that use this feature, and the population of users that are still on Java 6 are two sets that may never intersect :)

The last thing we need to here is put this behind a config flag. People can get very particular about the code generated and there's now a lot added when this feature is active.

ctrimble:
@joelittlejohn I didn't know if the android people were all ready to move up to 1.7, so I left in support.  If you are comfortable with moving the version up, then perhaps we should just drop the 1.6 support, so the project doesn't have to lug it around.

In either case, I would like to get the `target` version exposed to the rules and test suite.  I could really use some Java8 features and I know the android programmers are not ready to make that move.

For instance, methods like this would be very handy for me:

```
  public T with(Consumer<T> withCalls) {
    withCalls.accept(this);
    return this;
  }
```

As for adding a flag, that is not a problem.  `includeByNameAccessors` seems like it would be consistent with the other flags in the project.

joelittlejohn:
@ctrimble Very good point, I hadn't considered Android. I also agree having a target version will be useful for Java 8.

That flag sounds fine (maybe includeDynamicAccessors?), it's a bit cryptic but the docs will help :smile:

ctrimble:
I like includeDynamicAccessors better.  I will add that in and start preparing for a merge.  I could really use input on the PR I opened specifically for the target option.

ctrimble:
Added the `includeDynamicAccessors` property.  It seems strange that this option also adds builder methods.

joelittlejohn:
Took me a while to understand that comment :) I don't think that's a problem.

ctrimble:
I have rebased this on master.  Some naming alignment and clean up should be done before a merge happens.

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/PropertiesByNameRule.java
```diff
@@ -0,0 +1,371 @@
...
+import com.sun.codemodel.JType;
+import com.sun.codemodel.JTypeVar;
+import com.sun.codemodel.JVar;
+import com.sun.codemodel.JClass;
+
+import static com.sun.codemodel.JExpr.*;
+import static com.sun.codemodel.JMod.*;
+import static org.jsonschema2pojo.util.LanguageFeatures.*;
+
+public class PropertiesByNameRule implements Rule<JDefinedClass, JDefinedClass> {
```
ctrimble: This rule should be called DynamicPropertiesRule, to align naming with the flag.

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/RuleFactory.java
```diff
@@ -356,5 +356,13 @@ public NameHelper getNameHelper() {
     public Rule<JType, JType> getMediaRule() {
         return new MediaRule(this);
     }
+    
+    /**
+     * Provides a rule instance that adds methods for getting/setting
+     * properties by name.
+     */
+    public Rule<JDefinedClass, JDefinedClass> getPropertiesByNameRule() {
```
ctrimble: This should be getDynamicPropertiesRule, to align the name with the flag.

jsonschema2pojo-core/src/test/java/org/jsonschema2pojo/rules/PropertiesByNameRuleTest.java
```diff
@@ -0,0 +1,69 @@
...
+import com.sun.codemodel.JCodeModel;
+import com.sun.codemodel.JDefinedClass;
+import com.sun.codemodel.JFieldRef;
+import com.sun.codemodel.JMethod;
+import com.sun.codemodel.JVar;
+
+import static com.sun.codemodel.JMod.*;
+import static com.sun.codemodel.JExpr.*;
+
+public class PropertiesByNameRuleTest {
```
ctrimble: This should be DynamicPropertiesRuleTest, to align name with flag.

jsonschema2pojo-integration-tests/src/test/java/org/jsonschema2pojo/integration/PropertiesByNameIT.java
```diff
@@ -0,0 +1,354 @@
...
+import static org.jsonschema2pojo.integration.util.CodeGenerationHelper.config;
+import static org.jsonschema2pojo.integration.util.CodeGenerationHelper.generateAndCompile;
+import static org.junit.Assert.*;
+
+import java.lang.reflect.InvocationTargetException;
+import java.util.Map;
+
+import org.junit.Test;
+
+public class PropertiesByNameIT {
```
ctrimble: This should be DynamicPropertiesIT, to align name with flag.

jsonschema2pojo-maven-plugin/src/main/java/org/jsonschema2pojo/maven/Jsonschema2PojoMojo.java
```diff
@@ -445,6 +445,15 @@
     private String targetVersion = "1.6";
 
     /**
+     * Whether to include dynamic getters/setters/builders or to omit these accessor methods.
```
ctrimble: This should be `Whether to include dynamic getters, setters, and builders or to omit these methods.` in all documentation.

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/DynamicPropertiesRule.java
```diff
@@ -0,0 +1,380 @@
...
+    /**
+     * protected boolean set$( String name, Object value ) { switch( name ) { //
+     * cases for the properties defined on this schema. default: return
+     * super.set$( name, value ); // if extends return false; // if not extends.
+     * } }
+     * 
+     * @param jclass
+     * @param currentSchema
+     * @param node
+     */
```
ctrimble: Drop this documentation, it is innacurate.

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/DynamicPropertiesRule.java
```diff
@@ -0,0 +1,380 @@
...
+    public static final String DEFINED_GETTER_NAME = "declaredPropertyOrNotFound";
+
+    private RuleFactory ruleFactory;
+
+    public DynamicPropertiesRule(RuleFactory ruleFactory) {
+        this.ruleFactory = ruleFactory;
+    }
+
+    @Override
+    public JDefinedClass apply(String nodeName, JsonNode node, JDefinedClass jclass, Schema currentSchema) {
```
ctrimble: This should have javadocs

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/DynamicPropertiesRule.java
```diff
@@ -0,0 +1,380 @@
...
+import com.sun.codemodel.JFieldRef;
+import com.sun.codemodel.JInvocation;
+import com.sun.codemodel.JMethod;
+import com.sun.codemodel.JOp;
+import com.sun.codemodel.JSwitch;
+import com.sun.codemodel.JType;
+import com.sun.codemodel.JTypeVar;
+import com.sun.codemodel.JVar;
+
+public class DynamicPropertiesRule implements Rule<JDefinedClass, JDefinedClass> {
```
ctrimble: There should be documentation here describing the functionality.

ctrimble:
@joelittlejohn I think this PR is ready to go, unless you see anything that should be cleaned up.

PR #391, opened 2015-07-16
Jophde fix lib projects

jophde:
I couldn't get my library project to build.  The issue was the plugin was calling android.applicationVariants always.  For library projects it should be android.libraryVariants.


joelittlejohn:
Thanks!

jophde:
When do you expect to be pushing a new release with this change?

PR #400, opened 2015-08-19
Improve the property 'deserializationClassProperty' 

iduetxe:
The option deserializationClassProperty is an object which indicates its sons and the property used to deserialized. Using this information the plugin generates the Jackson annotation @JsonTypeInfo and @JsonSubTypes


ctrimble:
@iduetxe this PR is old and does not seem to have much traction.  I am going to close this.  If you want to pick this back up, let me know.

PR #405, opened 2015-08-20
Add configuration option to make generated types serializable

samskiter:
WIP.

 have made quick untested changes to most classes. No tests added yet... Not sure what to do about docs for the @since line...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,17 @@
 # Changelog
 
+## 0.4.14
+* Fix 'Unable to load class org.jsonschema2pojo.NoopAnnotator' from Gradle plugin ([#395](https://github.com/joelittlejohn/jsonschema2pojo/pull/395))
+* Library projects should use android.libraryVariants instead of android.applicationVariants  ([#391](https://github.com/joelittlejohn/jsonschema2pojo/pull/391))
+* Allow nullable type (instead of Object) when JSON type is an array with ["string", "null"] ([#390](https://github.com/joelittlejohn/jsonschema2pojo/issues/390))
+* Allow custom names to be supplied when creating an enum ([#385](https://github.com/joelittlejohn/jsonschema2pojo/issues/385))
+* Allow disabling additionalProperties generation with a global config option ([#376](https://github.com/joelittlejohn/jsonschema2pojo/issues/376))
+* 'javaInterfaces' which contain generics produce syntactically incorrect imports  ([#317](https://github.com/joelittlejohn/jsonschema2pojo/issues/317))
+
+## 0.4.13
+* RequiredArrayRule not working for properties with _ or -. ([#377](https://github.com/joelittlejohn/jsonschema2pojo/issues/377))
+* Maven plugin shows sourcePaths error for includes/excludes when sourcePaths is not defined ([#374](https://github.com/joelittlejohn/jsonschema2pojo/issues/374))
+
 ## 0.4.12
 * Didn't work in Android Studio with with current master branch '040e3ae'  ([#370](https://github.com/joelittlejohn/jsonschema2pojo/issues/370))
 * Use the correct ClassLoader in createFromParcel for lists ([#356](https://github.com/joelittlejohn/jsonschema2pojo/pull/356))
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -35,3 +35,6 @@
 * Huang Wei
 * Sim Sun
 * Srikalyan Swayampakula
+* Nate Stoddard
+* Josh Deffibaugh
+* Jaeho Suk
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ A very simple Maven example:
 <plugin>
     <groupId>org.jsonschema2pojo</groupId>
     <artifactId>jsonschema2pojo-maven-plugin</artifactId>
-    <version>0.4.13</version>
+    <version>0.4.14</version>
     <configuration>
         <sourceDirectory>${basedir}/src/main/resources/schema</sourceDirectory>
         <targetPackage>com.example.types</targetPackage>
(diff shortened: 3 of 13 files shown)


joelittlejohn:
Cool, good stuff. Feel free to add 0.4.15 into `@since`.

joelittlejohn:
Updated the title here to something that will fit well into the release notes.

samskiter:
Right, for reference, this is how intelliJ does it:

```java
public static long computeDefaultSUID(@Nullable PsiClass psiClass) {
        if (psiClass == null) {
            return -1L;
        }

        final PsiManager    manager      = psiClass.getManager();
        final Project       project      = manager.getProject();
        final JavaPsiFacade psiFacade    = JavaPsiFacade.getInstance(project);
        final PsiClass      serializable = psiFacade.findClass(SERIALIZABLE_CLASS_NAME, GlobalSearchScope.allScope(project));

        if (serializable == null) {
            return -1L;
        }

        final boolean isSerializable = psiClass.isInheritor(serializable, true);

        if (!isSerializable) {
            return 0L;
        }

        final SerialVersionUIDBuilder serialVersionUIDBuilder = new SerialVersionUIDBuilder(psiClass);

        try {
            final ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream();
            final DataOutputStream      dataOutputStream      = new DataOutputStream(byteArrayOutputStream);
            final String                className             = psiClass.getQualifiedName();

            dataOutputStream.writeUTF(className);

            final PsiModifierList   classModifierList = psiClass.getModifierList();
            int                     classModifiers    = MemberSignature.calculateModifierBitmap(classModifierList);
            final MemberSignature[] methodSignatures  = serialVersionUIDBuilder.getNonPrivateMethodSignatures();

            if (psiClass.isInterface()) {
                classModifiers |= 0x200;
                if (methodSignatures.length == 0) {
                    classModifiers &= 0xfffffbff;
                }
            }
            dataOutputStream.writeInt(classModifiers);

            final PsiClass[] interfaces = psiClass.getInterfaces();

            Arrays.sort(interfaces, INTERFACE_COMPARATOR);
            for (PsiClass aInterfaces : interfaces) {
                final String name = aInterfaces.getQualifiedName();
                dataOutputStream.writeUTF(name);
            }

            final MemberSignature[] nonPrivateFields       = serialVersionUIDBuilder.getNonPrivateFields();
            final MemberSignature[] staticInitializers     = serialVersionUIDBuilder.getStaticInitializers();
            final MemberSignature[] nonPrivateConstructors = serialVersionUIDBuilder.getNonPrivateConstructors();

            Arrays.sort(nonPrivateFields);
            Arrays.sort(nonPrivateConstructors);
            Arrays.sort(methodSignatures);

            writeMemberSignatures(dataOutputStream, nonPrivateFields);
            writeMemberSignatures(dataOutputStream, staticInitializers);
            writeMemberSignatures(dataOutputStream, nonPrivateConstructors);
            writeMemberSignatures(dataOutputStream, methodSignatures);

            dataOutputStream.flush();

            final MessageDigest digest           = MessageDigest.getInstance("SHA");
            final byte[]        digestBytes      = digest.digest(byteArrayOutputStream.toByteArray());
            long                serialVersionUID = 0L;

            for (int i = Math.min(digestBytes.length, 8) - 1; i >= 0; i--) {
                serialVersionUID = serialVersionUID << 8 | (long)(digestBytes[i] & 0xff);
            }

            return serialVersionUID;

        } catch (IOException exception) {
            final InternalError internalError = new InternalError(exception.getMessage());
            internalError.initCause(exception);
            throw internalError;
        } catch (NoSuchAlgorithmException exception) {
            final RuntimeException securityException = new SecurityException(exception.getMessage());
            securityException.initCause(exception);
            throw securityException;
        }
    }

    private static void writeMemberSignatures(DataOutputStream dataOutputStream, MemberSignature[] signatures) throws IOException {
        for (MemberSignature field : signatures) {
            dataOutputStream.writeUTF(field.getName());
            dataOutputStream.writeInt(field.getModifiers());
            dataOutputStream.writeUTF(field.getSignature());
        }
    }
```

samskiter:
I think the intention there is to "version" the class based on everything public about it...

joelittlejohn:
The problem with a random number here is that every release will use different version and will be considered incompatible from the last. For most people, I think a fixed value would be preferable (of the two easy options).

We might be able to do something similar but simpler than the intellij strategy. How about hashing together all field names and their types?

samskiter:
Yes, a good point. I think thats a good strategy, afterall the only public think about our objects are the names/types right?

(for my own reference for later)
```
private static int FIELD = (JMod.PUBLIC | JMod.PRIVATE | JMod.PROTECTED
            | JMod.STATIC | JMod.FINAL
            | JMod.TRANSIENT | JMod.VOLATILE);
JFieldVar field = jclass.field(JMod.PRIVATE, propertyType, propertyName);
```

joelittlejohn:
0.4.15 has plenty in it so I'm going to cut that release and move this into 0.4.16. Give me a shout when you're ready for a review/merge. Cheers

samskiter:
Sorry! I'm away on holiday, without computer. Will be more available from
next Tuesday onward...

Sam

joelittlejohn:
@samskiter No problem - no rush.

samskiter:
By the way, what's the fastest way to test and iterate the code? do you use a particular development environment?

joelittlejohn:
@samskiter For Java I use Eclipse. The best way to iterate on this project (IMO) is to create a good integration test (IT) for your feature, then work to meet this. If you are interested in what code is being generated during the IT you can add `-Ddebug=true` and the Java source will be printed to stdout. There's a helper used by all the ITs to generate and compile sources.

samskiter:
Really need to work on this, sorry for delays.

tomliuxd:
I suggest to make the configuration as number, instead of boolean.  

when the number is 0, serializable is disabled
when the number is <0, serial UID is automatically generated
when the number > 0, use the number ad serial UID.

tomliuxd:
FYI, we are adding "java.io.Serializable" to JSON file to enable serializable (without serialVersionUUID)

Here is one example:

`{
	"type": "object",
	"$schema": "http://json-schema.org/draft-04/hyper-schema",
	"title": "Measurement",
	"description": "Measurement to represent item dimensions like length, width, height and weight etc.",
	"id": "common_measurement:v1",
	"javaInterfaces" : ["java.io.Serializable"],
	"properties": 
	{
		"value": 
		{
			"type": "string",
			"description": "Value this measurement represents."
		},

		"unit": 
		{
			"type": "string",
			"description": "Unit in which the value is represented."
		}
	},

	"required": 
	[
		"value",
		"unit"
	]
}
`

samskiter:
Having every class have the same serialVersionUUID isn't a great idea. I
really need to finish the PR!

tomliuxd:
In most cases, people want to have control on serial version UID.  It can be either defined by some configuration option, or individually defined on each JSON file.  

For us, we like to maintain it by some configuration option.  We have a data model which include hundreds of JSON files.  It would be a nightmare for us to maintain serial version UID in each file.  Since the data model is released as a single JAR file, it would be very convenience for us to just label everything with same UID.
 
we uses JSON files to define a data model which include many classes.  We assign version number to the data model and like to label all generated java class with it (by serialVersionUUID).

In my suggestion, people have all three options.  they can choose whatever they want

0) disabled
<0) enabled with auto generated UID (or defined in JSON file)
>0) enabled with all classes assigned with same UID

samskiter:
@joelittlejohn certainly in need of review, i'll try and make compatible for merge...

samskiter:
Moved to https://github.com/joelittlejohn/jsonschema2pojo/pull/543

PR #406, opened 2015-08-26
Allow default enum value to be returned in fromValue instead of throwing IllegalArgumentException

yomik:
If flag javaIgnoreUnknownValue is set and is true, allow the default value for an enum property (if set) to be returned instead of an exception being thrown when factory argument is not a valid enum value.

This is useful e.g. when API gets updated with new enum values, to allow apps running on older versions to still be able to parse the new API response.

This is just a quick implementation I made for a feature I really wish existed ; no unit tests were written but it was tested on my existing schema.

I made it as a PR in case other people might need this too.

### DIFF
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/EnumRule.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/EnumRule.java
@@ -16,24 +16,6 @@
 
 package org.jsonschema2pojo.rules;
 
-import static java.util.Arrays.*;
-import static org.apache.commons.lang3.StringUtils.*;
-import static org.jsonschema2pojo.rules.PrimitiveTypes.*;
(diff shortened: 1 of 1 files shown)


joelittlejohn:
Thanks for submitting this @yomik! I will admit, I'm not hot on the idea of merging this. I think there's something dangerous about choosing the default value here and even behind a config option, it just feels like this option encourages a bad pattern. 

I'm going to close this for now, but if people start to give this PR some love then maybe we can think about adding some tests and merging it.

yomik:
Yeah I definitely see your point, I'm not hot on this either; in an ideal world, our API would be versioned so as to avoid the issues mentioned, but it is not my case so this workaround would have been helpful :)

Anyway, no worries, a better solution would be to allow custom Rules to be set, which I hope can see the light someday.

perdonare:
so how to remove fromValue method

PR #411, opened 2015-08-28
fixes #410 - new JPackage is created when handling $ref reflecting the location of included schema

### LINKED ISSUE #410: $ref pointing to subdirectory generates classes in wrong packages
When $ref includes a schema from different directory, referred class is generated in the package containing referring schema, e.g.:

*schema/bar/a.json*
```javascript
{
  "$schema" : "http://json-schema.org/draft-04/schema#",
  "type" : "object",
  "properties" : {
    "refA" : {
      "$ref": "../foo/b.json"
    }
  }
}
```

*schema/foo/b.json*
```javascript
{
  "$schema" : "http://json-schema.org/draft-04/schema#",
  "type" : "object",
  "properties" : {
    "propA" : {
      "type" : "string"
    }
  }
}
```

Both classes are generated in *bar* package.

I assumed that:
- sourceDirectory is the default package of generated classes 
- directory structure along with the schemas inside will be reflected in generated packages and classes

It's mainly caused by a this block of code in SchemaRule class:
```java
        if (schemaNode.has("$ref")) {
            schema = ruleFactory.getSchemaStore().create(schema, schemaNode.get("$ref").asText());
            schemaNode = schema.getContent();

            if (schema.isGenerated()) {
                return schema.getJavaType();
            }

            return apply(nodeName, schemaNode, generatableType, schema); // <-- see description below
        }
```

When *generatableType* is a *JPackage*, it points to current package instead of package of referred schema.

I have a piece of code that fixes this issue:
- generates classes in proper directories when $ref points to subdirectory (at any level)
- handles $ref containing #
- unit tests for common and edge cases, covering 100% of added/modified code
- leaves other protocols (http, java, resource, file, etc.) intact

But the questions is: Is it an bug or an improvement?
If it's an improvement, we should discuss the compliance with the spec of your plugin and handling of other protocols if needed.
If it's a bug, here's the patch fix necessary changes and tests:
https://github.com/joelittlejohn/jsonschema2pojo/compare/master...badhabit85:patch-1

rfoltyns:
new JPackage is created when handling $ref reflecting the location of included schema

### DIFF
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/SchemaRule.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/SchemaRule.java
@@ -16,9 +16,12 @@
 
 package org.jsonschema2pojo.rules;
 
-import com.fasterxml.jackson.databind.JsonNode;
 import org.jsonschema2pojo.Schema;
+import org.jsonschema2pojo.util.PackageUtil;
+
+import com.fasterxml.jackson.databind.JsonNode;
 import com.sun.codemodel.JClassContainer;
+import com.sun.codemodel.JPackage;
 import com.sun.codemodel.JType;
 
 /**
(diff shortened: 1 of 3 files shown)


rfoltyns:
Worked fine using 0.4.27.

Issue #415, opened 2015-09-03
equals() methods generated by jsonschema2pojo trigger warnings in sonarqube


beresfordt:
The equals() methods generated are of the general form

```
@Override
public boolean equals(Object other) {
    if (other == this) {
        return true;
    }
    if ((other instanceof ClassName) == false) {
        return false;
    }
    ClassName rhs = ((ClassName) other);
    return new EqualsBuilder().append(field, rhs.field).append(additionalProperties, rhs.additionalProperties).isEquals();
}
```

This triggers warnings in sonarqube due to squid rule [S2162](https://github.com/SonarSource/sonar-java/blob/master/java-checks/src/main/resources/org/sonar/l10n/java/rules/squid/S2162.html)

In short using instanceof to in equals to check class triggers a warning because subclasses of `ClassName` would pass the instanceof conditional.

To make the equals method compliant it should null check, then compare getClass of both objects:

```
if (other == null) {
    return false;
}
....
if (this.getClass() != other.getClass()) {
    return false;
}
```

joelittlejohn:
I think the implications of both approaches were well understood by the person that submitted this code originally, and I think they were inspired by Effective Java.

Here's a good summary of Josh Bloch's thoughts on the subject:

https://www.artima.com/intv/bloch17.html

This is a tricky one. Given the arguments on both sides, I prefer `getClass`, however there's a backwards compatibility concern to throw into the mix now, which makes me reticent to merge your PR.

beresfordt:
Thanks for the link; the debate between getClass and instanceof had completely passed me by until now! 

Having read Bloch's, and others, arguments for it I'm still in the getClass camp; instanceof breaking the symmetry of equals is a killer for me

I'd assumed it was an oversight rather than a design decision; if you have concerns about backwards compatibility bounce the request; it obviously would change behaviour for people relying on superclass equaling subclasses

ctrimble:
I am curious if this is worth a flag.

PR #416, opened 2015-09-03
Fix for equals method triggering warning in sonarqube

beresfordt:
fix for issue #415

Updated the equals method generated so that it compares getClass of this and other instead of using instanceof which can break on subclasses


joelittlejohn:
I'm going to close this one for now, it will be here for reference. I might merge this when we move to 0.5 (or something along those lines). For now, the silent change in behaviour is too problematic IMO.

PR #424, opened 2015-09-22
URN and JPA Annotation Support

Labi0:
Added a bit of a hack on the URN support but nevertheless still useful. Creates the package and class name based on the urn in the "id" attribute of the schema.
Added JPA Annotation support very rough working code.
Added MiscSupport Annotator to provide generated-class tracking (this functionality could be moved into NoopAnnotator but did not want to touch it).


joelittlejohn:
Thanks for submitting this. It's nice to have these things published so that others can see this.

I don't think it will be possible to merge this but thanks for making the code available.

PR #426, opened 2015-09-28
"extends" schema resolution now handles extending a type that isn't a…

### LINKED ISSUE #425: "extends" schema resolution assumes current schema is at root of file
Example of a schema that currently fails (also checked with http://www.jsonschema2pojo.org)

```json
{
  "definitions": {
    "parent": {
      "type": "object",
      "properties": {
        "propertyOfParent": {
          "type": "string"
        }
      }
    },
    "child": {
      "type": "object",
      "extends": {"$ref": "#/definitions/parent"},
      "properties": {
        "propertyOfChild": {
          "type": "string"
        }
      }
    }
  },
  "type": "object",
  "properties": {
    "child": {"$ref": "#/definitions/child"}
  }
}

```

markdingram:
…t the

root of the schema file

Resolves #425


joelittlejohn:
Great, thanks for submitting this and including a test case :D

PR #428, opened 2015-10-07
Make the quick lookup map 'final'

ankon:
There is no point in the lookup *not* being final really :)


joelittlejohn:
Thanks! :+1:

PR #430, opened 2015-10-07
adding "required" attribute to JsonProperty annotation

jmazzitelli:
This adds "required=true" or "required=false" to @JsonProperty annotations on fields, getters and setters but unfortunately this ONLY works if the "required" flag is set on the individual property in the JSON schema. If using "required" array in the schema, that still is ignored. I don't know how to fix that. Hopefully, someone else does :)

So if you have this in your schema, this will add the "required" annotation attribute:

    "properties": {
        "msg": {
          "type": "string",
          "required": true
        }
    }

    but this will not:

      "required": ["msg"]

### DIFF
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/Annotator.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/Annotator.java
@@ -73,8 +73,10 @@ public interface Annotator {
      *            JSON property
      * @param propertyName
      *            the name of the JSON property that this getter gets
+     * @param propertyNode
+     *            the schema node defining this property
      */
-    void propertyGetter(JMethod getter, String propertyName);
+    void propertyGetter(JMethod getter, String propertyName, JsonNode propertyNode);
 
     /**
      * Add the necessary annotation to mark a Java method as the setter for a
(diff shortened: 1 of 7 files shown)


jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/Annotator.java
```diff
@@ -73,8 +73,10 @@
      *            JSON property
      * @param propertyName
      *            the name of the JSON property that this getter gets
+     * @param propertyNode
+     *            the schema node defining this property
      */
-    void propertyGetter(JMethod getter, String propertyName);
+    void propertyGetter(JMethod getter, String propertyName, JsonNode propertyNode);
```
ctrimble: This change alters the existing contract for Annotator and will break existing custom annotators.  This change should be additive.

thachhoang: @ctrimble if we overload `propertyGetter`, then in the rule classes, do we call the new version / the old version / both?

bwoester:
```suggestion
    default void propertyGetter(JMethod getter, String propertyName, JsonNode propertyNode) {
        propertyGetter(getter, propertyName);
    }
```

@thachhoang with this, we can call the new overload from the rule classes. Existing custom annotators will still get the the expected parameters delivered, but the jackson annotator can override the new propertyGetter and make use of the propertyNode.

ctrimble:
@jmazzitelli  I am closing this PR, since it does not fully support required and has been stalled out for some time.  We can reopen it if you feel like picking it back up.

jmazzitelli:
Given that this is 10 years old, I think I'm good LOL

PR #432, opened 2015-10-19
Fix path to jar in windows script for the case it is run not from home dir

redreev:
Hi guys, I've run windows cli jsonschema2pojo.bat cli and it failed with error saying jar file is not found. Problem was that I've run the script not from jsonschema2pojo home, but from my own folder, and path to the jsonschema2pojo jar in the script is relative to jsonschema2pojo home.

Here's fix for the issue. I've changed relative path to the jar to absolute path, so now the windows cli runner works fine not just from jsonschema2pojo home.


joelittlejohn:
Thanks! :+1:

PR #435, opened 2015-10-26
Make Inflector more amenable to thread safety.

markdingram:
- reduced public API to 2 synchronized methods: singularize/pluralize
- eagerly construct singleton to avoid construction race
- remove unnecessary static lists/methods

(encountering sporadic failures from a parallel Maven build using RAML JAXRS Codegen Maven plugin)


joelittlejohn:
Thanks! :+1: 

This class was taken from http://dzone.com/snippets/java-inflections and used/licensed with permission from the author. This same exact class is used without modification in a few other projects, so I guess we have the best version now :D

PR #444, opened 2015-11-11
 This change is continuation of https://github.com/iduetxe/jsonschema…

jmgrozas:
This change is continuation of https://github.com/iduetxe/jsonschema2pojo

1 - Add complete support for hierarchy with jackson2 annotation JsonTypeInfo combinated values, for example:

```
{
    "type": "object",
    "javaType" : "com.sample.ExampleDTO",
    "deserializationClassProperty":{
        "propertyName":"childType", // Property name support
        "idValueType":"NAME", // id values types support for combinated JsonTypeInfo.Id
        "includeValueType":"PROPERTY", // JsonTypeInfo.As all values for supports 
        "children": // chidren declaration
            [
                {
                    "className": "com.sample.ExampleChild1DTO",
                    "value": "Child1"
                },
                {
                    "className": "com.sample.ExampleChild2DTO",
                    "value": "Child2"
                }
            ]
    }
}

```

All Enums will be validated, if error occurs print a specific message about it. 

Special behavior for the value EXTERNAL PROPERTY annotation will be put on field not on class.

2 - Add support for JsonIgnore annotation, for field, setter or getter, for Example:

```
    "ignore" : true, // put JsonIgnore annotation on field getter and setter
    "ignoreSetter" : true,// put JsonIgnore annotation only on setter
    "ignoreGetter" : true // put JsonIgnore annotation only on getter
```

```
    "properties":{
        "signature": {
            "type": "string",
            "minLength": 1, 
            "maxLength": 32,
            "ignore" : true,
            "pattern": "^[a-zA-Z0-9]+$"
        },

     ....

```

This mechanism will enable us to play with the different techniques of serialization

### DIFF
--- a/jsonschema2pojo-ant/pom.xml
+++ b/jsonschema2pojo-ant/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <artifactId>jsonschema2pojo</artifactId>
         <groupId>org.jsonschema2pojo</groupId>
-        <version>0.4.17</version>
+        <version>0.4.17-SNAPSHOT</version>
     </parent>
 
     <artifactId>jsonschema2pojo-ant</artifactId>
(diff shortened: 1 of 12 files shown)


joelittlejohn:
Thanks for publishing these changes publicly (they may well be useful to others), however I'm not interested in bringing these changes into the core.

PR #454, opened 2015-11-27
Specify target JVM version

ctrimble:
Exposes a `targetVersion` property to the rules and test compiler, specifying the target Java version to generate for.

This fixes an issue during cross compilation where the source generated is valid for the JVM running the test suite but not for the JVM being targeted.  For instance, auto boxing a cast like `(int)value` works in `Java7` but not in `Java6`.

__Interface Changes__:

- Maven: adds a `targetVersion` config property, with a default value of `maven.compiler.target`.
- Gradle: adds a `targetVersion` property, with a default value of `1.6`
- Ant: adds a `targetVersion` property, with a default value of `1.6`
- CLI: adds options `-tv` and `--target-version`, with a default value of `1.6`

__Limitations__:  This patch does not set the bootclasspath from `toolchains.xml`.  So, there can still be issues where `Java7` classes and methods are referenced when generating code for `Java6`.  I attempted to add support using classes from Maven, but there is a classpath conflict with the testing for `Ant`.  `-Xlint:-options` has been set during testing, to suppress warnings related to this.

The changes here were taken from #386.

### DIFF
--- a/jsonschema2pojo-cli/src/main/java/org/jsonschema2pojo/cli/Arguments.java
+++ b/jsonschema2pojo-cli/src/main/java/org/jsonschema2pojo/cli/Arguments.java
@@ -131,6 +131,9 @@ public class Arguments implements GenerationConfig {
 
     @Parameter(names = { "-da", "--disable-accessors" }, description = "Whether to omit getter/setter methods and create public fields instead.")
     private boolean disableAccessors = false;
+    
+    @Parameter(names = { "-tv", "--target-version" }, description = "The target version for generated source files.")
+    private String targetVersion = "1.6";
 
     private static final int EXIT_OKAY = 0;
     private static final int EXIT_ERROR = 1;
@@ -319,6 +322,11 @@ public class Arguments implements GenerationConfig {
     @Override
     public boolean isIncludeAccessors() {
         return !disableAccessors;
+    }
+
+    @Override
+    public String getTargetVersion() {
+        return targetVersion;
     }
 
 }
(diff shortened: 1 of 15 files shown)


ctrimble:
@joelittlejohn this change adds a new configuration option.  The naming is strange for the Ant task and I am not very proficient at Gradle, so any notes around those items would be appreciated.

ctrimble:
@joelittlejohn I think the PR that adds the target option is ready to merge.

ctrimble:
Looks like I forgot getTarget( String target ) in the ant task and associated documentation.

jsonschema2pojo-integration-tests/src/test/java/org/jsonschema2pojo/integration/ant/Jsonschema2PojoTaskIT.java
```diff
@@ -65,7 +65,7 @@ public void antTaskDocumentationIncludesAllProperties() throws IntrospectionExce
         String documentation = FileUtils.readFileToString(new File("../jsonschema2pojo-ant/src/site/Jsonschema2PojoTask.html"));
 
         for (Field f : Jsonschema2PojoTask.class.getDeclaredFields()) {
-            assertThat(documentation, containsString(f.getName()));
+            assertThat(documentation, containsString(">"+f.getName()+"<"));
```
joelittlejohn: Good spot :smile:

joelittlejohn:
Looking good to me. One thing, I think target here is a little confusing as it's easy to confuse this with the target directory. I prefer targetVersion everywhere, as you've used in the CLI.

joelittlejohn:
Other than that change, I can't see anything else to comment on. Nice work, I'll merge when you're ready.

jsonschema2pojo-cli/src/main/java/org/jsonschema2pojo/cli/Arguments.java
```diff
@@ -131,6 +131,9 @@
 
     @Parameter(names = { "-da", "--disable-accessors" }, description = "Whether to omit getter/setter methods and create public fields instead.")
     private boolean disableAccessors = false;
+    
+    @Parameter(names = { "-tv", "--target-version" }, description = "Target version for generated source files.")
```
ctrimble: This is the best way to describe this setting.  Make all descriptions like this one.

ctrimble:
@joelittlejohn `target` has been changed to `targetVersion`.  I also made the documentation consistent across all the interfaces.  This is ready to merge.

Issue #461, opened 2015-12-10
Databases element of type array generated class name Databasis instead of Database


rpatrick00:
I have a schema that has a property defined like this:

<pre>
                "databases": {
                    "type": "array",
                    "items": {
                        "type": "object",
                        "properties": {
                            "id": { "type": "string" },
                            "nodes": {
                                "type": "array",
                                "items": {
                                    "type": "object",
                                    "properties": {
                                        "id": { "type": "string" },
                                        "connectString": { "type": "string" }
                                    }
                                }
                            }
                        }
                    }
                }
</pre>

The code being generated is generating a class named "Databasis" instead of "Database" (singular form of "databases").  How can I get the generator to generate the correct name?

joelittlejohn:
Closed by #462

PR #462, opened 2015-12-11
#461 - fixing the singular form of databases to be database

rpatrick00:
Added database in what looked like the right place and added a test  to verify that it i sbeing converted as expected.  Also, corrected the spelling of mattress in the InflectorTest.


joelittlejohn:
Thanks!

PR #465, opened 2015-12-13
Prefix / Suffix support when javaType is not specified

pwilder:
Please let me know if I need:

* More tests
* More comments
* Style changes
* Alternate behaviour (e.g. capitlization on nodeName and prefix)

Thanks,

Philip

### DIFF
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/ObjectRule.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/ObjectRule.java
@@ -425,9 +425,27 @@ public class ObjectRule implements Rule<JPackage, JType> {
     }
 
     private String getClassName(String nodeName, JPackage _package) {
-        String className = ruleFactory.getNameHelper().replaceIllegalCharacters(capitalize(nodeName));
+        String prefix = ruleFactory.getGenerationConfig().getClassNamePrefix();
+        String suffix = ruleFactory.getGenerationConfig().getClassNameSuffix();
+        String capitalizedNodeName = capitalize(nodeName);
+        String fullNodeName = createFullNodeName(capitalizedNodeName, prefix, suffix);
+
+        String className = ruleFactory.getNameHelper().replaceIllegalCharacters(fullNodeName);
         String normalizedName = ruleFactory.getNameHelper().normalizeName(className);
         return makeUnique(normalizedName, _package);
+    }
+
+    private String createFullNodeName(String nodeName, String prefix, String suffix) {
+        String returnString = nodeName;
+        if (prefix != null) {
+            returnString = prefix + returnString;
+        }
+
+        if (suffix != null) {
+            returnString = returnString + suffix;
+        }
+
+        return returnString;
     }
 
     private String makeUnique(String className, JPackage _package) {
--- a/jsonschema2pojo-integration-tests/src/test/java/org/jsonschema2pojo/integration/config/PrefixSuffixIT.java
+++ b/jsonschema2pojo-integration-tests/src/test/java/org/jsonschema2pojo/integration/config/PrefixSuffixIT.java
@@ -28,7 +28,7 @@ public class PrefixSuffixIT {
         ClassLoader resultsClassLoader = generateAndCompile("/schema/properties/primitiveProperties.json", "com.example");
         resultsClassLoader.loadClass("com.example.PrimitiveProperties");
     }
-    
+
     @Test
     public void customClassPrefix() throws ClassNotFoundException{
 
@@ -37,12 +37,19 @@ public class PrefixSuffixIT {
     }
     
     @Test
-    public void defaultClassSufix() throws ClassNotFoundException{
+    public void noCapsCustomClassPrefix() throws ClassNotFoundException{
+
+        ClassLoader resultsClassLoader = generateAndCompile("/schema/properties/primitiveProperties.json", "com.example", config("classNamePrefix","abstract"));
+        resultsClassLoader.loadClass("com.example.abstractPrimitiveProperties");
+    }
+
+    @Test
+    public void defaultClassSuffix() throws ClassNotFoundException{
 
         ClassLoader resultsClassLoader = generateAndCompile("/schema/properties/primitiveProperties.json", "com.example");
         resultsClassLoader.loadClass("com.example.PrimitiveProperties");
     }
-    
+
     @Test
     public void customClassSuffix() throws ClassNotFoundException{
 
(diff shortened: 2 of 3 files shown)


jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/ObjectRule.java
```diff
@@ -425,11 +425,30 @@ private void addInterfaces(JDefinedClass jclass, JsonNode javaInterfaces) {
     }
 
     private String getClassName(String nodeName, JPackage _package) {
-        String className = ruleFactory.getNameHelper().replaceIllegalCharacters(capitalize(nodeName));
+        String prefix = ruleFactory.getGenerationConfig().getClassNamePrefix();
+        String suffix = ruleFactory.getGenerationConfig().getClassNameSuffix();
+        String fullNodeName = createFullNodeName(nodeName, prefix, suffix);
+
+        //Note that only the first letter of the fullNodeName will be capitalized not the node name
+        //so prefix + nodeName becomes PrefixnodeName
```
ctrimble: To be consistent, the root name should probably be capitalized.  If you define `javaType` as the value that jsonschema2pojo would pick, the functionality should be identical.

jsonschema2pojo-integration-tests/src/test/java/org/jsonschema2pojo/integration/config/PrefixSuffixIT.java
```diff
@@ -28,53 +29,74 @@ public void defaultClassPrefix() throws ClassNotFoundException {
...
         resultsClassLoader.loadClass("com.example.NotExistingPrimitiveProperties");
     }
 
-    
+    @Test
+    public void customClassPrefixNoJavaType() throws ClassNotFoundException {
+
+        ClassLoader resultsClassLoader = generateAndCompile("/schema/properties/primitivePropertiesNoJavaType.json",
+                "com.example", config("classNamePrefix","Prefix"));
+        resultsClassLoader.loadClass("com.example.PrefixprimitivePropertiesNoJavaType");
```
ctrimble: The value here should be `com.example.PrefixPrimitivePropertiesNoJavaType`

jsonschema2pojo-integration-tests/src/test/java/org/jsonschema2pojo/integration/config/PrefixSuffixIT.java
```diff
@@ -28,53 +29,74 @@ public void defaultClassPrefix() throws ClassNotFoundException {
...
+        ClassLoader resultsClassLoader = generateAndCompile("/schema/properties/primitivePropertiesNoJavaType.json",
+                "com.example", config("classNameSuffix","Suffix"));
+        resultsClassLoader.loadClass("com.example.PrimitivePropertiesNoJavaTypeSuffix");
+    }
+
+    @Test
+    public void customClassPrefixAndSuffixNoJavaType() throws ClassNotFoundException {
+        ClassLoader resultsClassLoader = generateAndCompile("/schema/properties/primitivePropertiesNoJavaType.json",
+                "com.example", config("classNamePrefix", "prefix", "classNameSuffix","Suffix"));
+        resultsClassLoader.loadClass("com.example.PrefixprimitivePropertiesNoJavaTypeSuffix");
```
ctrimble: The value here should be `com.example.PrefixPrimitivePropertiesNoJavaTypeSuffix`

jsonschema2pojo-integration-tests/src/test/java/org/jsonschema2pojo/integration/config/PrefixSuffixIT.java
```diff
@@ -28,53 +29,74 @@ public void defaultClassPrefix() throws ClassNotFoundException {
...
+    }
```
ctrimble: It doesn't look like the existing functionality does anything to the prefix and suffix, so a test to confirm the existing functionality and a test to make sure the new functionality matches would be good.  Perhaps check `classNamePrefix ` = `prefix` and `classNameSuffix` = `suffix`, expecting `com.example.prefixPrimitivePropertiesNoJavaTypesuffix`.

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/ObjectRule.java
joelittlejohn: Agreed. Without capitalizing nodeName, I think the majority of people will be disappointed with how this behaves.

pwilder: Sounds good.

jsonschema2pojo-integration-tests/src/test/java/org/jsonschema2pojo/integration/config/PrefixSuffixIT.java
pwilder: Once the change above is made, agreed.

jsonschema2pojo-integration-tests/src/test/java/org/jsonschema2pojo/integration/config/PrefixSuffixIT.java
pwilder: +1

jsonschema2pojo-integration-tests/src/test/java/org/jsonschema2pojo/integration/config/PrefixSuffixIT.java
pwilder: Interesting point, but I believe the code I wrote will currently capitalize the prefix which is by design. While the capitalization of the suffix is potentially optionally I think the capitalization of the prefix is important to adhere to java class name conventions.

pwilder: Interesting. I ran the test as you suggested and you are correct that in the existing code prefix is *not* capitalized. I think you can make a compelling argument that prefix should be capitalized to adhere to java conventions but I'm not eager to change more then necessary in this code inspection. I will adjust the behaviour of my new code to match that of the old.

ctrimble:
LGTM.  Just needs to be squashed into a single commit.

joelittlejohn:
@pwilder Would you be able to fixup e574960 and 23dae18 into 32b2499?

I'd start with:
```
git rebase -i 32b2499^
```

this will allow you to fixup the extra commits.

Let me know if you're having difficult and I can take care of this.

Thanks!!

pwilder:
I'll give the rebase a try and let you know if I hit issues.

pwilder:
Squashing on my local side was fairly trivial but the commit log suggests I've made things worse rather then better (5 commits vs. 3). Independently rolling back my local content or the remote content (per http://stackoverflow.com/a/588471) looks manageable but I'm somewhat worried that trying to do both will just lose all my changes (not that they are terribly difficult to reproduce in this case).

ctrimble:
@pwilder It looks like you merged changes from master in, instead of replaying history on it.

From where you are at, you can use these steps to create the branch you need:

__note:__ assumes the remote `upstream` is the main project and that your current checkout is just like `pwilder:master`.

```
git checkout -b master-cleanup # make a copy of your branch
git reset --hard HEAD^2        # throw away the last 2 commits
git reset --soft HEAD~2        # restage the 2 commits before that
git commit --amend --no-edit   # amend your first commit
git rebase upstream/master     # replay history on the upstream master
```

See if the diffs look correct:

```
git diff upstream/master
```

And if the history is correct:

```
git log -2
```

If everything is good, make this your master

```
git branch -D master
git branch -m master
```

Double check where you are at and force push back to your remote master.

joelittlejohn:
@pwilder I think what's happened here, is that you've done the correct squashing locally, but instead of force pushing back to your fork, you've pulled and merged from it instead.

When you rebase those local commits, your local history is basically incompatible with what's in GitHub. At this point, you want to `git push -f origin master` to force the remote repo to match your local repo. If you pull from it, you'll be keeping _all_ commits that exist locally and remotely and adding merge to reconcile the whole lot.

At this point, the easiest way to fix this is:

```
git reset --hard d78d121   # reset back to your good, squashed commit
git push -f origin master  # make origin/master match your local repo exactly 
```

I'm not so worried about rebasing against upstream (I don't mind a branch showing in the history, as long as we _can_ merge without conflicts). @ctrimble is suggesting a process that (for even more bonus points to you) keeps our log central log tidy too. Once you've done the above, you can do this very easily by running:

```
git remote add upstream git@github.com:joelittlejohn/jsonschema2pojo.git
git pull --rebase upstream master
```

If you have any trouble at all or you feel you're getting things into a worse state, just use `git reset --hard d78d121` to get back to your good, squashed commit.

pwilder:
Sorry for the delayed response, holidays have limited my availability.

I started with the instructions from joe and decided to keep things simple and just follow through to the end (with a minor change to use the https url for simplicity). I'm only seeing 1 commit record and the contents of the commit looks good so hopefully this is roughly what we are looking for. 

It's been an interesting exercise in git submission discipline that has made clear I could benefit from some more time tinkering with the various git commands. Thank you both for the support.

joelittlejohn:
Thanks @pwilder, glad this was a useful exercise and not an exercise in frustration! :smile:

PR #470, opened 2015-12-20
Cleanup for new date time type feature

ctrimble:
Cleans up some minor issues with the new date time functionality:

1. correct formatting
2. make documentation consistent with existing documentation
3. fail with `GenerationException` if date/time types are specified, but cannot be loaded.


joelittlejohn:
Thanks @ctrimble. Some excellent housekeeping!

Issue #255, opened 2014-10-03
Allow JSON to be passed as a String to SchemaMapper (add sourceType support in #221)


diegosoaresub:
Hello,

In http://www.jsonschema2pojo.org/ I could convert JSON to POJO, but using the API this not work, only Json Schema to POJO. Is possible do this with the API?

In this case below, I didn't get exception but It doesn't generate the java file.

JCodeModel codeModel = new JCodeModel();
SchemaMapper schemaMapper = new SchemaMapper();
schemaMapper.generate(codeModel, className, packageName, "{\"titulo\": \"JSON\"}");
codeModel.build(new File("output"));

sanalsan:
Hello,
Is there any hope for this bug ?

joelittlejohn:
@sanalsan happy to accept a pull request for this if you are interested in fixing it.

sanalsan:
Hello Joe,

can you please tell me the error in it ?

package com.schoolspeak.modelgenerator;

import android.os.Bundle;
import android.os.Environment;
import android.support.v7.app.AppCompatActivity;

import com.sun.codemodel.JCodeModel;

import org.jsonschema2pojo.DefaultGenerationConfig;
import org.jsonschema2pojo.GenerationConfig;
import org.jsonschema2pojo.Jackson2Annotator;
import org.jsonschema2pojo.SchemaGenerator;
import org.jsonschema2pojo.SchemaMapper;
import org.jsonschema2pojo.SchemaStore;
import org.jsonschema2pojo.SourceType;
import org.jsonschema2pojo.rules.RuleFactory;

import java.io.File;
import java.io.IOException;

public class MainActivity extends AppCompatActivity {

    @Override
    protected void onCreate(Bundle savedInstanceState) {
        super.onCreate(savedInstanceState);
        setContentView(R.layout.activity_main);
        generateClassFromJsonString();
    }
    private File getFile(){
        File mPath = new
File(Environment.getExternalStorageDirectory(), "Model");
        if (!mPath.exists()) {
            try {
                mPath.mkdir();
            } catch (Exception e) {
                e.printStackTrace();
            }
        }
        return mPath;
    }
    private void generateClassFromJsonString(){
        JCodeModel codeModel = new JCodeModel();
        String jsonString="{ \"accounting\" : [   \n" +
                "                                { \"firstName\" :
\"John\",  \n" +
                "                                  \"lastName\"  : \"Doe\",\n" +
[...]

PR #475, opened 2015-12-28
Allow JSON to be passed as a String to SchemaMapper

raks81:
This is a proposed fix for #255 to allow example JSON to be passed to SchemaMapper as String. Currently it needs to be passed as an URL which is not always convenient. Fix #221 only added support for JSON schema. This PR adds support for example JSON as well.


jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/SchemaMapper.java
```diff
@@ -121,8 +121,17 @@ public JType generate(JCodeModel codeModel, String className, String packageName
         JPackage jpackage = codeModel._package(packageName);
 
         ObjectMapper mapper = new ObjectMapper();
-        JsonNode schemaNode = mapper.readTree(json);
-
+        JsonNode schemaNode = null;
+        switch (ruleFactory.getGenerationConfig().getSourceType()) {
+		case JSON:
+			JsonNode jsonNode = mapper.readTree(json);
+	        schemaNode = schemaGenerator.schemaFromExample(jsonNode);
```
joelittlejohn: Indentation problem here

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/SchemaMapper.java
```diff
@@ -121,8 +121,17 @@ public JType generate(JCodeModel codeModel, String className, String packageName
         JPackage jpackage = codeModel._package(packageName);
 
         ObjectMapper mapper = new ObjectMapper();
-        JsonNode schemaNode = mapper.readTree(json);
-
+        JsonNode schemaNode = null;
+        switch (ruleFactory.getGenerationConfig().getSourceType()) {
```
joelittlejohn: For this simple case, I think `if`/`else` would be neater.

joelittlejohn:
Thanks a lot for submitting this, I've made some minor review comments. If you fix these up, I will merge and release this in 0.4.19.

raks81:
@joelittlejohn I have fixed code as per your comments. Please review. Thanks!

joelittlejohn:
Thanks @raks81. One final step, please squash these commits into one.

raks81:
@joelittlejohn I have squashed the commmits. Please review. Thanks!

joelittlejohn:
:+1:

PR #476, opened 2015-12-30
Unique class name strategy

SergioModulusData:
## Problem

If a class already exists with the same name when generating we add a underscore to the name '_'. Using a weirdly formatted jsonSchema, [adp](https://developers.adp.com/wp-content/uploads/External/hireworker-v1-schema.json), I hit a limitation from the OS on the name length. 
## Attempted solution

Using a number with two underscore '__' instead of adding underscore to the class name when the class already exists. If the class already ends with a number and has two underscore we search for the number and add a number.

I did the modification to work with ADP crazy repeating schema. Could be useful for others.


joelittlejohn:
I like this.

It's quite a major change to the generated types (for some projects) but I think it's a good one. I think we should include this and bump to 0.5.0.

thachhoang:
@joelittlejohn should we include this in 1.0.0-alpha2?

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/util/MakeUniqueClassName.java
```diff
@@ -0,0 +1,42 @@
...
+public class MakeUniqueClassName {
+
+    /**
+     * When the class name is not unique we will use two underscore '__' and a digit representing the number of time
+     * this class was found
+     */
+    public static String makeUnique(String className) {
+        String returnClassName = className;
+        // Last character is a digit and there is 2 underscore in the className
+        if (Character.isDigit(className.charAt(className.length() - 1)) && StringUtils.contains(className, "__")) {
```
joelittlejohn: I think this check is a little too flexible. I think we should replace this with a regex like `.+\_\_(\d+)$`, and use the capture group to pull out the current number.

PR #482, opened 2016-01-08
includeDynamicAccessors default false

ctrimble:
Changes the default value of `includeDynamicAccessors` to `false`.


joelittlejohn:
Thanks for the quick update! :+1:

joelittlejohn:
I think we should also update this argument to `--include-dynamic-accessors`:

https://github.com/joelittlejohn/jsonschema2pojo/blob/ff0ad46407f9d0b70f6a074d14d3bd33dadf8f5e/jsonschema2pojo-cli/src/main/java/org/jsonschema2pojo/cli/Arguments.java#L148

This will be a minor breaking change, but I'd rather have this argument change than have inconsistency between the tools.

ctrimble:
Thought I might have missed one.  Let me see about getting that fixed.

PR #483, opened 2016-01-08
Change includeDynamicAccessors default to false in the CLI.

ctrimble:
This is a breaking change.

Removes the `--disable-dynamic-accessors`/`-dda` option and replaces it with an `--include-dynamic-accessors`/`-ida` option instead.


ctrimble:
This is related to #482

joelittlejohn:
:+1: :smiley_cat:

Issue #481, opened 2016-01-06
Generator uses first property name to create java class names instead of json type name


samuel-kerrien:
We have discovered that when defining reusable schema element it appears that the java class generated is named after the first field that uses this type.

To illustrate my point, here is a sample schema (file: example.json):

``` json
{
  "$schema": "http://json-schema.org/schema#",
  "type": "object",
  "properties": {
    "sex": {
      "$ref": "#/definitions/shared_class"
    },
    "job": {
      "$ref": "#/definitions/shared_class"
    }
  },
  "definitions": {
    "shared_class": {
      "type": "object",
      "properties": {
        "uri": {
          "type": "string"
        }
      }
    }
  }
}
```

note that the 'shared_class' definition is used twice in the schema.
The resulting generated class are as follow:

``` java
@JsonInclude(JsonInclude.Include.NON_NULL)
@Generated("org.jsonschema2pojo")
@JsonPropertyOrder({
    "uri"
})
public class Sex {

    @JsonProperty("uri")
    private String uri;
    protected final static Object NOT_FOUND_VALUE = new Object();

    ...
}
```

``` java
@JsonInclude(JsonInclude.Include.NON_NULL)
@Generated("org.jsonschema2pojo")
@JsonPropertyOrder({
    "sex",
    "job"
})
public class Example {

    @JsonProperty("sex")
    private Sex sex;

    @JsonProperty("job")
    private Sex job;
    protected final static Object NOT_FOUND_VALUE = new Object();

...
}
```

Note that if in the schema, the property 'job' had come first, the generated class would have been Job and not Sex.

I suppose the correct behaviour would be in the case where there is a $ref to use the name of the type rather than the property name. In this case the generated class would be 'SharedClass'.

Obviously, I can add use the 'javaType' property as follow to force that class to be generated but this also forces me to put java code reference into my schema that it doesn't feel right.

``` json
...
"shared_class": {
      "type": "object",
      "javaType": "com.company.example.SharedClass",
      "properties": {
        "uri": {
          "type": "string"
        }
      }
    }
...
```

dadrus:
Same issue applies while referencing external schema files. Here a slightly modified example from above, which introduces an `address_definition` element and externalizes the `job` definition. 
```.json
{
  "$schema": "http://json-schema.org/schema#",
  "type": "object",
  "properties": {
    "address_definition": { "$ref": "address.jsonschema" },
[...]

Issue #464, opened 2015-12-11
For objects in definition block use name as classname


wsalembi:
The following JSON schema (dummy.json) generates 3 classes: BillingAddress, ShippingAddress and Dummy.

```
{
  "$schema": "http://json-schema.org/draft-04/schema#",

  "definitions": {
    "address": {
      "type": "object",
      "properties": {
        "street_address": { "type": "string" },
        "city":           { "type": "string" },
        "state":          { "type": "string" }
      },
      "required": ["street_address", "city", "state"]
    }
  },

  "type": "object",

  "properties": {
    "billing_address": { "$ref": "#/definitions/address" },
    "shipping_address": { "$ref": "#/definitions/address" }
  }
}
```

Why not generate a single Address class and reuse it for the two properties?
I know it can be achieved by specifying javaType. But a generation option to enforce this rule globally would be really helpful.

dadrus:
#481 describes the same issue. A possible fix is referenced there as well.

Issue #392, opened 2015-07-21
Generating a POJO using oneOf, anyOf or allOf with constraints


CasperDB:
I'm generating a POJO using the jsonschema2pojo-maven-plugin and need to pass constraints in using oneOf, anyOf & allOf.

When I try the following code in a schema validator it works:
```
{
  "type": "object",
  "properties": {
    "name": {
      "description": "The advertised product",
      "type": "string"
    },
    "id" : {
      "oneOf": [
        { "type": "string", "maxLength": 5 },
        { "type": "number", "minimum": 0 }
      ]
    }
  }
}
```

However, the POJO after building has no constraints:
```
 @JsonProperty("id")
    private Object id;
```

I would like to see some restrictions annotated, is this not supported with the plugin yet?

joelittlejohn:
No, I'm afraid these are not supported. 

It would be useful if you could explain how you would expect this to work in this case? What Java would you like to see produced?

joelittlejohn:
(see also #91)

CasperDB:
I'm not sure what I want it to look like yet... My intention is to replace a jaxb object with a jackson object. Something similar to the xsd choice @XmlElement:

```XML
<xsd:element name="name" type="xsd:string" minOccurs="0"/>
            <xsd:choice>
                <xsd:element name="address" type="address"/>
                <xsd:element name="phone-number" type="phoneNumber"/>
                <xsd:element name="note" type="xsd:string"/>
            </xsd:choice>
```

```
private String name;
 
    @XmlElements(value = { 
            @XmlElement(name="address", 
                        type=Address.class),
            @XmlElement(name="phone-number", 
                        type=PhoneNumber.class),
            @XmlElement(name="note", 
                        type=String.class)
    })
```

I'm currently looking at Jackson annotations that might be useful.

samskiter:
Maybe you could blow oneOf out into several properties.

```java
 @JsonProperty("idString")
 @Nullable
    private String idString;

 @JsonProperty("idNumber")
 @Nullable
    private String idNumber;

  public void setIdString(String newIdString)
  {
    idNumber = null;
    idString = newIdString
  }
```
It would require a custom gson/jackson serialization/deserialization step for that part thought. It could be pretty generic and we could write it once into the jsonschema2pojo library.

amilinko:
Maybe **oneOf** rule can be implemented as class inheritance. In the example above, it could look like this:
```java
public abstract class id {
}

[...]

PR #487, opened 2016-01-08
Issue 481 fix

dadrus:
This fix solves the #464 and #481 issues. The implementation of the fix is done entirely in the PropertyRule class.

**IMPORTANT:** It will break the so far implemented naming convention of types. Instead of deriving the typename from the name of the first using property, or file name, the type name is now always based on its name from the reference (for internal references) or the actual file name for external references. Please see also my comments on #481 for more information.

**PLEASE SEE ALSO:** _TODO_s in the implementation of the `getPropertyTypeName` function (see diff view below).

Regards,
Dimitrij

### DIFF
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/PropertyRule.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/PropertyRule.java
@@ -18,6 +18,13 @@ package org.jsonschema2pojo.rules;
 
 import static org.apache.commons.lang3.StringUtils.*;
 
+import java.io.UnsupportedEncodingException;
+import java.net.MalformedURLException;
+import java.net.URI;
+import java.net.URL;
+import java.net.URLDecoder;
+
+import org.apache.commons.io.FilenameUtils;
 import org.jsonschema2pojo.Schema;
 
 import com.fasterxml.jackson.databind.JsonNode;
@@ -70,8 +77,9 @@ public class PropertyRule implements Rule<JDefinedClass, JDefinedClass> {
     public JDefinedClass apply(String nodeName, JsonNode node, JDefinedClass jclass, Schema schema) {
 
         String propertyName = ruleFactory.getNameHelper().getPropertyName(nodeName);
+        final String propertyTypeName = getPropertyTypeName(schema, node, nodeName);
 
-        JType propertyType = ruleFactory.getSchemaRule().apply(nodeName, node, jclass, schema);
+        JType propertyType = ruleFactory.getSchemaRule().apply(propertyTypeName, node, jclass, schema);
 
         node = resolveRefs(node, schema);
 
@@ -110,6 +118,39 @@ public class PropertyRule implements Rule<JDefinedClass, JDefinedClass> {
         }
 
         return jclass;
+    }
+
+    private String getPropertyTypeName(Schema schema, JsonNode node, String nodeName) {
+        if (node.has("$ref")) {
+            final String reference = node.get("$ref").asText();
+
+            if (reference.startsWith("#/")) {
+                // self reference with type definition
+                // use the name of the type
+                return reference.substring(reference.lastIndexOf('/') + 1);
+            } else {
+                // global reference (other file, url whatever)
+                final URI uri = schema == null || schema.getId() == null ? URI.create(reference) : schema.getId().resolve(reference);
+
+                try {
+                    return getNodeName(uri.toURL());
+                } catch (final MalformedURLException e) {
+                    throw new IllegalArgumentException(String.format("The referenced URL %s for %s is invalid", uri, nodeName), e);
+                }
+            }
+        } else {
+            return nodeName;
+        }
+
+    }
+
+    private String getNodeName(URL url) {
+        try {
+            return FilenameUtils.getBaseName(URLDecoder.decode(url.toString(), "UTF-8"));
+        } catch (final UnsupportedEncodingException e) {
+            throw new IllegalArgumentException(String.format("Unable to generate node name from URL: %s", url.toString()), e);
+        }
+
     }
 
     private void propertyAnnotations(String nodeName, JsonNode node, Schema schema, JDocCommentable generatedJavaConstruct) {
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/PropertyRule.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/PropertyRule.java
@@ -25,7 +25,9 @@ import java.net.URL;
 import java.net.URLDecoder;
 
 import org.apache.commons.io.FilenameUtils;
+import org.apache.commons.lang3.StringUtils;
 import org.jsonschema2pojo.Schema;
+import org.jsonschema2pojo.util.URLUtil;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.sun.codemodel.JBlock;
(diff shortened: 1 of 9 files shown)


dadrus:
To make it better accessible, I copy the notes/questions from the mentioned *TODO*s here. In all cases they are related to the `id` of a schema, which yet is not supported.

- The first is about the nested references to self? How to handle them: e.g: `"$ref": "#/definitions/level/sublevel/actual_element"` with type information available only on `actual_element` and `level/sublevel` used as namespaces to avoid type conflicts e.g. level could also contain an `actual_element` definition which differs from the `sublevel/actual_element`. Depending on whether a `level/sublevel` element defines a type or is just used for structuring purposes, it may be worth to generate either a java class with a nested static class or java subpackages for such cases. This rule however seems to be not applicable if our `actual_element` contains an `id`, which defines the namespace of it, thus can be mapped to a java package.
- The second is about external references, e.g. `"$ref": "my_type.json"`. The uri may by any valid URL (file, classpath, java, http, etc). The url may point to a schema with a type name already defined e.g. in the document referencing it, but with different contents. How to handle this? The usage of `id` can help to overcome this ambiguity.

What does the comunity think about it?

ctrimble:
I really like the idea of the schema's URI being directly related to the type name.  Here are some of my thoughts on this:

1. It would definitely need to be something for `v0.5.0` or later.
2. I would really like to see a one to one relationship between schema files and Java source files.
3. Nested schemas should be rendered as nested static classes, since this would eliminate namespace conflicts when generating the types.
4. I wonder if something could be done to provide a strategy for this.  This kind of change could really cause problems for people.

dadrus:
I've already prepared something which goes into this direction. The 3a1a96de7bf5dc6dcf2904badbb06e7cd5612e8d commit, referenced in the #481 issue includes, let's say the first drop for this. Since the JSON schema file can define multiple types, like an XML schema does, I personally see the one two one relationship between not a file, but a schema for a type definition and a Java source file. That means, that from my understanding, the id defines a namespace for a type, which corresponds to a package in Java. 

Here a basic example. Let name this file `job.json`
```.json
{
    "id": "http://test.com/2016/job.json"
    "$schema": "http://json-schema.org/schema#",
    "description": "A Job definition",
    "type": "object",
    "properties": {
        "title": { "type": "string" }
    },
    "required": ["title"]
}
```

Without an id, the corresponding Java class is generated into the *"default"* package. With the *id* present, like in the example above, the class will be generated into the *com.test._2016.job* package. In the showed case the schema file contains just one type definition - of the *job*. Let me make the example more complex and add another type to the *job.json* file:

```.json
{
    "id": "http://test.com/2016/job.json"
    "$schema": "http://json-schema.org/schema#",
    "description": "A Job definition",

    "definitions": {
        "description": {
             "type": "object",
             "properties": {
                 "title": { "type": "string" }
            }
        }
    },

    "type": "object",
    "properties": {
        "title": { "$ref": "#/definitions/description" }
    },
    "required": ["title"]
}
```

Now the schema file defines two type, and thus two schemas, one for the *job* and one for *description*. Since the *description* schema does not define its own *id*, it belongs to namespace defined by the top level *id*. You say, a nested schema, like here, should be rendered as a nested static class. Why should it be nested? It has its own name *description* and having an (indirect) *id* it has its namespace. My understanding may be wrong, but as long as a type definition is not anonymous (defined inplace as a type of a property, thus does not have a name), it can be a standalone Java type, regardless whether the type defines an enum, or a class. I see it as a design error, if a namespace (defined by an *id*) defines multiple types (json schemas) with same name. If there is a need for same type name - just give it an *id*. You do it while defining an XML schema as well, right?

That also means, that not the location (folder structure) makes the types defined in the corresponding files unique, but the *id* given for each type. The aforementioned commit implements exactly this: As long as a type is referenced using *"$ref"*, it is generated as a standalone Java type. I've tried to describe the behavior in the comments for the #481 issue.

I may miss important things, but to be honest, I encountered all those issues, I've linked to #481, and was looking for a fast solution, which would satisfy my needs. The current result is available with 3a1a96de7bf5dc6dcf2904badbb06e7cd5612e8d. I do not expect everybody to accept it is as is, but it could be a starting point, or even just a source of ideas.

dadrus:
Maybe it becomes more obvious, if a schema file does not define a top level type. Here is modified example from http://json-schema.org/latest/json-schema-core.html (I just added the `"type": "object"`):

```.json
"schema1": {
    "id": "schema.json",
    "type": "object"
},
"schema2": {
    "id": "otherschema.json",
    "type": "object"
    "nested": {
        "id": "#bar",
        "type": "object"
    },
    "alsonested": {
        "id": "t/inner.json#a",
        "type": "object"
    }
},
"schema3": {
    "id": "some://where.else/completely#",
    "type": "object"
}
```

`schema1`, `schema2` and `schema3` all define a standalone type. Each of them should correspond to a generated Java class. You see, here the type can't be derived from the file name. The `nested` and `alsonested` should be generated as nested static classes in the class generated for `schema2`.

ctrimble:
The only reason I suggested inner static classes is that they work like a package, for name spacing purposes, and will prevent collisions at the file system level.  Again, this is just a suggestion.

This is what I would generate for the example you gave, assuming it is in `schema.json`:

```
public class Schema {
  ...
  public static class Schema1 {
    ...
    public static class Nested { ... } 
    public static class AlsoNested { ... }
  }
  public static class Schema2 { ... }
  public static class Schema3 { ... }
}
```

That would prevent files from colliding, in the case where conflicting names are present in neighboring schemas.  Also, it seems to me that mimicking the schema structure in the java types would lead to the least amount of surprise for those using the types.

dadrus:
That's true. This is an other approach to prevent collisions. You last sentense is also very true. Going away from java types to packages would indeed lead to a lot of surprizes for people used to the behavior of the current implementation.

BTW. The schema example above is not supported by the current implementation. At least nothing was generated given the code I have in my *schema_id_support* branch. Maybe I have broken something, which was previously possible. Anyway I've recently enabled the generation of types from such schema files, which given the mentioned above *id* support would result in the following generated output:

```.java
package schema;
public class Schema1 { ... }
```

```.java
package otherschema;
public class Schema2 { ... }
```

```.java
package else.where.completely;
public class Schema3 { ... }
```

```.java
package bar;
public class Nested { ... } 
```

```.java
package t.inner
public class AlsoNested { ... }
```

If however the above schema is changed to:

```.json
"schema1": {
    "id": "schema.json",
    "type": "object"
},
"schema2": {
    "id": "otherschema.json",
    "type": "object"
    "properties": {
        "nested": {
            "type": "object"
        },
        "alsonested": {
            "type": "object"
        }
    }
},
"schema3": {
    "id": "some://where.else/completely#",
    "type": "object"
}
```

follwing is generated:
```.java
package schema;
public class Schema1 { ... }
```

```.java
package otherschema;
public class Schema2 { 
    public static class Nested { ... }
    public static class AlsoNested { ... }
}
```

```.java
package else.where.completely;
public class Schema3 { ... }
```

The bullet point 4 from your first comment is definitely very important. Both in sense of the description of the expected behaviour for the implementation, and also to inform everybody.

I've not pushed it yet, since all my work is need driven and the plugin is in my case a tool. Because of this I've not implemented/updated unit/integration tests. If you're interested in the details I can push it with given testability limitations. 

An other thing I was looking for is the the *oneOf* support. I see a lot of discussions here (#391 is very prominent). Do you know whether somebody has already done something in this direction (I mean at least basic code)?

Regards,
Dimitrij

dadrus:
soory, I pushed the wrong button

ctrimble:
I would take the `oneOf` discussion over to #392 and have a look at the network graph to see what commits may be out there.

sveba:
Is there any chance to have this fix in master soon? And also have 0.4.19 released after that :)

joelittlejohn:
@sveba from the comments and conversation so far, I don't believe this fix is complete and ready to merge. It's unlikely that this will appear in the next release.

kiaSso:
Hi guys, i downloaded the verion 0.4.23 and this fix doesn't appear to be there.
How can I do to get only this fix without affect the rest of the code?
Is there a planned release soon?

Many thanks

DamageShadow:
Will it be merged anytime soon?

PeteTh:
We too would like a fix for this

mpern:
Any ETA on merging this?

PeteTh:
Can we get this merged ? all seem happy with the fix, it's been over a year now.

joelittlejohn:
This change isn't ready to merge.

It breaks compatibility for existing users of generated types in a major way and still has a number of unresolved issues that make it unsuitable for inclusion in the core.

Feel free to merge this into your own custom version.

PR #489, opened 2016-01-10
Issue 485 fix

dadrus:
`UrlConverterTest.urlIsCreatedFromFilePath` uses an absolute *ux stype path. On Windows the Java FileSystem implementation converts it to a valid Windows path, that is, a `C:` is prepended to it. If instead a relative path is used by the test, the returned URL will contain the path to the project as well. This way in both cases we have to check whether the returned URL ends with the expected path, which is done by this commit. It fixes the #485 issue


joelittlejohn:
:+1:

Issue #484, opened 2016-01-08
ArgumentsTest fails under Windows


dadrus:
The given test uses *ux specifics and does not work on Windows

joelittlejohn:
Fixed by #490

PR #490, opened 2016-01-10
Issue 484 fix

dadrus:
ArgumentsTest test use an absolute *ux stype path. On Windows the Java FileSystem implementation converts it to a valid Windows path, that is, a C: is prepended to it. If instead a relative path is used by the test, the returned URL will contain the path to the project as well. This way in both cases we have to check whether the returned URL ends with the expected path, which is done by this commit. It fixes the #484 issue.
Actually the reason for the issue is absolutely the same as for the issue #485.

### DIFF
--- a/jsonschema2pojo-cli/src/test/java/org/jsonschema2pojo/cli/ArgumentsTest.java
+++ b/jsonschema2pojo-cli/src/test/java/org/jsonschema2pojo/cli/ArgumentsTest.java
@@ -58,7 +58,7 @@ public class ArgumentsTest {
         });
 
         assertThat(args.didExit(), is(false));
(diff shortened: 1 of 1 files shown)


dadrus:
Actually, I think, it would be better not to use direct paths at all, but create corresponding Files/Folders using the `TemporaryFolder` JUnit Rule and check the expectations based on the `File` representation.
This way there will be no OS FS dependencies at all. Same applies for the fix of the #485 issue.

joelittlejohn:
I think the approach you propose is definitely preferable. I'm going to merge your changes here as a first step and we can look to improve this more permanently/cleanly in future.

PR #495, opened 2016-01-24
Android gradle plugin fixes

ramib85:
Hi,
I'm an Android developer,
I use this project on a daily basis by utilizing the cli during our project's build process.
I tried to use the gradle plugin but unfortunately it didn't compile correctly, although the source code was generated correctly, it seems that the generated code isn't added to the source set of the project correctly.
I'm using several kinds of build variants which are essentially different build setups, when using this advanced configuration, the generated code must be added to each build variant's source set.
Although the gradle plugin's code does something like that, it doesn't actually compile when the project's code depends on the generated code.
I've reproduced compilation errors on the android example code projects in the gradle plugin's source code.

In this pull request, I offer a solution that works, I've reimplemented how the generated code is being added to the build variant's source set. I've based my implementation on other gradle plugins for Android.
Such as: 
https://github.com/Flipboard/psync/blob/master/psync/src/main/groovy/com/flipboard/psync/PSyncPlugin.groovy

Commit notes:
Added Support for code generation for different build variants of Android project.
Fixed : adding the generated code to each build variant source set correctly, so it can be picked up by the compiler.
Added example usage of a generated class into the Android app and library classes.
Added correct compile time dependencies to the example projects such that the generated code could compile successfully.
Updated Android SDK and gradle versions.

### DIFF
--- a/jsonschema2pojo-gradle-plugin/example/android/app/build.gradle
+++ b/jsonschema2pojo-gradle-plugin/example/android/app/build.gradle
@@ -12,13 +12,13 @@ buildscript {
 }
 
 android {
-    compileSdkVersion 21
-    buildToolsVersion "21.1.2"
+    compileSdkVersion 23
+    buildToolsVersion "23.0.2"
 
     defaultConfig {
         applicationId "jsonschema2pojo.joelittlejohn.github.com.androidexample"
         minSdkVersion 17
-        targetSdkVersion 21
+        targetSdkVersion 23
         versionCode 1
         versionName "1.0"
     }
@@ -29,11 +29,26 @@ android {
         }
     }
 
+    packagingOptions {
+        exclude 'META-INF/LICENSE.txt'
+    }
+
 }
 
 dependencies {
     compile fileTree(dir: 'libs', include: ['*.jar'])
-    compile 'com.android.support:appcompat-v7:21.0.3'
+    compile 'com.android.support:appcompat-v7:23.1.1'
+
+    // Required for @Generated annotation
+    compile 'org.glassfish:javax.annotation:10.0-b28'
+    // Required if generating Gson annotations
+    compile 'com.google.code.gson:gson:2.4'
+    // Required if generating equals, hashCode, or toString methods
+    compile 'commons-lang:commons-lang:2.6'
+    // Required if generating JSR-303 annotations
+    compile 'javax.validation:validation-api:1.1.0.CR2'
+    // Required if generating Jackson 2 annotations
+    compile 'com.fasterxml.jackson.core:jackson-databind:2.1.4'
 
 }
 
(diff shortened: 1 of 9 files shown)


joelittlejohn:
Thanks. It looks like you have made some very good additions and adjustments here.

I don't use Gradle or develop for Android myself, so it would be useful if others that do could try these changes. Do you foresee any issues for people that use Gradle but not for Android development? Is there anything added here that they would not want?

I'll take a look through the code and add some comments.

jsonschema2pojo-gradle-plugin/example/android/app/build.gradle
```diff
@@ -29,11 +29,26 @@ android {
...
+
 }
 
 dependencies {
     compile fileTree(dir: 'libs', include: ['*.jar'])
-    compile 'com.android.support:appcompat-v7:21.0.3'
+    compile 'com.android.support:appcompat-v7:23.1.1'
+
+    // Required for @Generated annotation
+    compile 'org.glassfish:javax.annotation:10.0-b28'
```
joelittlejohn: I don't understand why this is required. @Generated is part of Java's core: http://docs.oracle.com/javase/6/docs/api/javax/annotation/Generated.html

joelittlejohn: Ah, maybe this is an Android only problem?

ramib85: Android doesn't contain the full Java language and is missing @Generated annotation.
I've added it only to the Android example project.

It would be nice if annotationStyle=none would take care not to create @Generated annotation.  Or if there was another parameter to control if @Generated would be added to the generated class.

ramib85:
Regarding usage for non-Android development,
I've taken care to create a separate gradle task that will be used for Android project.
The Gradle plugin will call a different code for Java and Android projects.

ramib85:
can the pull request be merged if there aren't any standing issues?

joelittlejohn:
I don't see a problem with merging and releasing this. Is there anything that Gradle users should know on upgrading to this release? Either the Android or non-Android users?

ramib85:
For Non-Android users, there is no change.
For Android users, the plugin will ignore _targetDirectory _  and will generate the classes into a generated source folder for each build variant and add this directory to the java source set, so the compiler will find and compile the newly generated source files.
This change was made to be compliant with Android's source folder structure and to support different build variants.

joelittlejohn:
Instead of ignoring the `targetDirectory` for Android projects, I think the different source folders should go into the specified `targetDirectory`. At present the various source folders will go into a hard coded location of  `$project.buildDir/generated/source/js2p/$variant.flavorName/$variant.buildType.name/` and unfortunately I don't think this is acceptable since people will want to have control over where these sources go.

ramib85:
I've tried to implement saving the source folders under the specified targetDirectory.
Although the files are generated correctly, they aren't added to the java source sets of the build variant, which causes the compilation of the project to fail because those files were not compiled.

Android's gradle build seems to look for build variant generated source files only at a very specific path, 
`$project.buildDir/generated/source/js2p/$variant.flavorName/$variant.buildType.name/`
Using this path is compliant with other files that Android build generates.
Putting the generated files under the `$project.build` path has the advantage of being ignored by version control and cleaned when running 'clean' on the project. 

In my company's Android projects, we compile against the generated code and it doesn't really matter to the developers where those files are saved.

joelittlejohn:
@ramib85 what I mean is that you still add these directories to the classpath as part of your pull request, but instead of using `$project.buildDir/generated/source/js2p/$variant.flavorName/$variant.buildType.name/` you would use `$targetDirectory/$variant.flavorName/$variant.buildType.name/`

Typically the targetDirectory will be inside the project build dir, in fact you can see in the readme we suggest:

`${project.buildDir}/generated-sources/js2p`

ramib85:
As far as I know on Android projects directory structure, 
The android gradle build expects this pattern for generated code,
only the name of the **XXXX** directory is changeable, this path is needed for adding the generated files to the source set.
$project.buildDir/generated/source/**XXXX**/$variant.flavorName/$variant.buildType.name/

Example of my project structure:
http://i.imgur.com/kTHJ04y.png

joelittlejohn:
Thanks, I'll defer to your superior knowledge on how Gradle builds for Android work!

PR #499, opened 2016-02-02
Add javaName schema property

SpaceBison:
This PR adds an option for defining custom field names instead of the ones inferred from property names. For example, this:
```json
{
  "type": "object",
  "properties": {
    "a": {
      "javaName": "b",
      "type": "string"
    }
  }
}
```
will generate this:
```java
@Generated("org.jsonschema2pojo")
@JsonPropertyOrder({
    "a"
})
public class SimpleTest {

    /**
     * 
     * Corresponds to the "a" property.
     * 
     */
    @JsonProperty("a")
    private String b;

    /**
     * 
     * Corresponds to the "a" property.
     * 
     * @return
     *     The b
     */
    @JsonProperty("a")
    public String getB() {
        return b;
    }

    /**
     * 
     * Corresponds to the "a" property.
     * 
     * @param b
     *     The b
     */
    @JsonProperty("a")
    public void setB(String b) {
        this.b = b;
    }
    /* ... */
}
```

### DIFF
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/JavaNameRule.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/JavaNameRule.java
@@ -1,0 +1,38 @@
+/**
+ * Copyright © 2010-2014 Nokia
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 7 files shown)


joelittlejohn:
A nice and very useful addition I think. I'll take a look through the code
and comment on anything I think needs to change. Thanks for submitting this!

PR #501, opened 2016-02-03
Remove unused suppress warning annotations

### LINKED ISSUE #498: Unnecessary @SuppressWarnings java warnings after upgrade
After upgrading to version 0.4.18 we've started receiving a load of Java warnings like so:

Unnecessary @SuppressWarnings("unchecked")

ctrimble:
This PR adds an integration test to detect compilation warnings in the source files generated by the project.  Warnings being produced by the includeDynamicAccessors feature have also been resolved.

__Note__: there is currently a bug with the Eclipse compiler that causes it to report suppressed warnings.  In those cases, the warnings being suppressed have been allowed.

Fixes #498

### DIFF
--- a/jsonschema2pojo-integration-tests/src/test/java/org/jsonschema2pojo/integration/CompilerWarningIT.java
+++ b/jsonschema2pojo-integration-tests/src/test/java/org/jsonschema2pojo/integration/CompilerWarningIT.java
@@ -1,0 +1,129 @@
+/**
+ * Copyright ¬© 2010-2014 Nokia
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jsonschema2pojo.integration;
+
+import java.io.File;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+
+import org.apache.commons.io.output.NullWriter;
+import org.hamcrest.FeatureMatcher;
+import org.hamcrest.Matcher;
+import org.hamcrest.Matchers;
+import org.jsonschema2pojo.integration.util.Compiler;
+import org.jsonschema2pojo.integration.util.Jsonschema2PojoRule;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameters;
+
+import javax.tools.Diagnostic;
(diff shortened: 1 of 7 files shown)


jsonschema2pojo-integration-tests/src/test/java/org/jsonschema2pojo/integration/CompilerWarningIT.java
```diff
@@ -0,0 +1,128 @@
...
+
+  public static Matcher<Iterable<Diagnostic<? extends JavaFileObject>>> onlyCastExceptions() {
+    return Matchers.everyItem(hasMessage(containsString("Type safety: Unchecked cast from")));
+  }
+
+  public static Matcher<Diagnostic<? extends JavaFileObject>> hasMessage( Matcher<String> messageMatcher ) {
+    return new FeatureMatcher<Diagnostic<? extends JavaFileObject>, String>(messageMatcher, "message", "message") {
+      @Override
+      protected String featureValueOf(Diagnostic<? extends JavaFileObject> actual) {
+        return actual.getMessage(Locale.getDefault());
```
ctrimble: This is probably broken.  It could cause the project to test differently in different locales.

ctrimble:
@joelittlejohn I have found one potential problem with this change.  Do you have any notes before I do a round of updates?

jsonschema2pojo-integration-tests/src/test/java/org/jsonschema2pojo/integration/DynamicPropertiesIT.java
```diff
@@ -229,7 +229,7 @@ public void setDeclaredPropertyTest(Map<String, Object> config, String schemaLoc
             throw e.getCause();
         }
 
-        assertThat("set$ for field " + fieldName + " of type " + fieldType + " works.",
+        assertThat("set for field " + fieldName + " of type " + fieldType + " works.",
```
ctrimble: This change is just cleanup.  I will move it to master.

joelittlejohn:
@ctrimble I think it would be good to split the fix for #498 into its own PR.

If you're struggling to get this to work well (cross-platform) then I must admit, this seems like a lot of tricky work for not much gain. I realise the risk of compiler warnings is always there, but its very low and doesn't have severe consequences. If you find this is proving hard to implement in a fast and portable way, then I'm happy to just fix #498 and move on.

ctrimble:
I am not struggling at all with this.  If you checkout the commit with the test, it shows the problems found in #498.  Moving to the head of the PR shows that the issue is resolved.  I wish the Eclipse compiler would respect those @ SuppressWarnings annotations when used through the tooling API, but at least the unused warnings are being generated.

If you think that the testing changes make the test suite worse, I would be happy to make changes, or just drop the test all together.  Otherwise, I think this is complete.

joelittlejohn:
Ah sorry, maybe I misinterpreted one of your comments in the source code. I got the impression that this was proving difficult to implement. I'm happy to merge this if it's ready to go in.

One thing that I couldn't quite work out from your changes, what's the reason for introducing the CompilerSupplier and passing this, instead of simply passing a JavaCompiler? Would it be possible to simply have `systemJavaCompiler()` and `eclipseCompiler()` return the JavaCompiler?

ctrimble:
Honestly, I am so used to the Supplier<T> interface now, that was more just habit than anything else.  The code would probably be clearer without that level of indirection.  I will take it out.

ctrimble:
CompilerSupplier has been removed.

ctrimble:
This is ready to merge.

joelittlejohn:
:+1:

PR #504, opened 2016-02-07
'oneOf' support

ctrimble:
An implementation of 'oneOf' using generated deserializers.  This change is no where near complete, but it does show a way that support could be added for this operator.  Any input on this would be appreciated.

Current issues:

- Annotators do not have access to the RuleFactory or current Schema.  These were added to the method for annotating fields.
- Widens the responsibility of annotators, since code is being generated in the annotator now.
- No support for objects.
- No support for validating that value actually matches an option's schema.
- What to do if none match?


ctrimble:
This now works for `string` and `integer` with `Jackson2` and `Gson`.

ctrimble:
@joelittlejohn I am curios about your thoughts on this.  I am not happy with the way this is structured yet, but it does show that I am going to end up basically writing a validator to support `oneOf`.  Is this something the project wants to take on?

joelittlejohn:
@ctrimble So far, I have been very, very reticent to implement (or get close to implementing) a validator as part of this project. It's not the goal of this project to validate data, only to help Java engineers create a structure that can hold their data with minimal fuss. JSR-303/349 was a quick win, but again, it's just annotations (more metadata for the Java world) and _not_ a validator.

I'd be loathe to go down this route as I think validation is best handled by another library. Integrating this would also put a big burden on the maintenance of this project and make its goals less clear. It concerns me when libraries sprawl too far, and I think validation is best done at the raw JSON level by a schema validation library.

What do you think?

ctrimble:
I am also hesitant to get validation involved in the project.  This is why I took pause after implementing these first two types.  It became clear that I would end up with a validator, or at least something close.  

Unfortunately, validation is at the heart of the JSON Schema 'oneOf' key word.  Because of it's definition, binding and validation are linked.  It seems like users are running head long into this.  I have been avoiding it, but I will be forced to deal with it at some point.

I had considered trying to leverage the JSR-303/349, since this would bring some consistence to things.  After considering it, there were some real issues.  We need to validate before we bind, not the other way around.  Also, having to inject validators into the deserialization pipelines may make using the project to difficult.

As far as an external validation library goes, it will definitely need to work on the raw JSON form for both Gson and Jackson.  I am only aware of one option for Java, and that is fge's [json-schema-validator](https://github.com/fge/json-schema-validator).  That project is only draft 3 and seems to have stagnated.  Perhaps the correct thing to do is create a separate validator and leverage it from jsonschema2pojo.

ctrimble:
As a stop gap, should a PR get created that just bails out to a generic type when a `oneOf`, `anyOf` or `allOf` is encountered?

joelittlejohn:
I think exploring your stop-gap solution of using a 'generic' type is the best option. It may also make sense to this only in the limited cases where Jackson's polymorphic deserialization can be leveraged (i.e. the cases where some explicit indicator can be added to the JSON).

ctrimble:
Closing since this is not the direction the project wants to go with oneOf support.

samskiter:
Just read the end of this discussion. Regarding validation - including that into the project would be a bonus as far as I'm concerned. But I see you point that it might extend the responsibilities too far - especially when dealing with several different deserializers.

If you look at protobufs, they are effectively doing both and a JSON system that behaves the same way is something I'd want available to me

fge's project is ok, but it's *very* large - it increased an Android app by about 2.7MB! It does support draft4 though - you might have missed than.

regarding oneOf, I agree that it really is tied up with validation - I'm becoming more tempted to suggest it is removed from draft v5. Almost exactly the same behaviour can be accomplished by using named properties for each of the schemas in your oneOf and also using maxProperties/minProperties = 1.

e.g.:

```
{
    oneOf : [
        $ref : schema1,
        $ref : schema2,
        $ref : schema3
    ]
}
```

becomes:

```
{
    properties:
    {
        firstThing: { $ref : schema1 },
        secondThing: { $ref : schema2 },
        thirdThing: { $ref : schema3 }
    },
    maxProperties : 1,
    minProperties : 1
}
```
This is actually how I would prefer it was equated to in this library, but it would again require validation.

PR #513, opened 2016-02-23
Jsonschema2pojoRule report problems to STDERR

ctrimble:
Capturing diagnostics in the Jsonschema2pojoRule was removing output from STDERR.  This makes that feature optional and adds a test to make sure STDERR output is generated when there are compilation problems.


joelittlejohn:
Great, thanks!!

PR #528, opened 2016-03-13
Fix javaName breaking required properties

SpaceBison:
The previous implementation of `javaName` rule was causing required properties to not be annotated properly (and probably breaking other things). This PR is intended to fix that by better integrating the feature into the project's structure.

`javaName` is applied now within the `NameHelper` class so that the other modules can take it into account while resolving generated Java fields. This requires passing the `JsonNode` of the relevant property to the helper method alongside the property's name. Some of the other modules have been altered to reflect that.


joelittlejohn:
Great, thanks! :+1:

Issue #309, opened 2015-03-13
Make the fields used in hashcode and equals generation configurable


d4rxh4wx:
https://github.com/joelittlejohn/jsonschema2pojo/pull/241#issuecomment-58324684

ltoinel:
+1

joelittlejohn:
Hi all, how would you propose configuring this? Would you like to add a field to each property, have a global configuration value containing a list of field names, or do something like include only 'required' fields?

d4rxh4wx:
Hi. I would go with a list of field names. In that way, you can see in one place which fields are participating to equals and hashcode.

joelittlejohn:
@d4rxh4wx do you mean in the object itself? like:

```json
{
  "type" : "object",
  "javaHashcodeEqualsFields" : ["foo","bar"]
}
```

d4rxh4wx:
Yes. That sounds good.

mikaelec:
+1

How about an option for using super classes' fields for equals and hashCode?

skarian92:
Hi
So any updates on the problem.How can I configure HashcodeEqualFields??

rozhok:
+1 I don't like to include dependency to commons-lang and want to see plain java8 equals/hashCode like 
```java
@Override
public boolean equals(Object o) {
    if (this == o) {
        return true;
    }
    if (o == null || getClass() != o.getClass()) {
        return false;
    }
    Message that = (Message) o;
    return Objects.equals(type, that.type) &&
            Arrays.equals(updated, that.updated) &&
            Arrays.equals(deleted, that.deleted) &&
            Objects.equals(content, that.content);
}
@Override
public int hashCode() {
    return Objects.hash(type, updated, deleted, content);
}
```

szhem:
+1 

I'd like hashcode, equals and toString with java8 style too

joelittlejohn:
We now have equals, hashCode and toString methods that do not depend on commons-lang, and we also allow fields to be excluded from these (see `excludedFromEqualsAndHashCode` schema extension).

PR #530, opened 2016-03-14
#309 - Make the fields used in hashcode and equals generation configurable

SpaceBison:
Added `excludedFromEqualsAndHashCode` option for specifying which properties should be excluded from `equals()` and `hashCode()` methods. It can be used both as an object-level array containing excluded property names or as a property-level boolean switch.

Example:

```
{
  "type": "object",
  "excludedFromEqualsAndHashCode" : [ "excludedByArray" ],
  "properties": {
    "notExcluded" : {
      "type" : "string"
    },
    "excludedByProperty" : {
      "type" : "string",
      "excludedFromEqualsAndHashCode" : true
    },
    "notExcludedByProperty" : {
      "type" : "string",
      "excludedFromEqualsAndHashCode" : false
    },
    "excludedByArray" : {
      "type" : "string"
    }
  }
}
```

`notExcluded` and `notExcludedByProperty` will be used while generating `hashCode()` and `equals()`, while `excludedByProperty` and `excludedByArray` won't.

I felt that excluding properties from the methods would be more fitting than listing the included ones (as discussed in issue #309), because that way the default behavior, which is including all properties in the methods, is modified rather than redefined.

### DIFF
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/ObjectRule.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/ObjectRule.java
@@ -16,27 +16,7 @@
 
 package org.jsonschema2pojo.rules;
 
-import static org.apache.commons.lang3.StringUtils.*;
-import static org.jsonschema2pojo.rules.PrimitiveTypes.*;
-import static org.jsonschema2pojo.util.TypeUtil.*;
-
-import java.lang.reflect.Modifier;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.HashMap;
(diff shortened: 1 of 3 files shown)


anuragpro:
Is this feature getting merged?

SpaceBison:
Took me long enough, but I updated this PR at last.

SpaceBison:
Merged in [5eeb01e](https://github.com/joelittlejohn/jsonschema2pojo/commit/5eeb01e1f84f6f61d63eb74ae6bbfeabbbd2acf2), now it builds. :)

joelittlejohn:
Yeah. It looks like Travis changed their base image and broke installing of Oracle JDKs.

PR #536, opened 2016-03-21
Convert line endings from CRLF to LF.

johnjohndoe:
+ This commit introduces no code change!
+ Please do not commit files with Windows line endings (CRLF) since this
  causes a permanent **full-file-diff** on Unix machines. The Git config
  can be setup to automatically convert line endings back and forth on a
  Windows machine.

---

> Formatting and whitespace issues are some of the more frustrating and subtle problems that many developers encounter when collaborating, especially cross-platform. It’s very easy for patches or other collaborated work to introduce subtle whitespace changes because editors silently introduce them, and if your files ever touch a Windows system, their line endings might be replaced. Git has a few configuration options to help with these issues.

> `core.autocrlf`

> If you’re programming on Windows and working with people who are not (or vice-versa), you’ll probably run into line-ending issues at some point. This is because Windows uses both a carriage-return character and a linefeed character for newlines in its files, whereas Mac and Linux systems use only the linefeed character. This is a subtle but incredibly annoying fact of cross-platform work; many editors on Windows silently replace existing LF-style line endings with CRLF, or insert both line-ending characters when the user hits the enter key.

> Git can handle this by auto-converting CRLF line endings into LF when you add a file to the index, and vice versa when it checks out code onto your filesystem ...

Source: [Git documentation - Configuration - Formatting-and-Whitespace](https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#Formatting-and-Whitespace)


joelittlejohn:
@johnjohndoe the files that have Windows line endings are .bat files, they're the Windows version of some scripts and are required to be understood by Windows tooling when we release them.

I think the experience will be broken for some users if we change the line endings here. But do you think differently?

johnjohndoe:
As stated above the repository for multi-platform project should not mix line-endings. Git can handle this for you:

> Git can handle this by auto-converting CRLF line endings into LF when you add a file to the index, and vice versa when it checks out code onto your filesystem ...

joelittlejohn:
@johnjohndoe This is a local configuration option though, right?

We need to _release_ files with Windows line endings. This isn't about mixing line endings of different styles in a single code base, or having people edit in their own preferred line-ending style, it's about creating and releasing correctly formatted bat files.

johnjohndoe:
Yes, the configuration is on behalf of the client.
The release process could take care of the conversion as Git does with clone, checkout, ...

joelittlejohn:
I'm not going to merge this as I prefer to have these scripts runnable on permanently runnable Windows and not modified by the release process.

johnjohndoe:
I like to add that you should be aware that having files showing up as changed all the time makes it harder for less experienced developers to contribute. Each time one clones the repository, checks out a branch or pull changes they will have to deal with the full file diff of these files getting in the way.

PR #539, opened 2016-03-26
Override parent builders even if the child schema has no properties

### LINKED ISSUE #538: Overridden builders are not added to empty child class
Whenever child class doesn't add any properties but just inherits from parent class it will be generated completely empty even though parent class has many properties with builders.

Example
address.json:
```json
{
  "type": "object",
  "javaType": "Address",
  "properties": {
    "city": { "type": "string" },
    "street": { "type": "string" }
  }
}
```
address2.json:
```json
{
  "type": "object",
  "extends": {
    "$ref": "address.json"
  }
}
```
Address2.java:
```java
@Generated("org.jsonschema2pojo")
public class Address2 extends Address
{
    @Override
    public String toString() {
        return ToStringBuilder.reflectionToString(this);
    }
}
```
Address2 should include builder methods from Address:
```java
    @Override
    public Adderess2 withCity(String city) {
        super.withCity(city);
        return this;
    }

    @Override
    public Adderess2 withStreet(String street) {
        super.withStreet(street);
        return this;
    }
```

Everything works as expected when I create address3 which has one additional property:
address3.json:
```json
{
  "type": "object",
  "extends": {
    "$ref": "address.json"
  },
  "properties": {
    "country": { "type": "string" }
  }
}
```

thachhoang:
Fix #538.

### DIFF
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/ObjectRule.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/ObjectRule.java
@@ -121,9 +121,7 @@ public class ObjectRule implements Rule<JPackage, JType> {
             ruleFactory.getDescriptionRule().apply(nodeName, node.get("description"), jclass, schema);
         }
 
-        if (node.has("properties")) {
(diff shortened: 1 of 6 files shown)


joelittlejohn:
Great, thanks!

PR #547, opened 2016-04-11
Added missing methods and documentation for prefix/suffix stuff

samskiter:
I noticed some stuff was missing. Could probably have pushed this directly but it's quite a few changes


joelittlejohn:
Thanks :+1:

Issue #555, opened 2016-04-25
Constructors are incompatible with extends keyword


samskiter:
If schema A extends schema B and schema A has properties X,Y and Z and schema B has properties L,M and N and `includeConstructors` is set to true, the constructor of B correctly has arguments for L, M and N but the constructor of A only has arguments for X,Y and Z.

I would expect A to have LMNXYZ in its constructor.

Looking at the code, this looks non-trivial to fix.

joelittlejohn:
Closed by #558

Issue #554, opened 2016-04-25
Improve naming through extends keyword


samskiter:
**Problem**
The extends keyword just adds "Parent" to the end of the refrencing schema name. It seems that from any ref, a decent name could be generated deterministically (as it is in JsonSchema2Pojo.java).

The only gap I can see in the current naming rules is for an embedded referenced schema.

So this schema might cause issues:

```
{
  "$schema": "http://json-schema.org/draft-04/schema#",

  "definitions": {
    "address": {
      "type": "object",
      "properties": {
        "street_address": { "type": "string" },
        "city":           { "type": "string" },
        "state":          { "type": "string" }
      },
      "required": ["street_address", "city", "state"]
    }
  },

  "type": "object",

  "properties": {
    "billing_address": { "$ref": "#/definitions/address" },
    "shipping_address": { "$ref": "#/definitions/address" }
  }
}
```

And this is just a generalisation of the same thing with the extends keyword, e.g.:

```
test.json
{
  "$schema": "http://json-schema.org/draft-04/schema#",

  "definitions": {
    "address": {
      "type": "object",
      "properties": {
        "street_address": { "type": "string" },
        "city":           { "type": "string" },
        "state":          { "type": "string" }
      },
      "required": ["street_address", "city", "state"]
    }
  },

  "type": "object",
  "extends": { "$ref": "#/definitions/address" },
}
```

In the first case, the referenced address schema would be Billing_Address.java, which is a bit misleading, but not terrible,

In the second it would be called TestParent.java which is a pretty useless name.

**Proposal**

Come up with a deterministic way of naming schemas, not dependent on the route to that schema.

Currently JsonSchema2Pojo.java can come up with a good name from the path and target package for a given schema file. E.g. if the root directory is `/my/path/to/schema`, the target package is `a.b.c` and the schema is present at `/my/path/to/schema/d/test.json` - the generate class would be `a.b.c.d.Test`.

Now if test.json, the path to the address schema is `/my/path/to/schema/d/test.json#definitions/address` - could this create a class named: `a.b.c.d.Test.definitions.Address` ? I.e. use the jsonRef elements to name/position the class.

In the first example above, test.java looks like:

```
package a.b.c.d;

import a.b.c.d.definitions.Address;

public class Test {
    public Address billing_address;
    public Address shipping_address;
}
```

In the second, it looks like:

```
package a.b.c.d;

import a.b.c.d.definitions.Address;

public class Test extends Address {
}
```

[...]

PR #557, opened 2016-04-26
[Issue 555] Add test for constructors + extends

samskiter:
PR for fix for #555 ... WIP

### DIFF
--- a/jsonschema2pojo-integration-tests/src/test/java/org/jsonschema2pojo/integration/ExtendsIT.java
+++ b/jsonschema2pojo-integration-tests/src/test/java/org/jsonschema2pojo/integration/ExtendsIT.java
@@ -108,6 +108,36 @@ public class ExtendsIT {
 
     @Test
     @SuppressWarnings("rawtypes")
+    public void constructorHasParentsProperties() throws Exception {
+        ClassLoader resultsClassLoader = schemaRule.generateAndCompile("/schema/extends/subtypeOfB.json", "com.example", config("includeConstructors", true));
+
+        Class type = resultsClassLoader.loadClass("com.example.SubtypeOfB");
+        Class supertype = resultsClassLoader.loadClass("com.example.SubtypeOfBParent");
+
+        assertThat(type.getSuperclass(), is(equalTo(supertype)));
+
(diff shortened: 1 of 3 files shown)


samskiter:
I've put this in the main repository incase anyone else cares to have a go at fixing it.... it seems to be that the Schema object needs to be aware of the concept of a superclass. then we could iterate superclasses to compile the list of properties.

samskiter:
@joelittlejohn So I'm digging around and coming up with all sorts of ideas for this and #554 
Maybe we could have a chat tonight about both?

I'm currently wondering whether doing what I proposed in 554 would help this bug at all. Related to that, I'm wondering whether `Rule.apply()` could be changed to have a different signature:

`R apply(String nodeName, T generatableType, Schema currentSchema);`

As the `node` argument could probably be made available from `currentSchema.getContent()`.

It would be really useful to talk through the architecture you have now and what you think of my proposals...

Thanks,

Sam

ctrimble:
If you are going to be discussing the signature of Rule, I suggest that you modify it to take a single context object, possibly immutable.  This would allow the contract to be widened, as the project evolves, without too much impact on those implementing the interface.

samskiter:
@ctrimble Yea I was coming to the same conclusion - the context can just be the existing Schema object but with some more information stored in it. I think storing the resolved JType in it as is currently done is OK so it would be mutable

joelittlejohn:
I like the idea of a adding a context object to that signature.

Regarding naming, my main concern is that we enumerated as many edge-cases as possible, in particular those where refs and fragments paths are used.

One thing to be careful of here is biting off too much in one go. It would be good to make some small steps towards this that we can verify and merge (e.g. add the context object to make the signature more powerful) before doing anything radical around naming.

Thanks for taking an interest in these things @samskiter. I'm not likely to have time to devote to this as I've recently started a new job and have calls most evenings.

samskiter:
Ok, plan of action is context object, then naming. As for this constructor
bug, I can't quite see the best way to fit in the information about "parent
class" and "parent class properties". I suppose both could be worked out
every time they are needed. Anything else is just an optimisation...

samskiter:
Further question - what are the rules for adding schemas to the
schemastore? Any reason all Schema objects shouldn't be added?

PR #558, opened 2016-04-27
Fix for extends with constructor

samskiter:
So this isn't the prettiest, but it might do for now.

There's at least 2 new recursive methods and I had to make use of 'instanceof' :(

what do you think to storing classes JFieldVar's on the schema object? It would remove the need for the `instanceof` call...?

### DIFF
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/ObjectRule.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/ObjectRule.java
@@ -50,6 +50,8 @@ import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.HashMap;
+import java.util.LinkedHashSet;
 
 import javax.annotation.Generated;
 
@@ -177,13 +179,13 @@ public class ObjectRule implements Rule<JPackage, JType> {
      * @param node
      * @return
      */
-    private List<String> getConstructorProperties(JsonNode node, Schema schema, boolean onlyRequired) {
+    private LinkedHashSet<String> getConstructorProperties(JsonNode node, Schema schema, boolean onlyRequired) {
 
         if (!node.has("properties")) {
-            return new ArrayList<String>();
+            return new LinkedHashSet<String>();
         }
 
-        List<String> rtn = new ArrayList<String>();
+        LinkedHashSet<String> rtn = new LinkedHashSet<String>();
 
         NameHelper nameHelper = ruleFactory.getNameHelper();
         for (Iterator<Map.Entry<String, JsonNode>> properties = node.get("properties").fields(); properties.hasNext();) {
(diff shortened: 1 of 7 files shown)


samskiter:
Should fix https://github.com/joelittlejohn/jsonschema2pojo/pull/558

samskiter:
@ctrimble what do you think of this?

ctrimble:
LGTM.  In the testing, you should probably construct one of these children and make sure things are being passed to the parent properly.  Your current testing shows that the constructors are there, not that they work.

samskiter:
Yay! I managed to get rid of the use of 'instanceof'!

@ctrimble @joelittlejohn I also have an addition to this which adds copy constructors (performs a deep copy). Is this something you'd be interested in?

joelittlejohn:
@samskiter one last question I have on this one, will this change handle the situation where a property is defined in the parent and also defined in the child? (will it handle this without creating duplicate args in the constructor I mean). It should favour the child.

samskiter:
@joelittlejohn Just to clarify how this would work - If I construct then up-cast my child class to it's parent, the parent's fields should probably be initialised as per the constructor. I suppose the generated constructor could replicate the arg into the super constructor?

e.g.

```
private String name;
private String description;
public Child(String name, String description) {
    super(name);
    this.name = name;
    this.description = description;
}
```

If that's what you mean, no it wouldn't. A linked hashset rather than a list would probably do the job to sort duplicates internally so it may not be too hard to fix up...

I'll need to write another test case or two for this if it's what you want. We could add it as an improvement to be done after we get through we accept/reject the current set of PRs I have outstanding (if it is acceptable to use this in it's current state, of course). That would be preferable to me to avoid another round of rebasing :)

Let me know how you'd like to proceed...

joelittlejohn:
@samskiter It looks like this would be a good one to start with since it doesn't appear to depend on any other PRs. I'm keen to get the first of these PRs into a ready state before we move on to the others. We need to move forward with care here. I'm sorry our schedules haven't lined up well here but I really would like to try to focus on getting one PR right before merging others.

If using a Set rather than a List here would solve the problem then it would be good to make this change now. I'd like to do something sensible here rather than leave this problem in place (if we can) as it will otherwise simply create noise.

It would be helpful if you could remove the separate PRs that introduce test cases only (#557 is superseded by this PR, am I right?). It would allow me to understand all your PRs more easily I think.

Re the copy constructors, it certainly sounds like a good change. Lets get some of your current work merged first!

samskiter:
What I mean with rebasing is #561 is based off this and #560. theres a fair bit of git rearranging ive been doing my side to keep things looking nice your side. But if the change is trivial, it should be fine and I agree with you it's correct to get each piece right as we go..

I've removed the two dud PRs - they were there as I wasn't sure if I would have time to address and as an invitation for others :)

I'll look into this change now. #560 should be totally separate so could be reviewed now too i you have the time - I have some time to work on this this week.

samskiter:
@joelittlejohn I think I've added and fixed that case now. Can you have a look at the test please?

joelittlejohn:
Test looks good to me. One comment on the last commit: could you use `Set` (interface) on the left-hand-side of all declarations, and for method return types? Thanks.

samskiter:
Sure can do, but I tend not to in the case of using LinkedHashSet to replace List as it shows that it is still basically a list of reliable order. In fact I had another commit that changed the way arrays are generated when unique is set that exposes the fact there it uses a LinkedHashSet.

joelittlejohn:
Good point, happy to leave it as it is. I hadn't considered that. Shame Java doesn't have a an Ordered interface :)

samskiter:
Great, well I'll make sure to submit that one at some point too. Looks like we can move onto the more controversial change...

PR #559, opened 2016-04-28
add failing test for embedded refs to schema root

samskiter:
I will be working on a fix for this. I think that the library currently makes some incorrect assumptions about the way that refs should work and that's currently causing issue in other work im trying to do (on basic allOf support). Have a look at this really small test case and see if you agree.


samskiter:
fixed by #560

Issue #542, opened 2016-04-01
Internal references are not working properly


kayahr:
I noticed that jsonschema2pojo can't handle a large JSON schema I have and which can be successfully used by validators. I always got this error message:

    There's a problem: Path not present: definitions

I stripped down the schema bit by bit to find the part which is responsible for the problem and when I found it there was no error and even no complicated structure. It was only a simple (and correct) property reference. The reason for the failure was the reference level (It was referenced from a referenced type which was referenced from the root type).

Here is a small test schema to reproduce the problem:

    {
        "$schema": "http://json-schema.org/draft-04/schema#",
        "definitions": {
            "def1": {
                "type": "object",
                "properties": {
                    "prop1": {
                        "type": "string"
                    }
                },
                "additionalProperties": false
            },
            "def2": {
                "type": "object",
                "properties": {
                    "prop2": {
                        "$ref": "#/definitions/def1"
                    }
                },
                "additionalProperties": false
            },
            "def3": {
                "type": "object",
                "properties": {
                    "prop3": {
                        "$ref": "#/definitions/def2"
                    }
                },
                "additionalProperties": false
            }
        },
        "type": "object",
        "properties": {
            "prop": {
                "$ref": "#/definitions/def3"
            }
        },
        "additionalProperties": false
    }

Paste this into the online jsonschema2pojo and watch it fail. If you change the root reference from `def3` to `def2` then it works. Looks like jsonschema2pojo can follow references only two levels deep. It fails when a third level reference is found.

joelittlejohn:
Please retest with 0.4.23.

PR #560, opened 2016-04-29
Fix broken embedded refs

samskiter:
I've changed the way schema refs are resolved through schemastore to fix this.

### DIFF
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/SchemaStore.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/SchemaStore.java
@@ -19,6 +19,7 @@ package org.jsonschema2pojo;
 import static org.apache.commons.lang3.StringUtils.*;
 
 import java.net.URI;
+import java.net.URISyntaxException;
 import java.util.HashMap;
 import java.util.Map;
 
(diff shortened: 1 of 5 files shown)


samskiter:
There's a lot of string garbling to get this right...

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/SchemaStore.java
```diff
@@ -76,21 +77,42 @@ protected URI removeFragment(URI id) {
...
+        String stringId = id.toString();
+        if (stringId.endsWith("#")) {
+            try {
+                id = new URI(stripEnd(stringId, "#"));
+            } catch (URISyntaxException e) {
+                throw new IllegalArgumentException("Bad path: " + stringId);
+            }
+        }
+            
+        if (selfReferenceWithoutParentFile(parent, path) || substringBefore(stringId, "#").isEmpty()) {
```
samskiter: prevents us trying to use the contentresolver on a non-absolute incomplete URI

jsonschema2pojo-core/src/test/java/org/jsonschema2pojo/SchemaStoreTest.java
```diff
@@ -76,6 +76,20 @@ public void createWithSelfRef() throws URISyntaxException {
...
+    public void createWithEmbeddedSelfRef() throws URISyntaxException {
+
+        URI schemaUri = getClass().getResource("/schema/embeddedRef.json").toURI();
+
+        SchemaStore schemaStore = new SchemaStore();
+        Schema topSchema = schemaStore.create(schemaUri);
+        Schema embeddedSchema = schemaStore.create(topSchema, "#/definitions/embedded");
+        Schema selfRefSchema = schemaStore.create(embeddedSchema, "#");
+
+        assertThat(topSchema, is(sameInstance(selfRefSchema)));
```
samskiter: @joelittlejohn This is the core thing that I think I have fixed here. Do you agree with this test?

joelittlejohn:
This looks good to me. Does this change fix #542?

samskiter:
Hmmm, looks like it would. I wouldn't expect it to have even worked 2 levels deep in that example

Issue #368, opened 2015-06-03
Recursive schema $ref generates an extra class


rpatrick00:
I am trying to create a schema for a recursive etcd directory listing.  From what I can tell, it should look something like this:

``` json
{
    "id": "EtcdDirectoryListing",
    "$schema": "http://json-schema.org/draft-04/schema#",
    "description": "The etcd Manifest used for recursive directory listings",
    "type": "object",
    "properties": {
    "action": { "type": "string" },
    "node": {
        "id": "EtcdNode",
        "description": "Etcd key-value node",
        "type": "object",
        "properties": {
        "key": { "type": "string" },
        "dir": { "type": "boolean" },
        "nodes": {
            "type": "array",
            "items": {
            "$ref": "#/properties/node"
            }
        },
        "modifiedIndex": { "type": "integer" },
        "createdIndex": { "type": "integer" }
        }
    }
    }
}
```

The problem I am having is that the Maven plugin is generating both a Node.java and a Node_.java class.  I was expecting it to generate a single Node.java class that contains a List<Node> element in it.  I assume that I am doing something stupid so any help would be appreciated.

dadrus:
This is same as #481. Possible solution is proposed in its description.

PR #561, opened 2016-04-29
Schema contract changes

samskiter:
~~This is a very large PR as it includes changes from #560 and #558. The diff should therefore decrease if those changes are accepted~~

This is a change to the way the schemastore is used, essentially it's used more, for every potential object schema we come across.

There is also the beginnings of a context-like object in ObjectRule, call `SchemaClassInfo`. It may look a bit confusing having both the `classHierarchySchema` and `classDefinitionSchema` in as separate instance, but this allows a layer of indirection between the two which allows me to support `allOf` in a very simple way (only where it is exactly equivalent to extends) - see this https://github.com/samskiter/jsonschema2pojo/commit/94d9f1083a23df245b93de493717c9c053026a5d

I appreciated that you might not want to add allOf support and there is a lot of debate, so I tried to separate the two. So this is a bit of a speculative PR - it's really useful for me but maybe not so much for you! If it is ok, let me know if you would also like the allOf in there too.

Cheers,

Sam

### DIFF
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/SchemaStore.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/SchemaStore.java
@@ -32,6 +32,14 @@ public class SchemaStore {
     protected FragmentResolver fragmentResolver = new FragmentResolver();
     protected ContentResolver contentResolver = new ContentResolver();
 
+    public SchemaStore() {
+
+    }
+
+    public SchemaStore(ContentResolver contentResolver) {
+        this.contentResolver = contentResolver;
+    }
+
     /**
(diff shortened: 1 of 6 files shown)


samskiter:
@joelittlejohn here's the latest, rebased to make it simpler

samskiter:
@joelittlejohn did you have any thoughts on this?

samskiter:
@joelittlejohn Sorry to bug again. Where should I go with this in your opinion?

dclaze:
Bump

joelittlejohn:
5 years on, and I just resolved conflicts in this code again and used it all 😄 These commits are now merged into master and helped me solve #368 🎉

Issue #563, opened 2016-05-04
Lower camel case fields when using JSON


MungoRae:
Is it currently possible to have fields generated from JSON be lower camel case?

For example if I have JSON:
```
{
  "Property1" : "Any Value",
  "PropertyTwo" : "any value2",
  "property3" : "Value"
}
```

I want the class to look like
```
public class Clazz {
    private String property1;
    private String propertyTwo;
    private String property3;
}
```
At the moment it creates the fields as `Property1` which goes against common code style.

Thanks

joelittlejohn:
This is a regression, we should be generating lower camel case. Will definitely look to fix this. Thanks for raising it!

joelittlejohn:
Closed by #565

PR #565, opened 2016-05-10
Class fields are now created to be lower camel case. Added interation test.

MungoRae:
Fix for issue #563. Made fields lower camel case.

Thanks

### DIFF
--- a/jsonschema2pojo-integration-tests/src/test/resources/schema/ref/httpRefs.json
+++ b/jsonschema2pojo-integration-tests/src/test/resources/schema/ref/httpRefs.json
@@ -2,10 +2,10 @@
     "type" : "object",
     "properties" : {
         "address" : {
-            "$ref" : "https://jsonschema2pojo.googlecode.com/git-history/jsonschema2pojo-0.3.4/jsonschema2pojo-integration-tests/src/test/resources/schema/ref/address.json"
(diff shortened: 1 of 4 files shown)


samskiter:
Looks to me. i'm surprised you had to make a new method in nameutils... so I'm not sure that there's a test for schema either. I just had a look but couldn't find anything. Could you perhaps add a test for that input type too? I would probably expect it to be in `PropertyIT`

joelittlejohn:
Yes, agreed. @MungoRae could you move your integration test to PropertiesIT? We can merge this for the next release.

jsonschema2pojo-integration-tests/src/test/resources/schema/ref/httpRefs.json
```diff
@@ -2,10 +2,10 @@
     "type" : "object",
     "properties" : {
         "address" : {
-            "$ref" : "https://jsonschema2pojo.googlecode.com/git-history/jsonschema2pojo-0.3.4/jsonschema2pojo-integration-tests/src/test/resources/schema/ref/address.json"
```
MungoRae: This was failing one of the tests so I replaced it with a working URL to make sure the test passed. Not sure if this is the URL that should be used. Oddly enough this seemed to be passing yesterday.

MungoRae:
I updated the integration test to be inside of PropertiesIT and I changed it to use a schema instead to fit in with the rest of the tests in that set.

joelittlejohn:
Great, thanks! I guess some last piece of Google Code was finally killed off.

PR #569, opened 2016-05-14
Declare sets as linkedhashsets

samskiter:
This is the small change I mentioned to declare linkedhashsets publicly

### DIFF
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/ArrayRule.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/ArrayRule.java
@@ -17,7 +17,7 @@
 package org.jsonschema2pojo.rules;
 
 import java.util.List;
-import java.util.Set;
+import java.util.LinkedHashSet;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import org.jsonschema2pojo.Schema;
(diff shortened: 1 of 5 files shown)


joelittlejohn:
@samskiter I don't think it would be right to make this change. I was happy to leave this style in _this_ project in your change for #558 but I don't think we can roll this out to all users of this project.

IMO
* This isn't idiomatic in the Java world (despite there being an argument for doing it)
* These sets aren't required to be ordered
* There's no reason to restrict generated types to LinkedHashSet in particular (even if people want ordering, any implementation will do).

It's good to be conservative about changes that cause significant alteration to generated output, as this affects a lot of users. I don't think there's a strong case for doing it here and I think a lot of people won't like this for the reasons above.

Sorry if I wasn't clear when commenting on #558, I am happy that there is a logical reason to use LinkedHashSet there but I don't strong case to roll this out to the code we generate.

samskiter:
@joelittlejohn Fair enough, your call :)

IMO (just to have another bash at changing your mind ;) )

- If order is not required, why don't we declare as `Collection` regardless of whether unique is set or not. I would say that order is very much required when it comes to JSON arrays.
- Even by using Set you could argue you are restricting type too much. What if we wanted to switch to some 'UniqueList' implementation instead - we'd have to change the public declaration. The LinkedHashSet isn't ideal, but as you say, Java lacks much in the way of interfaces to defined the behaviour.

It is a long way to move from List to Set just because I've added 'unique'.

This change would never be a breaking change, but I agree it's important to be careful what it put out to users.

joelittlejohn:
Not convinced on either of your bullets :) but I'm not sure enumerating the reasons why would be useful.

Given the trade-offs here, declaring Set is closest to the expectations of the Java community IMO, which is why I think it's most appropriate for our generated sources.

PR #585, opened 2016-06-17
Swagger2 annotations

activey:
Leaving it here for now 'cause I have no time for writing up unit tests ;(

### DIFF
--- a/jsonschema2pojo-ant/pom.xml
+++ b/jsonschema2pojo-ant/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <artifactId>jsonschema2pojo</artifactId>
         <groupId>org.jsonschema2pojo</groupId>
-        <version>0.4.25-SNAPSHOT</version>
+        <version>0.4.24-SNAPSHOT</version>
     </parent>
 
     <artifactId>jsonschema2pojo-ant</artifactId>
--- a/jsonschema2pojo-ant/pom.xml
+++ b/jsonschema2pojo-ant/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <artifactId>jsonschema2pojo</artifactId>
         <groupId>org.jsonschema2pojo</groupId>
-        <version>0.4.24-SNAPSHOT</version>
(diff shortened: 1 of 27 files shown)


joelittlejohn:
Hi @activey. In this project we add annotations via 'annotators' e.g. [Jackson2Annotator](https://github.com/joelittlejohn/jsonschema2pojo/blob/5cdcaa20aff4e965e191c3e8f72f8b5c68d1e12a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/Jackson2Annotator.java). This is important as it allows annotations to be customized/extended/activated in a consistent way.

I think what you want here is a new annotation style: `JACKSON+SWAGGER`. When the AnnotatorFactory sees this annotation style is requested, it will create a [CompositeAnnotator](https://github.com/joelittlejohn/jsonschema2pojo/blob/5cdcaa20aff4e965e191c3e8f72f8b5c68d1e12a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/AnnotatorFactory.java#L34) that includes a Jackson2Annotator and a Swagger2Annotator, so something like:

```java
        case JACKSON_SWAGGER:
            return getAnnotator(new Jackson2Annotator(), new Swagger2Annotator());
```

You would then simply implement the Swagger2Annotator to add the annotations you need.

activey:
I believe u can use swagger annotations also with gson, they can be even used exclusively without any others.

activey:
This swagger thing here is quite similar what u have implemented for jsr303 annotations.

joelittlejohn:
@activey I wasn't aware that Swagger supported anything other than Jackson 2. You may be right that a separate property is fine for this.

Happy to resurrect this if someone wants to rebase and [add some integration tests](https://github.com/joelittlejohn/jsonschema2pojo/wiki/Contribute#3-add-integration-tests). Without tests I'm afraid this is not going anywhere.

santhosh-v:
This will be a great feature to support swagger since swagger is still alive.. any plan to get this forward ?

PR #599, opened 2016-07-26
support "required" keyword draft4 array-style when generating constru…

jjjhhhlll:
Hello,
I noticed that the "required" keyword only considers the draft3 style (boolean on a property) style when used with isConstructorsRequiredPropertiesOnly() == true.

This pull-request adds support for the draft4 style ("requires" as an array of property names in object) when generating constructors with only required properties.

Thanks again for this great library.

Enjoy,
Jon

### DIFF
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/ObjectRule.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/ObjectRule.java
@@ -24,10 +24,12 @@ import java.io.Serializable;
 import java.lang.reflect.Modifier;
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


jjjhhhlll:
just noticed this one too https://github.com/joelittlejohn/jsonschema2pojo/pull/379.  I think it is the same fix.

joelittlejohn:
Looks like your changes include a test, so we can merge this :)

Feel free to compare your changes to #379 and tweak this if there's anything there worth taking. Let me know when your ready and I will review and merge.

Thanks for contributing!

jjjhhhlll:
I like how that one loops over the JsonNode using ArrayNode interface.  It also checks that "required" items are actually textual.  I'll incorporate those things.

jjjhhhlll:
Ok. changes complete.

joelittlejohn:
Sorry, broken builds were something to do with the Sonatype oss-parent pom and an old dependency disappearing from the public apache Maven repo.

jjjhhhlll:
No worries.  Maven has advantages, but they come with problems like this sometimes.  Thanks for your help.

PR #609, opened 2016-08-14
Use unique enum constant names

### LINKED ISSUE #596: Adding symbols, JSON->JAVA?
When I am trying to convert json scheme to java object and I have an enum which are symbols like ":", "<", etc. The problem there is that when these enums are converted to java, they are done as : "__EMPTY__(":", "\u2265", "<");" and this token __EMPTY__ is generating an error which is correct since the names to these symbols cant be found. Any solutions for this problem? Thanks in advance.

thachhoang:
Fix #596.


joelittlejohn:
Great, thanks!

PR #611, opened 2016-08-20
Generate hashCode and equals even without declared fields

### LINKED ISSUE #610: equals and hashcode invokes super even if superclass has no fields
I can see in the previous pull requests that it was fixed to invoke super equals and hashcode due to older versions not doing the same. This is working all right except for the case (as ours) where top most class in the hierarchy has no fields hence no equals and hashcode implementation generated for the same and the immediate childs of this class ends up calling Objects equals and hashcode which in effect nullifies everything if not comparing same instances.

Due to this issue we are not able to use the latest version of the library. Can this be looked into as soon as possible?

Many thanks and let me know if there are any questions with regard to the same.

thachhoang:
In these generated methods, call the super version only if
the type does not extend java.lang.Object.

Fix #610.

### DIFF
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/ObjectRule.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/ObjectRule.java
@@ -364,7 +364,7 @@ public class ObjectRule implements Rule<JPackage, JType> {
     }
 
     private void addJsonTypeInfoAnnotation(JDefinedClass jclass, JsonNode node) {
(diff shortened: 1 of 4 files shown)


thachhoang:
@joelittlejohn Thanks for adding me to the repo.

joelittlejohn:
No problem. Thanks for submitting this PR. I'll take a proper look tomorrow but from a brief glance it looks great 👍

joelittlejohn:
@thachhoang only comment here: could you move the integration tests into IncludeHashCodeAndEqualsIT?

thachhoang:
I'll move the tests. Should I move the test json files elsewhere?

joelittlejohn:
No, location you have used for those is good. Thanks.

PR #612, opened 2016-08-24
Supporting other primitives as backing types for enums

jcogilvie:
This pull request should allow us to support enums that are non-String.  As a consequence, toString no longer makes sense as the default serialization method, so instead we create value() which does the same thing.  

toString is still overridden iff the type is String-backed (to preserve backwards compatibility), but is no longer annotated as @JsonValue.

ITs have been added and broken tests updated.


joelittlejohn:
This looks great. Two comments:

Could you remove the pom.xml changes from this PR?

Could you add an integration test that proves that serializing and deserializing some JSON with these enums actually works?

Cheers

jcogilvie:
Done and done.

PR #617, opened 2016-09-08
Allow setting custom annotator and custom rule factory fields directly in Gradle

bgorven:
This allows me to pass in an annotator implementation defined in the gradle file, which otherwise was throwing ClassNotFoundException.


joelittlejohn:
Looks like a useful addition for Gradle people. Thanks 👍

PR #627, opened 2016-09-20
[Issue #626] Enable GenerateJsonSchemaJavaTask to set custom configuration with closure; not just single jsonSchema2Pojo object.

### LINKED ISSUE #626: maven supports multiple executions; gradle doesn't support multiple tasks

scr:
Fixes #626

This change enables adding extra tasks that inherit from GenerateJsonSchemaJavaTask and extend a configuration with a closure rather than only being able to use the global and singleton `jsonSchema2Pojo`:

```
task generateJsonSchema2PojoTwo(type: GenerateJsonSchemaJavaTask) {
  configuration {
     source = files("${project.sourceSets.main.output.resourcesDir}/json2")
     targetPackage = 'example.json2'
     propertyWordDelimiters = ['_', ' ', '-'] as char[]
     sourceType = SourceType.JSON
   }
}
```

### DIFF
--- a/jsonschema2pojo-gradle-plugin/example/java/build.gradle
+++ b/jsonschema2pojo-gradle-plugin/example/java/build.gradle
@@ -1,3 +1,7 @@
+import org.jsonschema2pojo.SourceType
+import org.jsonschema2pojo.gradle.GenerateJsonSchemaJavaTask
+import org.jsonschema2pojo.gradle.JsonSchemaExtension
+
 apply plugin: 'java'
 apply plugin: 'jsonschema2pojo'
 
@@ -32,4 +36,13 @@ jsonSchema2Pojo {
   targetPackage = 'example'
   includeJsr303Annotations = true
   propertyWordDelimiters = ['_'] as char[]
+}
+
+task generateJsonSchema2PojoTwo(type: GenerateJsonSchemaJavaTask) {
+  configuration {
+    source = files("${project.sourceSets.main.output.resourcesDir}/json2")
+    targetPackage = 'example.json2'
+    propertyWordDelimiters = ['_', ' ', '-'] as char[]
+    sourceType = SourceType.JSON
+  }
 }
--- a/jsonschema2pojo-gradle-plugin/example/java/src/main/resources/json2/simple.json
+++ b/jsonschema2pojo-gradle-plugin/example/java/src/main/resources/json2/simple.json
@@ -1,0 +1,5 @@
+{
+  "@type": "Simple",
+  "num" : 5,
+  "name": "MyName"
+}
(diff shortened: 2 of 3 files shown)


scr:
@ben-manes - can you take a look and help land this?

ben-manes:
Maybe explain more about the intent here?

The usually the solution is to have multiple sourceSets and have tasks scoped. That lets you have multiple without these types of workarounds. Another approach is containers, which was part of my original Flyway plugin but stripped out after contributing. Its been a while since I dug into all of this, but I think your approach is the most obvious but not idiomatic.

scr:
@ben-manes Wow, yes I like where you're going with this - looking at https://docs.gradle.org/current/dsl/org.gradle.api.tasks.SourceSet.html I see that many other plugins follow this pattern of adding to and drawing from source sets for their configuration.

Several of the builds I'm working with have different configs for json2pojo - the easiest distinction to share is that some inputs are in 'jsonschema' format and others are 'json'.  While there are other settings different, this is the easiest thing to describe and understand the need for multiple configurations.

Some of the builds are currently maven, which supports this with multiple <execution> blocks, each with its own configuration.

The current implementation is not only not following the sourceSet idiom, but also doesn't seem to support multiple tasks on the same project (I tried using sub-projects but quickly became dismayed that the subprojects had to have all of the dependencies copied and all of the additions of generated-sources, etc needed to be created manually in the rootProject).

My inspiration in this particular change is to keep the existing use of the default configuration that is attached to `project.jsonSchema2Pojo`, and allow separate tasks to be created of `type: GenerateJsonSchemaJavaTask` much like you might create a `type: Exec` task.

I'd love to re-arrange the plugin to be more idiomatic, but I do think that doing it right might be a larger, breaking change requiring updates to the documentation, possibly changes to client configurations to use the sourceSet style. Also, the inability to have multiple tasks in the same project is really blocking or slowing some of my current work; if you'd consider landing this or some small modification to it for the moment, I'd gladly work towards a strawman PR of "the right/idiomatic way".

ben-manes:
I would probably lean towards containers, since Android does really crazy things to Gradle. So that might be a safer, cleaner option.

It sounds odd to me that this change is necessary. Most of the time this isn't needed because,
- Break a monolithic project into multiple subprojects (modules) so there is no task conflict
- Reuse a template script by applying it in each build and then customizing in that module
- Use logic (Groovy) to perform conditional logic based on build parameters
- Use `doFirst` to reconfigure just prior to execution

It sounds like you should give sub-projects another go. You should also try asking on the [Gradle Forum](https://discuss.gradle.org) for advise. I'm pretty sure you can do everything without any changes to this plugin. I'd prefer seeing what the Gradle team recommends and then we can follow their advise if changes to the plugin are desirable.

joelittlejohn:
I'm going to tentatively close this one as it sounds like this isn't the idiomatic way to solve this problem with Gradle.

PR #643, opened 2016-11-07
Support for @JsonFormat annotation

shravanspurohit:
Adds @JsonFormat annotation to specified property. An example of the jsonschema is:

```
{
    "type" : "object",
    "properties" : {
        "stringAsJavaDate" : {
            "type" : "string",
            "format" : "date-time",
            "customDateTimePattern" : "yyyy-MM-dd'T'HH:mm:ss.SSSZ",
            "timezone" : "PST"
        },
        "stringAsJodaDateTime" : {
            "javaType" : "org.joda.time.DateTime",
            "format" : "date-time",
            "customDateTimePattern" : "yyyy-MM-dd'T'HH:mm:ss.SSSZ",
            "timezone" : "PST"
        }
    }
}
```

This PR is not ready to be merged. I created this to get feedback. I am still trying to get the integration tests to work.

### DIFF
--- a/jsonschema2pojo-ant/src/main/java/org/jsonschema2pojo/ant/Jsonschema2PojoTask.java
+++ b/jsonschema2pojo-ant/src/main/java/org/jsonschema2pojo/ant/Jsonschema2PojoTask.java
@@ -140,6 +140,8 @@ public class Jsonschema2PojoTask extends Task implements GenerationConfig {
     private String timeType = null;
 
     private String dateType = null;
+    
+    private boolean formatDateTime = false;
 
 
     /**
@@ -668,6 +670,18 @@ public class Jsonschema2PojoTask extends Task implements GenerationConfig {
     public void setIncludeDynamicAccessors(boolean includeDynamicAccessors) {
         this.includeDynamicAccessors = includeDynamicAccessors;
     }
+    
+    /**
+     * Sets the 'formatDateTime' property of this class
+     *
+     * @param formatDateTime
+     *            Whether the fields of type `date-type` have the `@JsonFormat` annotation 
+     *            with pattern set to the default value of `yyyy-MM-dd'T'HH:mm:ss.SSS` 
+     *            and timezone set to default value of `UTC`
+     */
+    public void setFormatDateTime(boolean formatDateTime) {
+        this.formatDateTime = formatDateTime;
+    }
 
     @Override
     public boolean isGenerateBuilders() {
@@ -899,6 +913,11 @@ public class Jsonschema2PojoTask extends Task implements GenerationConfig {
     @Override
     public boolean isUseBigDecimals() {
         return useBigDecimals;
+    }
+
+    @Override
+    public boolean isFormatDateTime() {
+        return formatDateTime;
     }
 
 }
--- a/jsonschema2pojo-ant/src/main/java/org/jsonschema2pojo/ant/Jsonschema2PojoTask.java
+++ b/jsonschema2pojo-ant/src/main/java/org/jsonschema2pojo/ant/Jsonschema2PojoTask.java
@@ -141,7 +141,7 @@ public class Jsonschema2PojoTask extends Task implements GenerationConfig {
 
     private String dateType = null;
     
-    private boolean formatDateTime = false;
+    private boolean formatDateTimes = false;
 
 
     /**
@@ -672,15 +672,15 @@ public class Jsonschema2PojoTask extends Task implements GenerationConfig {
     }
     
     /**
-     * Sets the 'formatDateTime' property of this class
+     * Sets the 'formatDateTimes' property of this class
      *
-     * @param formatDateTime
-     *            Whether the fields of type `date-type` have the `@JsonFormat` annotation 
-     *            with pattern set to the default value of `yyyy-MM-dd'T'HH:mm:ss.SSS` 
+     * @param formatDateTimes
+     *            Whether the fields of type <code>date-type</code> have the <code>@JsonFormat</code> annotation 
+     *            with pattern set to the default value of <code>yyyy-MM-dd'T'HH:mm:ss.SSS</code> 
      *            and timezone set to default value of `UTC`
      */
-    public void setFormatDateTime(boolean formatDateTime) {
-        this.formatDateTime = formatDateTime;
+    public void setFormatDateTime(boolean formatDateTimes) {
+        this.formatDateTimes = formatDateTimes;
     }
 
     @Override
@@ -916,8 +916,8 @@ public class Jsonschema2PojoTask extends Task implements GenerationConfig {
     }
 
     @Override
-    public boolean isFormatDateTime() {
-        return formatDateTime;
+    public boolean isFormatDateTimes() {
+        return formatDateTimes;
     }
 
 }
--- a/jsonschema2pojo-ant/src/site/Jsonschema2PojoTask.html
+++ b/jsonschema2pojo-ant/src/site/Jsonschema2PojoTask.html
@@ -302,8 +302,8 @@
         <td align="center" valign="top">No</td>
       </tr>
       <tr>
-        <td valign="top">formatDateTime</td>
-        <td valign="top">Whether the fields of type <code>date-type</code> have the <code>@JsonFormat</code> annotation with pattern set to the default value of <code>yyyy-MM-dd'T'HH:mm:ss.SSS</code> and timezone set to default value of <code>UTC</code>
+        <td valign="top">formatDateTimes</td>
+        <td valign="top">Whether the fields of type <code>date-time</code> have the <code>@JsonFormat</code> annotation with pattern set to the default value of <code>yyyy-MM-dd'T'HH:mm:ss.SSS</code> and timezone set to default value of <code>UTC</code>
         </td>
         <td align="center" valign="top">No (default <code>false</code>)</td>
       </tr>
--- a/jsonschema2pojo-ant/src/site/Jsonschema2PojoTask.html
+++ b/jsonschema2pojo-ant/src/site/Jsonschema2PojoTask.html
@@ -301,6 +301,12 @@
         </td>
         <td align="center" valign="top">No</td>
       </tr>
+      <tr>
+        <td valign="top">formatDateTime</td>
+        <td valign="top">Whether the fields of type <code>date-type</code> have the <code>@JsonFormat</code> annotation with pattern set to the default value of <code>yyyy-MM-dd'T'HH:mm:ss.SSS</code> and timezone set to default value of <code>UTC</code>
+        </td>
+        <td align="center" valign="top">No (default <code>false</code>)</td>
+      </tr>
     </table>
 
     <h3>Examples</h3>
--- a/jsonschema2pojo-cli/src/main/java/org/jsonschema2pojo/cli/Arguments.java
+++ b/jsonschema2pojo-cli/src/main/java/org/jsonschema2pojo/cli/Arguments.java
@@ -162,8 +162,8 @@ public class Arguments implements GenerationConfig {
     @Parameter(names = { "-ida", "--include-dynamic-accessors" }, description = "Include dynamic getter, setter, and builder support on generated types.")
     private boolean includeDynamicAccessors = false;
     
-    @Parameter(names = { "-fdt", "--format-date-time" }, description = "Whether the fields of type `date-type` have the `@JsonFormat` annotation with pattern set to the default value of `yyyy-MM-dd'T'HH:mm:ss.SSS` and timezone set to default value of `UTC`")
-    private boolean formatDateTime = false;
+    @Parameter(names = { "-fdt", "--format-date-times" }, description = "Whether the fields of type `date-time` have the `@JsonFormat` annotation with pattern set to the default value of `yyyy-MM-dd'T'HH:mm:ss.SSS` and timezone set to default value of `UTC`")
+    private boolean formatDateTimes = false;
     
     private static final int EXIT_OKAY = 0;
     private static final int EXIT_ERROR = 1;
@@ -405,8 +405,8 @@ public class Arguments implements GenerationConfig {
     }
 
     @Override
-    public boolean isFormatDateTime() {
-        return formatDateTime;
+    public boolean isFormatDateTimes() {
+        return formatDateTimes;
     }
 
 }
--- a/jsonschema2pojo-cli/src/main/java/org/jsonschema2pojo/cli/Arguments.java
+++ b/jsonschema2pojo-cli/src/main/java/org/jsonschema2pojo/cli/Arguments.java
@@ -161,7 +161,10 @@ public class Arguments implements GenerationConfig {
 
     @Parameter(names = { "-ida", "--include-dynamic-accessors" }, description = "Include dynamic getter, setter, and builder support on generated types.")
     private boolean includeDynamicAccessors = false;
-
+    
+    @Parameter(names = { "-fdt", "--format-date-time" }, description = "Whether the fields of type `date-type` have the `@JsonFormat` annotation with pattern set to the default value of `yyyy-MM-dd'T'HH:mm:ss.SSS` and timezone set to default value of `UTC`")
+    private boolean formatDateTime = false;
+    
     private static final int EXIT_OKAY = 0;
     private static final int EXIT_ERROR = 1;
 
@@ -399,6 +402,11 @@ public class Arguments implements GenerationConfig {
     @Override
     public boolean isUseBigDecimals() {
         return useBigDecimals;
+    }
+
+    @Override
+    public boolean isFormatDateTime() {
+        return formatDateTime;
     }
 
 }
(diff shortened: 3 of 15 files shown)


OverlyExcessive:
I think it looks good. 

`javaType`seems like a redundant property as there is a setting for formatting dates as JodaTime `DateTime` already.

Also as @joelittlejohn mentioned earlier, if we want to conform better to the latest JSON schema spec, then any property with `format: "date-time"` should be a RFC3339 date-time formatted string. So the addition of a `customDateTimePattern` will mean that you will be able to define schemas that violate the date formatting invariant.

shravanspurohit:
@OverlyExcessive are you suggesting that I need to validate the pattern we receive in `customDateTimePattern` to ensure it corresponds to RFC3339 format?

OverlyExcessive:
@shravanspurohit I'm suggesting that we get rid of the `customDateTimePattern` property as it is not a part of the official JSON Schema spec, if the `date-time` format is used then according to the latest JSON schema spec the provided value must be a RFC3339 formatted date-time.

shravanspurohit:
@OverlyExcessive I am a bit concerned with this. Here is why.

Firstly, this will break a lot of things for a lot of people. Right now, I can extend the ObjectMapper class and set the date format to anything I want.
```
public class CustomObjectMapper extends ObjectMapper {
	public CustomObjectMapper(){
		setDateFormat(new SimpleDateFormat("yyyy-MM-dd")); // Set any format here
		configure(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS, false);
	}
}
```
And I can register this class as a json provider on my server. The date format you set here would be the global configuration which applies to all date-time fields.
With the change you are proposing, when we use date-time in the schema, if we by default include the `@JsonFormat` annotation on the field and use the RFC3339 date-time format, this configuration will take precedence and the the format set in the CustomObjectMapper would not be honored and there by breaking existing things.

Secondly, people are using @JsonFormat annotation to use custom format. I personally turned to this because I had to change all my (Date) fields to include date and time except the dob field where I wanted just the date. Since I was working with legacy code, changing the date format in the CustomObjectMapper to `yyyy-MM-dd'T'HH:mm:ss.SSSZ` and using `@JsonFormat(pattern='yyyy-MM-dd')` for my dob field was basically a 2 line code change. 
I do agree that we want to follow json schema 4 specs but if you look at Jackson's documentation http://wiki.fasterxml.com/JacksonFAQDateHandling you will see that they introduced @JsonFormat specifically to allow us to use whatever format we want. 

What I would like to do is figure out a way in which jsonschema2pojo can use the full extent of capabilities Jackson provides.

shravanspurohit:
What are your thoughts on this, @joelittlejohn ?

joelittlejohn:
There isn't an easy answer here, but here's what I'm thinking right now:

I'd like this tool to (by default) do the right thing and create valid documents according to the rules laid out by the JSON Schema validation spec. Just as an aside, I'd also like this tool to support and encourage using the best and ubiquitous date format with no effort. We have a legacy here though and a backwards-compatibility problem - at the very least we should give people a way to keep the old behaviour.

I think the best route forward would be:

* By default, add the JsonFormat annotation when `"format":"date-time"` appears. This should use `yyyy-MM-dd'T'HH:mm:ss.SSSZ`.
* It should be possible to disable this by setting `<formatDatetimes>false</formatDatetimes>` in the plugin configuration.
* `customDateFormat` schema rule should be supported, and can be used if you really want to override the format and create data that isn't strictly valid according to the spec.
* `customTimeZone` schema rule should be supported, and can be used if want to override the timezone in which the instant is rendered.

shravanspurohit:
I like this solution. So this is what I understood. Correct me if I am wrong.

If `formatDateTimes` is set to `true` we will add the `@JsonFormat` property with pattern as `yyyy-MM-dd'T'HH:mm:ss.SSSZ` and use the `customDateFormat` if provided.

If `formatDateTimes` is set to `false` we will **NOT** add the `@JsonFormat` property with pattern as `yyyy-MM-dd'T'HH:mm:ss.SSSZ` by default. However, if the `customDateFormat` is provided, we will add the `@JsonFormat` to that field.

Basically, the setting `formatDateTimes` controls whether we add `@JsonFormat` with pattern `yyyy-MM-dd'T'HH:mm:ss.SSSZ` as default or not.

Regarding the implementation, I have 2 options.
1. In `Jackson2Annotator` class, I include a variable named `GenerationConfig` and initialize this when the annotator is instantiated.
2. In `Jackson2Annotator` class, for method `jsonFormat`, I pass in the `GenerationConfig` as a parameter since this isnt being used anywhere else. The caveat being we might have other annotations in future which might require the `GenerationConfig` too. But agile says dont plan for a future unknown :-)

OverlyExcessive:
@shravanspurohit For me I would say either way is fine. Would be nice if we could move this PR along, let me know if you need any assistance.

joelittlejohn:
@shravanspurohit I've made a change recently to ensure that Annotators receive the generation config, and that any custom annotator that has a constructor that takes a generation config will also received it.

shravanspurohit:
@joelittlejohn @OverlyExcessive Its done!

jsonschema2pojo-ant/src/main/java/org/jsonschema2pojo/ant/Jsonschema2PojoTask.java
```diff
@@ -668,6 +670,18 @@ public void setTargetVersion(String targetVersion) {
     public void setIncludeDynamicAccessors(boolean includeDynamicAccessors) {
         this.includeDynamicAccessors = includeDynamicAccessors;
     }
+    
+    /**
+     * Sets the 'formatDateTime' property of this class
+     *
+     * @param formatDateTime
+     *            Whether the fields of type `date-type` have the `@JsonFormat` annotation
```
joelittlejohn: I think `date-type` should be `date-time`. Also, I think the backticks should be `<code></code>` (?).

jsonschema2pojo-ant/src/main/java/org/jsonschema2pojo/ant/Jsonschema2PojoTask.java
```diff
@@ -140,6 +140,8 @@
     private String timeType = null;
 
     private String dateType = null;
+    
+    private boolean formatDateTime = false;
```
joelittlejohn: Could you rename this config option to `formatDateTimes`? (plural) This fits better with the conventions in this project and also describes this feature better I think.

jsonschema2pojo-ant/src/site/Jsonschema2PojoTask.html
```diff
@@ -301,6 +301,12 @@
         </td>
         <td align="center" valign="top">No</td>
       </tr>
+      <tr>
+        <td valign="top">formatDateTime</td>
+        <td valign="top">Whether the fields of type <code>date-type</code> have the <code>@JsonFormat</code> annotation with pattern set to the default value of <code>yyyy-MM-dd'T'HH:mm:ss.SSS</code> and timezone set to default value of <code>UTC</code>
```
joelittlejohn: `date-type` should be `date-time` I think

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/Annotator.java
```diff
@@ -160,6 +160,20 @@
...
+     * @param field
+     *            the field that contains data that will be serialized
+     * @param clazz
+     *            the owner of the field (class to which the field belongs)
+     * @param propertyName
+     *            the name of the JSON property that this field represents
+     * @param propertyNode
+     *            the schema node defining this property
+     */
+    void jsonFormat(JFieldVar field, JDefinedClass clazz, String propertyName, JsonNode node);
```
joelittlejohn: The implementation of this method doesn't seem to use `clazz` at all, could you remove this from the signature?

joelittlejohn: This method needs a new name I think, `jsonFormat` doesn't describe what this does, it's too general. I think a good name for this method (that fits with the other methods in this interface) would be: `dateField`.

jsonschema2pojo-cli/src/main/java/org/jsonschema2pojo/cli/Arguments.java
```diff
@@ -161,7 +161,10 @@
 
     @Parameter(names = { "-ida", "--include-dynamic-accessors" }, description = "Include dynamic getter, setter, and builder support on generated types.")
     private boolean includeDynamicAccessors = false;
-
+    
+    @Parameter(names = { "-fdt", "--format-date-time" }, description = "Whether the fields of type `date-type` have the `@JsonFormat` annotation with pattern set to the default value of `yyyy-MM-dd'T'HH:mm:ss.SSS` and timezone set to default value of `UTC`")
```
joelittlejohn: s/date-type/date-time/

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/GenerationConfig.java
```diff
@@ -412,5 +412,13 @@
      *         date-time) to generated Java types.
      */
     String getTimeType();
+    
+    /**
+     * Gets the `formatDateTime` configuration option 
+     *
+     * @return Whether the fields of type `date-type` have the `@JsonFormat` annotation
```
joelittlejohn: s/date-type/date-time/

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/Jackson2Annotator.java
```diff
@@ -45,9 +46,11 @@
  *      href="https://github.com/FasterXML/jackson-annotations">https://github.com/FasterXML/jackson-annotations</a>
  */
 public class Jackson2Annotator extends AbstractAnnotator {
-
+    private GenerationConfig generationConfig;
```
joelittlejohn: This class doesn't need this field, it should call `getGenerationConfig` from the parent.

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/Jackson2Annotator.java
```diff
@@ -45,9 +46,11 @@
...
+    
     public Jackson2Annotator(GenerationConfig generationConfig) {
         super(generationConfig);
+        this.generationConfig = generationConfig;
```
joelittlejohn: This can be removed.

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/Annotator.java
joelittlejohn: The implementation of this method doesn't seem to use `propertyName` at all, could you remove this from the signature?

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/AbstractAnnotator.java
```diff
@@ -95,4 +95,8 @@ public GenerationConfig getGenerationConfig() {
         return generationConfig;
     }
 
+	@Override
```
joelittlejohn: Many of these changes are using tabs instead of spaces. Could you remove all tabs from this PR and use spaces instead?

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/Jackson2Annotator.java
```diff
@@ -124,4 +127,29 @@ public boolean isAdditionalPropertiesSupported() {
     public void additionalPropertiesField(JFieldVar field, JDefinedClass clazz, String propertyName) {
         field.annotate(JsonIgnore.class);
     }
+
+    @Override
+    public void jsonFormat(JFieldVar field, JDefinedClass clazz, String propertyName, JsonNode node) {
+        boolean formatDateTime = generationConfig.isFormatDateTime();
+        String iso8601DateTimeFormat = "yyyy-MM-dd'T'HH:mm:ss.SSS";
+        String customDateTimePattern = node.has("customDateTimePattern") == true ? node.get("customDateTimePattern").asText() : null;
```
joelittlejohn: No need for `== true`

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/Jackson2Annotator.java
```diff
@@ -124,4 +127,29 @@ public boolean isAdditionalPropertiesSupported() {
...
+        String timezone = node.has("customTimezone") == true ? node.get("customTimezone").asText() : "UTC";
```
joelittlejohn: No need for `== true`

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/PropertyRule.java
```diff
@@ -134,6 +137,13 @@ private void propertyAnnotations(String nodeName, JsonNode node, Schema schema,
             ruleFactory.getNotRequiredRule().apply(nodeName, node.get("required"), generatedJavaConstruct, schema);
         }
     }
+    
+    private void formatAnnotation(JFieldVar field, JDefinedClass clazz, String propertyName, JsonNode node) {
+    	String format = node.has("format") == true ? node.get("format").asText() : null;
```
joelittlejohn: I'd simplify this to

```java
String format = node.path("format").asText();
```

jsonschema2pojo-integration-tests/src/test/java/org/jsonschema2pojo/integration/CustomDateTimeFormatIT.java
```diff
@@ -0,0 +1,182 @@
...
+        
+        if (isAnnotated) {
+            JsonFormat annotation = field.getAnnotation(JsonFormat.class);
+            
+            assertThat(annotation, notNullValue());
+            // Assert that the patterns match
+            assertEquals(format, annotation.pattern());
+            // Assert that the timezones match
+            assertEquals(timezone, annotation.timezone());
+        }
```
joelittlejohn: This project uses the following style for if:

```java
if (...) {

} else {

}
```

Could you follow this? Thanks

jsonschema2pojo-integration-tests/src/test/java/org/jsonschema2pojo/integration/CustomDateTimeFormatIT.java
```diff
@@ -0,0 +1,182 @@
...
+import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameters;
+
+import com.fasterxml.jackson.annotation.JsonFormat;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+
+@RunWith(Parameterized.class)
+public class CustomDateTimeFormatIT {
```
joelittlejohn: I appreciate that you're trying to minimize duplication here, but this test is _really_ hard to follow and maintain. The test itself has configurable behaviour and assertions, and it's simply very hard to see that examples are being tested.

Could you refactor this so that it not a parameterized test? These tests are good when theres is a large body of test data, but they're not good for readability and maintainability when they're being used to create a single test with many different switchable code paths.

Each example should have a clear, descriptive method (test) name. If there's duplication then you can add some helper methods to the test.

jsonschema2pojo-maven-plugin/src/main/java/org/jsonschema2pojo/maven/Jsonschema2PojoMojo.java
```diff
@@ -542,6 +542,17 @@
      * @readonly
      */
     private MavenProject project;
+    
+    /**
+     * Whether the fields of type `date-type` have the `@JsonFormat` annotation
```
joelittlejohn: s/date-type/date-time/

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/AbstractAnnotator.java
shravanspurohit: Hmm... I used in tabs in files where there were tabs and spaces if there were spaces. I'll verify this and fix it if needed.

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/Jackson2Annotator.java
shravanspurohit: Yeah, I just thought I'll make the logic obvious. Removed.

jsonschema2pojo-integration-tests/src/test/java/org/jsonschema2pojo/integration/CustomDateTimeFormatIT.java
shravanspurohit: Yeah, I felt the same but I decided to optimize it for performance. I'll make it more readable instead.

jsonschema2pojo-integration-tests/src/test/java/org/jsonschema2pojo/integration/CustomDateTimeFormatIT.java
shravanspurohit: Sure

shravanspurohit:
@joelittlejohn Done.

PR #645, opened 2016-11-12
Adding support for JSR-305 annotations

shrey-garg:
Adding support for JSR-305 compatible annotations, for defect detection using libraries like google's find bugs or while using Intellij Idea IDE etc.

It adds @NonNull annotations for all fields that are marked required and @Nullable on all fields that are not.

### DIFF
--- a/jsonschema2pojo-ant/src/main/java/org/jsonschema2pojo/ant/Jsonschema2PojoTask.java
+++ b/jsonschema2pojo-ant/src/main/java/org/jsonschema2pojo/ant/Jsonschema2PojoTask.java
@@ -95,6 +95,8 @@ public class Jsonschema2PojoTask extends Task implements GenerationConfig {
 
     private boolean includeJsr303Annotations = false;
 
+    private boolean includeJsr305Annotations = false;
+
     private SourceType sourceType = SourceType.JSONSCHEMA;
 
     private Path classpath;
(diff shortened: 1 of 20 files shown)


joelittlejohn:
Thanks @shrey-garg. I note that JSR-305 is dormant, which I guess means it is not going anywhere and is unlikely to ever be added to Java.

Thanks for putting this behind a configuration option. I've never come across this JSR before but I don't have any objections to adding this. I'll take a detailed look through the code and comment on anything that needs to change.

shrey-garg:
Thank you.

Will incorporate any suggestions that you might have.

jsonschema2pojo-ant/src/main/java/org/jsonschema2pojo/ant/Jsonschema2PojoTask.java
```diff
@@ -458,6 +460,19 @@ public void setIncludeJsr303Annotations(boolean includeJsr303Annotations) {
     }
 
     /**
+     * Sets the 'includeJsr303Annotations' property of this class
```
joelittlejohn: Typo here, this should be 305

shrey-garg: corrected

shrey-garg:
I have made changes to take RequiredArrayRule into consideration and have also added integration tests which validate if the proper annotations have been added or not.

joelittlejohn:
@shrey-garg This looks good. The only thing I'm going to do here is remove NotRequiredRule.java, because I don't think that it's right to add these comments to all existing code. It's a lot of noise for all projects.

shrey-garg:
@joelittlejohn The NotRequired comment would only be added if the JSR-305 flag is on, it won't be added in any other scenario.

joelittlejohn:
Ah, so it is. I misread this. Thanks!

PR #647, opened 2016-11-16
Add plugin infos

StefMa:
I've just created a "plain old" gradle project. 
Only for creating the json files to objects. Don't want to import it in my android project. 
So it ended up with:
```java
* What went wrong:
A problem occurred evaluating root project 'pojoGen'.
> Failed to apply plugin [id 'jsonschema2pojo']
   > generateJsonSchema: Java or Android plugin required
```

For further readers it would be great to know that info ;)


joelittlejohn:
👍

PR #665, opened 2016-12-17
Support URI format properties with a default value using URI.create

### LINKED ISSUE #664: Support URI format properties with a default value using URI.create
```json
{
  "type":"object",
  "properties": {
    "type": {
      "type": "string",
      "format": "uri",
      "default": "abc"
    },
    "types": {
      "items": {
        "type": "string",
        "format": "uri"
      },
      "uniqueItems": true,
      "type": "array",
      "default": [ "/abc/efg" ]
    }
  }
}
```

This generates the fields,
```java
@JsonProperty("type")
private URI type = null;
@JsonProperty("types")
@JsonDeserialize(as = java.util.LinkedHashSet.class)
private Set<URI> types = new LinkedHashSet<URI>(Arrays.asList(null));
```

But when the `format` is removed it generates a a String value,
```java
@JsonProperty("type")
private String type = "abc";
@JsonProperty("types")
@JsonDeserialize(as = java.util.LinkedHashSet.class)
private Set<String> types = new LinkedHashSet<String>(Arrays.asList("/abc/efg"));
```

Can we support creating default URIs? Presumably they could be private static fields.

thachhoang:
Fix #664.

### DIFF
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/DefaultRule.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/DefaultRule.java
@@ -30,6 +30,7 @@ import com.sun.codemodel.JType;
 
 import java.math.BigDecimal;
 import java.math.BigInteger;
+import java.net.URI;
 import java.text.ParseException;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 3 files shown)


joelittlejohn:
Brilliant, thanks @thachhoang

PR #671, opened 2016-12-27
Jackson inclusion option

vladimir-alekseenko:
Hi, I have implemented option to set the "inclusion level" for jackson1 and jackson2 serializers, works in ant, cli, gradle and maven. Has integration tests.


joelittlejohn:
@vladimir-alekseenko thanks a lot for working on this. This will be included in 0.4.30

PR #672, opened 2017-01-06
Allow pattern, required, and minimum/maximum to declare message arguments

DavidTanner:
This allows for any properties of an annotation to be assigned inside the schema document.

I would like to be able to assign custom message's for errors that are returned when a field doesn't match a regexp, or if a required field isn't present.  I left the min/max items/length alone and added a new size rule that can have the message set as well as the min/max value.


joelittlejohn:
Hi @DavidTanner. Thanks for sharing this code. It might be useful for others.

Unfortunately I don't see this ever being merged into the core of jsonschema2pojo. As far as I can tell you've implemented your own extension to JSON schema that uses a new syntax for validation and error messages. This is great but I'm afraid it doesn't belong in the core of this project as you've diverged from how JSON schema does things.

DavidTanner:
@joelittlejohn You are right that it is different from the JSON schema definition.  I was hoping since it keeps the regular functionality, but just adds the Java fields it would be acceptable.  Do you have a recommendation for how I might be able to accomplish the same thing while keeping the proper Schema?

Issue #673, opened 2017-01-06
Cannot use custom date,time types from external dependencies in maven/ant plugins


bbohl:
Trying to use the `<dateTimeType>` , `<dateType>`, and `<timeType>` feature of the maven plugin where the type is sourced from a dependency of my project.  It always receives a `ClassNotFoundException`.  The plugin tries to extend the classloader with the project dependencies and sets the current thread's contextloader, but the `org.jsonschema2pojo.rules.FormatRule` class doesn't use the same classloader.

PR #674, opened 2017-01-06
Fix customtypes classloader, issue #673

bbohl:


ctrimble:
@bbohl I am curious about the situation motivating this change, for my own benefit.  There may be some other problems related to this same situation.

bbohl:
@ctrimble, see issue #673.  Just trying to use custom classes for the date-time, date, and time formats with the maven plugin.  The classes come from a dependency in my project.  

The plugins inject a classloader that contain the dependencies to the current thread's contextcontextloader.  However `Class.forName` ends up using the original classloader, which don't contain the dependencies.

ctrimble:
Thanks for the details!

PR #679, opened 2017-01-19
Added option to not overwrite files if there are no changes

19870213:
Hi, we (at my workplace) noticed that FileCodeWriter will always overwrite files, even if there are no changes), this in turns triggers further recompilations across our multi-module project. (Even after accounting for https://issues.apache.org/jira/browse/MCOMPILER-209)

This change 'buffers' the output of FileCodeWriter and compares it to the current output file (if it's there) and if and only if it's different will it write out the file.

I'm not yet sure about the naming of the new property (overwriteEvenIfUnchanged, default true so that the default behaviour is unchanged) or the new classname (I'm now thinking 'BufferedCodeWriter is a more apt name instead of CachingCodeWriter). Feedback very welcome here.

I have added an integration test (OverwriteIfUnchangedIT) for this change and all tests are passing on my (Windows) machine.

### DIFF
--- a/jsonschema2pojo-ant/src/main/java/org/jsonschema2pojo/ant/Jsonschema2PojoTask.java
+++ b/jsonschema2pojo-ant/src/main/java/org/jsonschema2pojo/ant/Jsonschema2PojoTask.java
@@ -16,7 +16,8 @@
 
 package org.jsonschema2pojo.ant;
 
-import static org.apache.commons.lang3.StringUtils.*;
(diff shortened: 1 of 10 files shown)


joelittlejohn:
Thanks for pushing this so that others can use this code. I'm not sure this belongs in the core to be honest. I've never heard of any other user experiencing this as a problem. 

I don't like to add new configuration options that only one user will use. Lets see if others are interested in this before including it in the core.

Issue #688, opened 2017-02-14
Custom path separator characters


Tovkal:
Hi,

I've got some jsonschemas from [Sabre](developer.sabre.com) and need to generate the Java classes for it, using the maven plugin. Here's the [request schema](http://files.developer.sabre.com/doc/providerdoc/STPS/bfm/v300/BargainFinderMax_v3.0.0_Request.jsonschema) for one of their APIs, [BargainFinderMax](https://developer.sabre.com/docs/read/rest_apis/air/search/bargain_finder_max). The problem I have with these schemas is that the definitions' names are all fully-qualified Java names.

A simplified version of the schema that reproduces the problem would be this:

```
{
  "$schema" : "http://json-schema.org/draft-04/schema#",
  "title" : "com.tovkal.FooRQ",
  "type" : "object",
  "properties" : {
    "FooRQ" : {
      "type" : "object",
      "properties" : {
        "POS" : {
          "$ref" : "#/definitions/com.tovkal.Bar"
        }
      }
    }
  },
  "definitions" : {
    "com.tovkal.Bar" : {
      "type" : "object",
      "properties" : {
        "AgentSine" : {
          "type" : "string"
        }
      }    }
  }
}
```

Running this schema using the maven plugin throws this error:

```
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.jsonschema2pojo:jsonschema2pojo-maven-plugin:0.4.30:generate (default) on project Testing: Execution default of goal org.jsonschema2pojo:jsonschema2pojo-maven-plugin:0.4.30:generate failed: Path not present: com
	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:212)
	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:153)
	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:145)
	at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject(LifecycleModuleBuilder.java:116)
	at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject(LifecycleModuleBuilder.java:80)
	at org.apache.maven.lifecycle.internal.builder.singlethreaded.SingleThreadedBuilder.build(SingleThreadedBuilder.java:51)
[...]

PR #697, opened 2017-03-01
Add option to specify custom separator characters

Tovkal:
This adds the option suggested in #688, so nonstandard JSON Pointers found in $ref can be split correctly.

The option's default is the current value, `#/.`, so it should not break any existing projects.

Let me know if something is missing and needs to be done before merging, also let me know if the wording of the descriptions is a bit weird.

### DIFF
--- a/jsonschema2pojo-ant/src/site/Jsonschema2PojoTask.html
+++ b/jsonschema2pojo-ant/src/site/Jsonschema2PojoTask.html
@@ -409,8 +409,9 @@
         <td align="center" valign="top">No (default <code>false</code>)</td>
     </tr>
     <tr>
-        <td valign="top">customSeparatorCharacters</td>
-        <td valign="top">Separators characters to be used to split JSON Pointers ($ref).
+        <td valign="top">refFragmentPathDelimiters</td>
+        <td valign="top">A string containing any characters that should act as path delimiters when resolving $ref
+            fragments. By default, #, / and . are used in an attempt to support JSON Pointer and JSON Path.
         </td>
         <td align="center" valign="top">No (default <code>#/.</code>)</td>
     </tr>
--- a/jsonschema2pojo-ant/src/site/Jsonschema2PojoTask.html
+++ b/jsonschema2pojo-ant/src/site/Jsonschema2PojoTask.html
@@ -408,6 +408,12 @@
         </td>
         <td align="center" valign="top">No (default <code>false</code>)</td>
     </tr>
+    <tr>
+        <td valign="top">customSeparatorCharacters</td>
+        <td valign="top">Separators characters to be used to split JSON Pointers ($ref).
+        </td>
+        <td align="center" valign="top">No (default <code>#/.</code>)</td>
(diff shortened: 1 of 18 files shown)


joelittlejohn:
Thanks for submitting this. I'll add some comments.

jsonschema2pojo-ant/src/site/Jsonschema2PojoTask.html
```diff
@@ -408,6 +408,12 @@
         </td>
         <td align="center" valign="top">No (default <code>false</code>)</td>
     </tr>
+    <tr>
+        <td valign="top">customSeparatorCharacters</td>
+        <td valign="top">Separators characters to be used to split JSON Pointers ($ref).
```
joelittlejohn: There's a small typo here (_"Separators characters"_). To make it clear that this is not an array or list, I think this text should mention that it's a string. How about this as a description:

> _A string containing any characters that should act as path delimiters when resolving $ref fragments. By default, `#`, `/` and `.` are used in an attempt to support JSON Pointer and JSON Path._

joelittlejohn: This description can be used in a variety of places.

jsonschema2pojo-cli/src/main/java/org/jsonschema2pojo/cli/Arguments.java
```diff
@@ -418,4 +421,9 @@ public boolean isFormatDateTimes() {
         return formatDateTimes;
     }
 
+    @Override
+    public String isCustomSeparatorCharacters() {
```
joelittlejohn: There are a few getters in this and other classes that have an incorrect `is` prefix instead of `get`. This and others should be `get`.

jsonschema2pojo-integration-tests/src/test/java/org/jsonschema2pojo/integration/config/CustomSeparatorCharactersIT.java
```diff
@@ -0,0 +1,41 @@
...
+
+    @Rule public Jsonschema2PojoRule schemaRule = new Jsonschema2PojoRule();
+
+    @Test
+    public void defaultCustomSeparatorCharactersIsStandard() throws ClassNotFoundException, SecurityException, NoSuchMethodException {
+
+        ClassLoader resultsClassLoader = schemaRule.generateAndCompile("/schema/properties/nonStandardRef.json", "com.example",
+                config("customSeparatorCharacters", "#/"));
+
+        assertThat(resultsClassLoader.loadClass("com.example.Foo").getName(), is("com.example.Foo"));
```
joelittlejohn: These assertions are redundant, two plain calls to `loadClass` will suffice as this will throw an exception if the class is missing.

Tovkal:
Applied the requested changes, let me know if there's anything else.

joelittlejohn:
Thanks! :+1:

PR #700, opened 2017-03-06
Don't prefix a class if it already exists without the prefix.

### LINKED ISSUE #650: javaType doesn't work while setting  classNamePrefix = "SomePrefix"
When setting classNamePrefix = "SomePrefix", the following schema will generate a new type java.util.SomePrefix

{
  "javaType": "java.util.Map<String,String>",
   "type": "object"
}

DavidTanner:
Fixes #650 
For the test I needed to use a package other than java.util, or a security exception is thrown.


joelittlejohn:
Thanks a lot for contributing! :+1:

DavidTanner:
@joelittlejohn When will you be cutting the next release?

thachhoang:
@DavidTanner 0.4.32 is on Maven Central now.

joelittlejohn:
Ha, I kept meaning to comment here but didn't get around to it in the last few days. I still need to update the project docs...

Issue #699, opened 2017-03-06
jsonFormat ignored when using Java 8 date & time


CheeseE:
I have some date fields defined  as following:
```
        "effectiveDate": {
          "type": "string",
          "format": "date",
          "customDateTimePattern": "yyyy-MM-dd"
        }
```
In my maven plugin I've configured the plugin to use java.time.* for date fields as following:
```
    <configuration>
      <dateTimeType>java.time.LocalDateTime</dateTimeType>
      <dateType>java.time.LocalDate</dateType>
     </configuration>
```
After generating the classes the jsonFormat annotation is not applied to the classes:
```
    @JsonProperty("effectiveDate")
    private LocalDate effectiveDate;
```
Because of this all the date fields are serialized as array of integers.

joelittlejohn:
I think the new customDateTimePattern only applies for fields that have `"format": "date-time"`:

https://github.com/joelittlejohn/jsonschema2pojo/blob/8c50832998e587f6bed126ab5213df0bd39b9137/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/PropertyRule.java#L143

We should add some further support for `"format": "date"` fields I think.

PR #701, opened 2017-03-07
Extend @JsonFormat annotation to apply to date objects as well.

CheeseE:
Add @JsonFormat annotation to dated fields, it is a fix for #699


joelittlejohn:
We need to do something a little more sophisticated here. We shouldn't ever use `yyyy-MM-dd'T'HH:mm:ss.SSS` when we find `"format": "date"` in the schema.

We need to split these cases out and have a `formatDates` config option (c.f. `formatDateTimes`). We then need two methods on the annotator, `dateField` and `dateTimeField` (the current method needs to be renamed).

Issue #704, opened 2017-03-16
org.jsonschema2pojo.ContentResolverTest require internet connection


s13o:
This unit test is not an integratoinal test. Also a special setting like '-DforkMode=never' required for maven-surefire-plugin when it runs behind firewall. Will be better to start a simple local HTTP-server for such tests

joelittlejohn:
Hi. Could you explain what you mean by:

> Also a special setting like '-DforkMode=never' required for maven-surefire-plugin when it runs behind firewall.

?

joelittlejohn:
Yes you're right about this unit test, it's a little lazy. We could use https://github.com/rest-driver/rest-driver to avoid external http calls.

ctrimble:
@joelittlejohn [WireMock](http://wiremock.org/) has a nice [JUnit rule](http://wiremock.org/docs/junit-rule/) for mocking out http endpoints.

s13o:
Well, the "rest-driver" is an overkill for such simple test.
As I can see the main idea of it is just to test real http connection & exception handling  flow.
I did a pull request #705 - please check it, does it appropriative?

And about the '-DforkMode=never' - I'm working behind a proxy & firewall and it damn unconfortable to configure proxy settings in JVM and they are flashed what tests performed in a forked process...

s13o:
The 'WireMock' is really nice! Many thanks for the link.
But how many http-tests we have? the only one?
We can refactor this test later when we will desire to implement one more

thachhoang:
Fixed by #705.

PR #705, opened 2017-03-17
org.jsonschema2pojo.ContentResolverTest require internet connection (…

s13o:
An attempt to resolve the issue #704 without adding of external dependencies


jsonschema2pojo-core/src/test/java/org/jsonschema2pojo/util/LocalHttpServerBuilder.java
```diff
@@ -0,0 +1,157 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.Executors;
+
+
+/**
+ * @author {@link "https://github.com/s13o" "s13o"}
+ * @since 3/16/2017
+ */
+public class LocalHttpServerBuilder {
```
ctrimble: I know it doesn't seem like one endpoint is worth bringing in a library, but the project will have to lug this implementation around.  We should use some kind of library for this.  This entire PR should be somewhere in the 20 line range.

joelittlejohn: Yes I agree,  particularly as this would be a test dependency only. Let's use wiremock :+1:

s13o: np

s13o:
Updated
wiremock added, become more elegant :)

PR #713, opened 2017-03-26
Fix compilation errors for non-string enums

thachhoang:
Use the correct argument type for the factory method and the constructor of enums with non-string backing types: number and integer.


joelittlejohn:
Thanks @thachhoang :+1: :+1: :+1:

Issue #725, opened 2017-05-12
Schema with required[""] results in String index out of range: 0 


richardwalsh:
Example file:
```
{
    "$schema": "http://json-schema.org/draft-04/schema#",
    "title": "Silly",
    "type":"object",
    "description":"silly example",
    "required": [""],
    "properties": {
        "code" : {
            "type":"string",
            "description":"blah blah blah"
        }
    }
}
```

maven plugin error:
```
Caused by: java.lang.StringIndexOutOfBoundsException: String index out of range: 0
	at java.lang.String.charAt(String.java:658)
	at org.jsonschema2pojo.util.NameHelper.normalizeName(NameHelper.java:45)
	at org.jsonschema2pojo.util.NameHelper.getPropertyName(NameHelper.java:74)
	at org.jsonschema2pojo.rules.RequiredArrayRule.apply(RequiredArrayRule.java:54)
	at org.jsonschema2pojo.rules.RequiredArrayRule.apply(RequiredArrayRule.java:39)
	at org.jsonschema2pojo.rules.ObjectRule.apply(ObjectRule.java:134)
	at org.jsonschema2pojo.rules.ObjectRule.apply(ObjectRule.java:64)
	at org.jsonschema2pojo.rules.TypeRule.apply(TypeRule.java:85)
	at org.jsonschema2pojo.rules.TypeRule.apply(TypeRule.java:35)
	at org.jsonschema2pojo.rules.SchemaRule.apply(SchemaRule.java:73)
	at org.jsonschema2pojo.rules.SchemaRule.apply(SchemaRule.java:66)
	at org.jsonschema2pojo.rules.SchemaRule.apply(SchemaRule.java:30)
	at org.jsonschema2pojo.SchemaMapper.generate(SchemaMapper.java:88)
	at org.jsonschema2pojo.Jsonschema2Pojo.generateRecursive(Jsonschema2Pojo.java:116)
	at org.jsonschema2pojo.Jsonschema2Pojo.generateRecursive(Jsonschema2Pojo.java:118)
	at org.jsonschema2pojo.Jsonschema2Pojo.generate(Jsonschema2Pojo.java:74)
	at org.jsonschema2pojo.maven.Jsonschema2PojoMojo.execute(Jsonschema2PojoMojo.java:494)
	at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo(DefaultBuildPluginManager.java:106)

```
Notes:

[...]

PR #733, opened 2017-05-25
Fix for 725

michalkolodziejski:
In my project I've hit the #725 issue - while working with `machine generated` schema, that contains an empty element in `required` array.

For verification purposes I've used slightly modified schema from #725 (added nonempty element to mix up empty and nonempty ones in array).

```
{
    "title": "Silly",
    "type":"object",
    "description":"silly example",
    "required": ["", "code", ""],
    "properties": {
        "code" : {
            "type":"string",
            "description":"blah blah blah"
        }
    }
}
```

Also the schema was changed `requiredArray.json` (added empty element to `required` array) to make integration test fail when run without the change.

### DIFF
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/RequiredArrayRule.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/RequiredArrayRule.java
@@ -57,6 +57,9 @@ public class RequiredArrayRule implements Rule<JDefinedClass, JDefinedClass> {
 
         for (Iterator<JsonNode> iterator = node.elements(); iterator.hasNext(); ) {
             String requiredArrayItem = iterator.next().asText();
+            if (requiredArrayItem.isEmpty()) {
+                continue;
+            }
 
             JsonNode propertyNode = null;
 
--- a/jsonschema2pojo-integration-tests/src/test/resources/schema/required/requiredArray.json
+++ b/jsonschema2pojo-integration-tests/src/test/resources/schema/required/requiredArray.json
@@ -11,5 +11,5 @@
             "type" : "string"
         }
     },
-    "required": [ "requiredProperty"]
+    "required": [ "requiredProperty", ""]
 }


ctrimble:
The empty string is a valid key in JSON and it can be required.

michalkolodziejski:
@ctrimble the `required` array is for referencing fields (thus keys as You've suggested). What is the purpose of having an empty key (thus valid in JSON) for a property name, and what about duplicacy? Moreover - what would be an expected generated POJO? Empty property with an "empty name" (what about getters and setters then)?

ctrimble:
@michalkolodziejski I don't think that we should shave off parts of the specification that we do not see as useful.  There may be some real technical limitations here, since `""` as a name is foreign to Java, but I am not sure what those are.  The schema you posted has a definition, and it should fail on:

```
{ "code": "some text" }
```

and pass for:

```
{ "code": "some text", "": null }
```

michalkolodziejski:
@ctrimble please see OP for #725 - the real issue is `java.lang.StringIndexOutOfBoundsException: String index out of range: 0` which is raised while processing "empty" Array in`"required": [""],` which I've replaced to conform my situation with `"required": ["", "code", ""]`.

ctrimble:
Yes, the name helper assumes that the string has at least one character.  That means this is probably also broken:

```
{
  "type": "object",
  "properties": {
    "": {
      "type": "string"
    }
  }
}
```

Generating a field and methods for that is doable, when binding annotations are used.  I am not sure how well that will work with the JSR 303 annotations.

joelittlejohn:
Good points. If be interested just to check whether Jackson and Gson can handle this when the right annotations are present. If they can, we can certainly choose a Java name to represent this.

If it's not feasible to support this on Java due to the limitations of other libraries in the ecosystem, I'm happy to ignore the empty string in that list. We aren't a validator, this tool makes a lot of compromises to ensure that it can more often than not offer some generated output that can (at the very least) hold the data that the schema describes.

ctrimble:
So, I just set up a quick test case and it looks like there are a lot more technical issues in here than I thought.  In particular, `@JsonProperty("")` does not mean `a property with key ""` instead it means 'use the field name without modification`.  So:

```
@JsonProperty("")
public void set( String _EMPTY_KEY_ ) {
  this._EMPTY_KEY_ = _EMPTY_KEY_;
}
```

maps to the property `"set"` instead of the property `""`.  It seems like the empty string is a non starter.

ctrimble:
@michalkolodziejski @joelittlejohn sorry for distracting this PR.

LGTM.

joelittlejohn:
Thanks all :+1:

Issue #602, opened 2016-08-02
[Android] Writing to /Reading from Parcelable ignores superclass 


yomik:
With the following sample json files :
**a.json** : 

```
{
  "type": "object",
  "properties": {
      "foo": {
          "type": "string"
      }   
}
```

and **b.json** : 

```
{
  "type": "object",
  "extends" : {
    "$ref": "a.json"
  },
  "properties": {
    "bar": {
      "type": "integer"
    }
  }
}
```

The generated classes will look like : 

**A.java**

```
@JsonInclude(JsonInclude.Include.NON_NULL)
@Generated("org.jsonschema2pojo")
@JsonPropertyOrder({
    "foo"
})
public class A implements Parcelable
{
    @JsonProperty("foo")
    private String foo;
    @JsonIgnore
    private Map<String, Object> additionalProperties = new HashMap<String, Object>();
    public final static Parcelable.Creator<A> CREATOR = new Creator<A>() {

        @SuppressWarnings({
            "unchecked"
        })
        public A createFromParcel(Parcel in) {
            A instance = new A();
            instance.foo = ((String) in.readValue((String.class.getClassLoader())));
            instance.additionalProperties = ((Map<String, Object> ) in.readValue((Map.class.getClassLoader())));
            return instance;
        }

        public A[] newArray(int size) {
            return (new A[size]);
        }

    }
    ;

    /**
     * 
     * @return
     *     The foo
     */
    @JsonProperty("foo")
    public String getFoo() {
        return foo;
    }

    /**
     * 
     * @param foo
     *     The foo
     */
    @JsonProperty("foo")
    public void setFoo(String foo) {
        this.foo = foo;
    }

    @JsonAnyGetter
    public Map<String, Object> getAdditionalProperties() {
        return this.additionalProperties;
    }

    @JsonAnySetter
    public void setAdditionalProperty(String name, Object value) {
        this.additionalProperties.put(name, value);
    }

    public void writeToParcel(Parcel dest, int flags) {
        dest.writeValue(foo);
        dest.writeValue(additionalProperties);
    }

    public int describeContents() {
        return  0;
    }

}
```

and **B.java**

```
@JsonInclude(JsonInclude.Include.NON_NULL)
@Generated("org.jsonschema2pojo")
@JsonPropertyOrder({
    "bar"
})
public class B
    extends A
    implements Parcelable
{

    @JsonProperty("bar")
    private long bar;
    @JsonIgnore
[...]

PR #737, opened 2017-06-01
Fix for #602 : Writing to /Reading from Parcelable ignores superclass

yomik:
When a generated `Parcelable` model subclasses another `Parcelable` model, the subclass needs to call its parent to allow the parent's properties to be written to/read from the `Parcel`
- Calls `super.writeToParcel` if the parent implements `Parcelable`
- Creates a protected constructor with a Parcel as a single parameter that reads from the `Parcel`, calling `super(in)` if the parent implements `Parcelable`
- Moved the read from `Parcel` code that used to be into `CREATOR#createFromParcel` to the newly created constructor; `createFromParcel` now just returns `new Ctor(in)` instead.
- Explicitely added an empty constructor since the newly added ctor shadows the default empty one

### DIFF
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/ObjectRule.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/ObjectRule.java
@@ -158,6 +158,9 @@ public class ObjectRule implements Rule<JPackage, JType> {
         parcelableHelper.addWriteToParcel(jclass);
         parcelableHelper.addDescribeContents(jclass);
         parcelableHelper.addCreator(jclass);
+        parcelableHelper.addConstructorFromParcel(jclass);
+        // Add empty constructor
+        jclass.constructor(JMod.PUBLIC);
     }
 
     /**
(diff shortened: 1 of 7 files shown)


joelittlejohn:
@yomik Great, thanks for this. Can you add an integration test to demonstrate what you've described and ensure that it is fixed? I don't want some regression here in future to break your fix.

joelittlejohn:
Check out [ParcelableIT.java](https://github.com/joelittlejohn/jsonschema2pojo/blob/069cb7565acdf0fe2bedb1320b4c2e9ccce9a5fd/jsonschema2pojo-integration-tests/src/test/java/org/jsonschema2pojo/integration/config/ParcelableIT.java).

yomik:
Sorry, it was week-end already here :). Will do. Thanks

yomik:
The current test would (for some unknown reason) work even with an extends declaration.
I added a new test case that explicitely call writeToParcel and createFromParcel on the Parcelable object.
Before : the test would fail equality; the parent class fields are not set
After : the test passes, parent properties are now correctly set

joelittlejohn:
@yomik I'm seeing an error in the integration test:
```
Tests in error: 
  parcelableSuperclassIsUnparceled(org.jsonschema2pojo.integration.config.ParcelableIT): Unrecognised URI, can't resolve this: file:/home/jl/workspace/jsonschema2pojo/jsonschema2pojo-integration-tests/target/test-classes/schema/parcelable/parcelable-schema-base.json
```

yomik:
Oops, sorry about that. I forgot to do a clean before running the tests

yomik:
On a side-note, the file jsonschema2pojo-integration-tests/src/test/java/org/jsonschema2pojo/integration/config/IncludeToStringExcludesIT.java keeps getting modified because it lacks the proper copyright headers. I didn't include it in this PR since it is unrelated but just FYI

PR #739, opened 2017-06-01
Update recommended jackson version gradle plugin

JLLeitschuh:
The version suggested in the Gradle Plugin README is a version of jackson that doesn't have the `com.fasterxml.jackson.annotation.JsonPropertyDescription` annotation.
The new recommended version has the annotation.


joelittlejohn:
Thanks :+1:

Issue #742, opened 2017-06-12
Generating 2 constructors


locofrank:
The latest build was generating 2 constructors for several of my classes

I needed to roll back to 0.4.35 to fix my project

joelittlejohn:
Are you using Parcelable classes by any chance?

joelittlejohn:
And are the constructors duplicates? Do the classes compile? It would be useful to have some more detail here. Thanks!

yomik:
Oh, I think I'm the one to blame for that.
I reproduced this by setting the includeConstructors = true in the configuration. I overlooked that one.
I'm going to provide a fix for this, but now I wonder what other settings might affect the changes....
For instance, the code refers to isConstructorsRequiredPropertiesOnly, which is not in the documentation (some configuration that was added recently ?).

Anyway, I'm trying to fix this ASAP, and try to see what other side effects might come from other settings as well.

Sorry for the trouble

joelittlejohn:
Thanks for reporting this quickly @locofrank. Closed by #743.

joelittlejohn:
I'll release 0.4.37 shortly.

locofrank:
Wow! You folks are on top of it. Sorry I didn't respond quickly. Thanks for the quick fix.

PR #743, opened 2017-06-13
Fix for #742 : Generating 2 constructors

yomik:
Fix regression introduced by #602 when configuration option `includeConstructors` is true : empty default constructor added twice.

Updated existing integration test and added another one; test Parcelable generation with the option on and off to ensure compilation and serialization/deserialization work properly.

Please let me know if there other configurations that may generate an empty constructor (I looked through the code and couldn't find any).


joelittlejohn:
Thanks for jumping on this so quickly @yomik! Much appreciated.

PR #759, opened 2017-07-18
Getters and Setters can now be activated independently

lstav:
Made it possible for getters and setters to be activated independently. I am not sure that I did the dynamic setters/getters correctly, so any feedback will be appreciated. I would also like feedback on the testing.

### DIFF
--- a/jsonschema2pojo-ant/src/main/java/org/jsonschema2pojo/ant/Jsonschema2PojoTask.java
+++ b/jsonschema2pojo-ant/src/main/java/org/jsonschema2pojo/ant/Jsonschema2PojoTask.java
@@ -135,11 +135,15 @@ public class Jsonschema2PojoTask extends Task implements GenerationConfig {
 
     private boolean includeAdditionalProperties = true;
 
+    private boolean includeAccessors = true;
+
     private boolean includeGetters = true;
 
     private boolean includeSetters = true;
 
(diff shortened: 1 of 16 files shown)


joelittlejohn:
Thanks for submitting this PR @lstav. It looks like you have been very thorough.

One issue here is that you have removed the existing config option, and I can see why - it's not easy to make these play nicely together, particularly considering the default values. We may only be able to include this in a 1.0 release.

I wonder if we can find a way to include this but keep the `includeAccessors` option. We can even deprecate it. 

So, add getters when:

```java
isIncludeAccessors() || isIncludeGetters()
```

And add setters when:

```java
isIncludeAccessors() || isIncludeGetters()
```

This way we maintain the old behaviour but can support your new features too.

lstav:
@joelittlejohn Sounds good, I'll add those and update the PR.

lstav:
@joelittlejohn Just added the accessor methods again, let me know if this is what you were expecting. Also, I wasn't sure if I should put @Deprecated on the accessor methods, or if we would just add them later.

joelittlejohn:
I think includeGetters and includeSetters should be default `false`. This way, users who are currently using includeAccessors=`false` will not have getters/setters appear in their code after this change.

lstav:
I agree, that's a good idea, I just pushed the changes.

PR #792, opened 2017-10-11
Use FilenameUtils to fix relative source paths (and remove extra underscore classes that appear when using a relative source path)

dhalbrook:
Use FilenameUtils to fix relative source paths in the maven plugin.  This in turn fixes the extra "underscore" classes that get generated with a relative source path.


joelittlejohn:
Thanks for the fix! Much appreciated.

dhalbrook:
Sure thing.  Wish I could figure out a good way to test it.  I found this one out the hard way because our build system is set with a maven basedir with "/./" in it... so I'm motivated to get it fixed!

PR #801, opened 2017-11-13
Configuration option to force camel case

antoniobunicic:
Added configuration option to force camel case. It can be used when JSON field names (or schema name) contain uppercase sequences.

For example, if we have `EXAMPLE_SCHEMA.json` and `EXAMPLE_FIELD`, they become `EXAMPLESCHEMA.java` with property `eXAMPLEFIELD`, but we usually want something like `ExampleSchema.java` with property `exampleField`. 

What's needed to make this plugin work properly in these cases is just converting them to lowercase before generating class and field names, as well as getter and setter methods. 

That was added in this pull request, and the plugin now features a boolean configuration option 'forceCamelCase', _false_ by default.


prabu-manickam:
Thanks for adding this feature! I always wanted it. Unknowingly created an issue #1020 to suggest the same.

ctrimble:
@antoniobunicic This PR has been sitting for some time.  I am going to close this out.  If you want to pick it back up, let me know.

PR #810, opened 2017-11-26
add partial support of wildcards in "javaType" declarations

wigbam:
Added support for the following syntax in "javaType" declarations:
- `?`
- `? extends ...`

Unfortunately, due to limitations of [Sun JCodeModel](https://github.com/javaee/jaxb-codemodel/blob/master/codemodel/codemodel/src/main/java/com/sun/codemodel/JTypeWildcard.java) it is impossible to add support for `? super ...`

Maybe it is time to think about using a more comprehensive and better maintained [JCodeModel fork](https://github.com/phax/jcodemodel/)?


joelittlejohn:
Ah, nice change, thank you!

That fork looks great, I think we should switch.

PR #812, opened 2017-12-08
Replace commons-lang equals, hashCode, and toString with generated code

ehrmann:
commons-lang is a *relatively* large library for users to include just for adding `toString()`, `equals()`, and `hashCode()`. This removes that dependency and uses generated code instead.

It avoids calls like `Objects.hash()`, `Double.hashCode()`, and `Objects.equals()` that were added in later versions of Java.

Generated code looks something like this:
```
public class ComplexTypesArray {
    @Override
    public String toString() {
        StringBuilder sb = new StringBuilder();
        sb.append(ComplexTypesArray.class.getName())
                .append('@')
                .append(Integer.toHexString(System.identityHashCode(this)))
                .append('[');
        sb.append("additionalProperties");
        sb.append('=');
        sb.append(((this.additionalProperties == null)?"<null>":this.additionalProperties));
        sb.append(',');
        if (sb.charAt((sb.length()- 1)) == ',') {
            sb.setCharAt((sb.length()- 1), ']');
        } else {
            sb.append(']');
        }
        return sb.toString();
    }

    @Override
    public int hashCode() {
        int result = 1;
        result = ((result* 31)+((this.additionalProperties == null)? 0 :this.additionalProperties.hashCode()));
        return result;
    }

    @Override
    public boolean equals(Object other) {
        if (other == this) {
            return true;
        }
        if ((other instanceof ComplexTypesArray) == false) {
            return false;
        }
        ComplexTypesArray rhs = ((ComplexTypesArray) other);
        return ((this.additionalProperties == rhs.additionalProperties)||((this.additionalProperties!= null)
                && this.additionalProperties.equals(rhs.additionalProperties)));
    }
}
```

### DIFF
--- a/jsonschema2pojo-ant/src/main/java/org/jsonschema2pojo/ant/Jsonschema2PojoTask.java
+++ b/jsonschema2pojo-ant/src/main/java/org/jsonschema2pojo/ant/Jsonschema2PojoTask.java
@@ -32,6 +32,7 @@ import java.util.Iterator;
 import java.util.List;
 
 import org.apache.tools.ant.BuildException;
+import org.apache.tools.ant.Project;
 import org.apache.tools.ant.Task;
 import org.apache.tools.ant.types.Path;
 import org.apache.tools.ant.types.Reference;
@@ -117,8 +118,6 @@ public class Jsonschema2PojoTask extends Task implements GenerationConfig {
     private boolean useJodaLocalDates = false;
 
     private boolean useJodaLocalTimes = false;
-
-    private boolean useCommonsLang3 = false;
 
     private boolean parcelable = false;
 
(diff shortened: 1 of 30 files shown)


ehrmann:
I wasn't able to get jsonschema2pojo-gradle-plugin to pass tests. It looks like 

`classpath 'org.jsonschema2pojo:jsonschema2pojo-gradle-plugin:latest.integration'` in `jsonschema2pojo-gradle-plugin/example/java/build.gradle` actually pulls in the latest *repository* version of `jsonschema2pojo-*` rather than the one from the project.

joelittlejohn:
Great stuff. Thanks for this! Are you aware of any functional difference between the commons-lang implementations and the code you have generated?

ehrmann:
I should look into in more (or add more tests). I tried to match the behavior of `toString`. `Equals` should generally do the right thing.

`HashCode` is probably different. The only time I can think of a change to `hashCode` making a difference is a serialized HashMap. Luckily, [OpenJDK's HashMap](http://hg.openjdk.java.net/jdk7/jdk7/jdk/file/9b8c96f96a0f/src/share/classes/java/util/HashMap.java#l1025) rebuilds the map on deserialization. Since this doesn't change any fields or method signatures, these classes will have the same `SerialVersionUID`.

joelittlejohn:
Great. I don't think identical behaviour is required, here (particularly, as you say, for `hashCode`). I only ask so that we can understand any difference that may exist, and also preempt support questions.

I'll merge this for the next release.

ehrmann:
I might add a few tests to check the string format.

ehrmann:
I added a test class for toString as a separate commit. It looks like the output is generally compatible. The might be an edge case, but it's really close.

ehrmann:
@joelittlejohn Any idea when you'll make the next release?

PR #820, opened 2017-12-25
Fix failing tests

aldaris:
There were two tests failing:
* one in SerializationTest, because OSX's case insensitive file system found a different Superclass definition than the SuperClass it was meant to find
* one in CustomDateTimeFormatIT because the JSON serializer was using the system locale and the test was expecting results with english locale.


joelittlejohn:
Brilliant, thank you!

PR #821, opened 2017-12-26
Inconsistencies in Maven POMs

aldaris:
Looks like there are dependency version inconsistencies in the latest POM:

```
Dependency convergence error for org.apache.commons:commons-lang3:3.0.1 paths to dependency are:
  +-org.jsonschema2pojo:jsonschema2pojo-core:0.5.1
    +-org.jsonschema2pojo:jsonschema2pojo-scalagen:0.5.1
      +-org.apache.commons:commons-lang3:3.0.1
and
  +-org.jsonschema2pojo:jsonschema2pojo-core:0.5.1
    +-org.apache.commons:commons-lang3:3.2.1
```

It appears that the scalagen module uses a different version of commons-lang3 than the core module.

The below commits should address this and also a few minor tweaks to the POM to better use managed dependencies from the root POM.

### DIFF
--- a/jsonschema2pojo-scalagen/pom.xml
+++ b/jsonschema2pojo-scalagen/pom.xml
@@ -15,9 +15,7 @@
 
     <properties>
         <commons.collections.version>4.01</commons.collections.version>
-        <commons.lang.version>3.0.1</commons.lang.version>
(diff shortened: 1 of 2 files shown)


joelittlejohn:
Thanks!

PR #822, opened 2018-01-02
Add commons-lang3 version that used to be defined in parent pom.

metacosm:


joelittlejohn:
Great, thanks. I got distracted before I could push a fix for this last night.

Issue #832, opened 2018-02-08
Custom maven repository key leads to timeouts within custom repository structure


DennisRippinger:
When the plugin is executed within a strict nexus/artifactory environment it tries to download dependencies from custom repositories which are not reachable. 
```
[INFO] --- jsonschema2pojo-maven-plugin:0.5.1:generate (default) @ myapp ---
Downloading: http://repo1.maven.org/maven2/io/grpc/grpc-core/maven-metadata.xml
Downloading: https://oss.sonatype.org/content/repositories/snapshots/io/grpc/grpc-core/maven-metadata.xml
Downloading: http://repo.spring.io/plugins-release/io/grpc/grpc-core/maven-metadata.xml
```
The parent pom defines two (with the oss-parent three) custom repositories. One of them is maven central with the key `maven-central`. I assume the default is explicitly listed regarding some problem with gradle ([original commit](https://github.com/joelittlejohn/jsonschema2pojo/commit/a9d697e0066293c0c34d712fb1fe5559ba84c545)). 

I would suggest changing the key to `central`, which is the default key and usually handled in a nexus/artifactory environment as mirror.

joelittlejohn:
Closed by #833

PR #833, opened 2018-02-08
#832 Changed maven central repo listing id key to default

DennisRippinger:
As far as I understand the need for the explicit repository, this could be a solution in strict nexus/artifactory environments.


joelittlejohn:
Thanks! I haven't operated in a strict environment like this for a long time so it's useful to have your insight here.

PR #913, opened 2018-09-12
Improve Optionals for Getters code to allow for specifying on a field level

mbramwell1:
Change the Use Optionals for Getters functionality to allow for specifying to use optionals on a field level aswell as for the whole schema. Not a breaking change.

### DIFF
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/PropertyRule.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/PropertyRule.java
@@ -77,7 +77,7 @@ public class PropertyRule implements Rule<JDefinedClass, JDefinedClass> {
         ruleFactory.getAnnotator().propertyField(field, jclass, nodeName, node);
 
         if (isIncludeGetters) {
-            JMethod getter = addGetter(jclass, field, nodeName, node, isRequired(nodeName, node, schema));
+            JMethod getter = addGetter(jclass, field, nodeName, node, isRequired(nodeName, node, schema), usesOptional(nodeName, node, schema));
             ruleFactory.getAnnotator().propertyGetter(getter, jclass, nodeName);
             propertyAnnotations(nodeName, node, schema, getter);
         }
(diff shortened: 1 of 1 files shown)


mbramwell1:
At the moment you can only specify using optionals on a global level, which is fine I guess but does lead to extra code where it sometimes isnt needed. This cange allows for both use on a global level, or specifying to use optional on a field level by providing "java_optional": true

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/PropertyRule.java
```diff
@@ -129,6 +129,24 @@ private boolean isRequired(String nodeName, JsonNode node, Schema schema) {
         return false;
     }
 
+    private boolean usesOptional(String nodeName, JsonNode node, Schema schema) {
+        if (node.has("java_optional")) {
```
joelittlejohn: Could you rename these to `javaOptional`, as this is the naming convention for our extension properties.

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/PropertyRule.java
```diff
@@ -129,6 +129,24 @@ private boolean isRequired(String nodeName, JsonNode node, Schema schema) {
...
+            final JsonNode requiredNode = node.get("java_optional");
+            return requiredNode.asBoolean();
+        }
+
+        JsonNode requiredArray = schema.getContent().get("java_optional");
```
joelittlejohn: I think you mean `javaOptionalArray` here? `requiredArray` is a name which I assume was taken from another method.

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/PropertyRule.java
```diff
@@ -77,7 +77,7 @@ public JDefinedClass apply(String nodeName, JsonNode node, JDefinedClass jclass,
         ruleFactory.getAnnotator().propertyField(field, jclass, nodeName, node);
 
         if (isIncludeGetters) {
-            JMethod getter = addGetter(jclass, field, nodeName, node, isRequired(nodeName, node, schema));
+            JMethod getter = addGetter(jclass, field, nodeName, node, isRequired(nodeName, node, schema), usesOptional(nodeName, node, schema));
```
joelittlejohn: Could you name `usesOptional` to `useOptional` throughout? I think this scans better.

mbramwell1:
No probs. Review comments addressed.

Issue #908, opened 2018-09-07
Allow schema title to be used as class name


jrehwaldt:
When having schemas where certain field names are repeated with different implementations the code currently generated unique class names such as `Image__1`, etc. I would like to have a configuration to use the `title` as class name instead.

So that 
```
{
  "$schema": "http://json-schema.org/draft-04/schema#",
  "title": "SchemaTest body",
  "description": "Use title as class name.",
  "type": "object",
  "properties": {
    "image": {
      "type": "object",
      "title": "MyImage"
    }
  }
}
```

would result in

```
public class SchemaTest {

    private MyImage image;

    public SchemaTest() {
    }

    public SchemaTest(MyImage image) {
        super();
        this.image = image;
    }

    public MyImage getImage() {
        return image;
    }
}
```

```
public class MyImage{
}
```

Thank you for your great library. I am using it successfully and really like the `RuleFactory` design.

joelittlejohn:
This has been asked for before. Interesting that in your design you expect `"SchemaTest body"` to become `SchemaTest`. Was that a typo or do you expect jsonschema2pojo to do some transformation/sanitization here?

jrehwaldt:
In my implementation I only use the title for sub-hierarchies and stick with the filename otherwise. Not a deal breaker for me in general.

I would capitalize all first letters and strip spaces. Next illegal characters get replaced with underscore and that’s it.

jrehwaldt:
I am willing to contribute PRs for all my requests if there’s a chance for approval.

joelittlejohn:
Happy to accept a PR for this, as long as it is behind a configuration option, say, `useTitleAsClassname`. The default must be `false`.

jrehwaldt:
Just saw your comment now. If you prefer a boolean property I can change it.

I implemented it as part of `NameHelper` for now. It seeemed the best place to me as it also works for enums this way.

jrehwaldt:
Thank you very much! It was fun working on this one.

PR #916, opened 2018-09-14
Allow to resolve type names from `title` attribute in schema.

jrehwaldt:
Each word in a title is capitalized and whitespaces stripped.

Fixes gh-908


jrehwaldt:
I removed some unrelated changes by amending my initial commit. Future changes will be in separate commits obviously.

joelittlejohn:
Hmmm. Yes, please implement this as a simple Boolean. I think these are a lot easier for people to use.

Also, could you use the existing `replaceIllegalCharacters` method to sanitise the string. It replaces more than just the space char.

Cheers.

jrehwaldt:
The method `replaceIllegalCharacters` is called from `ObjectRule|EnumRule#getClassName`.

I refined the logic and added an IT.

Issue #910, opened 2018-09-07
Generate URL when format=uri and pattern=https?://


jrehwaldt:
This request is maybe a little esoteric, but I figured it doesn't hurt putting it out here.

When a field is spec'ed as follows `{"type": "string", "format": "uri", "pattern": "https?://" }` the derived type is `java.net.URI` due to the format specifier. Would you consider a PR the derives `java.net.URL` instead when `pattern` _starts with_ `"https?://"` and `format` is `uri`?

```
{
  "$schema": "http://json-schema.org/draft-04/schema#",
  "title": "SchemaTest body",
  "description": "Use URL instead of URI in special cases.",
  "type": "object",
  "properties": {
    "url": {
      "type": "string",
      "format": "uri",
      "pattern": "https?://"
    }
  }
}
```

Issue #909, opened 2018-09-07
Allow to specify generated additional interfaces via configuration


jrehwaldt:
I would like to add interfaces to all my generated classes without adding `javaInterfaces` to my schema definitions. Would you be willing to take a PR that adds a `GenerationConfig` option `additionalInterfaces` (or something like that), which are added to all generated root types?

jrehwaldt:
I guess this won't happen, so closing it.

Issue #906, opened 2018-09-07
Incorrect JSR305 annotations generated


jrehwaldt:
Currently, the rule `NotRequiredRule` generates incorrect `@Nullable` annotations for fields that get correctly generated with `@Nonnull`. This happens only on nested objects, due to [broken logic for checking the `required` array definition](https://github.com/joelittlejohn/jsonschema2pojo/blob/3b80b3452f3d34901ee1e1e886c946c6986aaec1/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/NotRequiredRule.java#L70). When retrieving the schema content we get the root schema instead of the relevant sub schema and therefore the implementation checks against an incorrect hierarchy within the schema.

# Schema
```
{
  "$schema": "http://json-schema.org/draft-04/schema#",
  "title": "SchemaTest body",
  "description": "Example for incorrect @Nullable annotations.",
  "type": "object",
  "properties": {
    "sub-array": {
      "type": "array",
      "items": {
        "type": "object",
        "required": [
          "nonnull-field"
        ],
        "properties": {
          "nonnull-field": {
            "type": "string"
          },
          "nullable-field": {
            "type": "string"
          }
        }
      }
    }
  }
}

```

# Behaviour
```
package tests;

import java.util.ArrayList;
import java.util.List;
import javax.annotation.Nullable;
import javax.validation.Valid;


/**
 * SchemaTest body
 * <p>
 * Example for incorrect @Nullable annotations.
 * 
 */
public class SchemaTest {

    /**
     * 
     * (Can be null)
     * 
     */
    @Nullable
    @Valid
    private List<SubArray> subArray = new ArrayList<SubArray>();

    /**
     * No args constructor for use in serialization
     * 
     */
    public SchemaTest() {
    }

    /**
     * 
     * @param subArray
     */
    public SchemaTest(List<SubArray> subArray) {
        super();
        this.subArray = subArray;
    }

    public List<SubArray> getSubArray() {
        return subArray;
    }

    @Override
    public String toString() {
        StringBuilder sb = new StringBuilder();
        sb.append(SchemaTest.class.getName()).append('@').append(Integer.toHexString(System.identityHashCode(this))).append('[');
        sb.append("subArray");
        sb.append('=');
        sb.append(((this.subArray == null)?"<null>":this.subArray));
        sb.append(',');
        if (sb.charAt((sb.length()- 1)) == ',') {
[...]

PR #917, opened 2018-09-17
Enable development of context-aware rules.

jrehwaldt:
Add JsonNode `parent` to rules.

This is required in order to implement context-aware rules, e.g. such as gh-910.

Breaking change.


jrehwaldt:
This also helps implementing gh-909 and fixing gh-906. I've also seen previous PRs neeing similar functionality (e.g. gh-622).

I am open to implement another approach if you are in favor of another solution.

joelittlejohn:
I think you're right that it would be good to get this change added before 1.0.0 is released :+1:

jrehwaldt:
Another refinement would be to add a general `Context` argument instead, which could then be extended further in the future. Part of this would be this parent, or maybe the whole branch (parent->parent's parent->...). Just putting it out as suggestion. Not sure if this would be better.

Something along the lines of

```
interface Context {
	JsonNode parent()
	JsonNode current()
    // extensible in future
}
```

jrehwaldt:
Thanks for considering and merging! :)

PR #918, opened 2018-09-17
Fields with format "uri" and pattern starting with "https?://" result in type java.net.URL

jrehwaldt:
A field is spec'ed as follows {"type": "string", "format": "uri", "pattern": "https?://..." } derives the type is java.net.URL.

Depends on gh-917 and is a breaking change (API and resulting generation).


joelittlejohn:
I don't think I can merge this one I'm afraid, though thank you for sharing the code for others to see.

You have your own definition of a URL here that is not exhaustive. URLs do not need to match `https?://`. Java supports http, https, ftp, file, and jar but additional stream handlers can be specified. Which of these should lead to URL being used? I understand that you are trying to restrict the allowable values further than java.net.URI, but this isn't generalisable into jsonschema2pojo in the way that you have defined it here IMO.

You can of course solve this by using `javaType`.

jrehwaldt:
Fair enough. That's why I opened an issue about this exact situation a week ago.

I understand the list is not exhaustive, but it doesn't invalidate the implementation. If at all, it makes it incomplete.

Actually, I didn't expect it to be merged, but figured it would make sense to propose to change the Rule API before you release 1.0.0, because as it stands context-aware rules are not possible. Maybe I have luck with this one...

jrehwaldt:
What about a schema-neutral implementation checking for something like this `...://...`?

jrehwaldt:
I generalized it . Feel free to close if undesired nonetheless.

PR #922, opened 2018-09-18
Derive child schemas for objects and arrays

jrehwaldt:
Derive child schemas for objects and arrays. This approach seems to be suggested by the JavaDoc of SchemaRule already.

The relevant bits are in `SchemaRule#L79`, where a child schema is obtained and passed down the rule tree as the schema for the relevant section. This seems to be the expected behavior by `NotRequiredRule`, among others.

Changes `Schema` to keep reference to its parent `Schema` instead of parent `JsonNode`. This means `Schema(id, JsonNode content, JsonNode parentContent)` is replaced by `Schema(id, JsonNode content, Schema parent)`.

`Schema#getParent` still returns _self_ as did `#getParentContent` previously.

Fixes gh-906 (JSR305 bug) and depends on gh-921 (integration tests for JSR305 bug)

### DIFF
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/Schema.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/Schema.java
@@ -28,13 +28,13 @@ public class Schema {
 
     private final URI id;
     private final JsonNode content;
-    private final JsonNode parentContent;
+    private final Schema parent;
     private JType javaType;
 
-    public Schema(URI id, JsonNode content, JsonNode parentContent) {
+    public Schema(URI id, JsonNode content, Schema parent) {
         this.id = id;
         this.content = content;
-        this.parentContent = parentContent;
+        this.parent = parent != null ? parent : this;
     }
 
     public JType getJavaType() {
@@ -59,12 +59,20 @@ public class Schema {
         return content;
     }
 
-    public JsonNode getParentContent() {
-        return parentContent;
+    public Schema getParent() {
+        return parent;
     }
     
     public boolean isGenerated() {
         return javaType != null;
+    }
+
+    public Schema derive(JsonNode content) {
+        if (content != this.content) {
+            return new Schema(id, content, this);
+        } else {
+            return this;
+        }
     }
 
 }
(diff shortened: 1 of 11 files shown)


jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/Schema.java
```diff
@@ -59,12 +59,20 @@ public JsonNode getContent() {
...
-        return parentContent;
+    public Schema getParent() {
+        return parent;
     }
     
     public boolean isGenerated() {
         return javaType != null;
     }
 
+    public Schema derive(JsonNode content) {
```
joelittlejohn: Could you add some javadoc to this method? I don't think it is self-explanatory based on the name alone.

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/Schema.java
```diff
@@ -28,13 +28,13 @@
...
-    private final JsonNode parentContent;
+    private final Schema parent;
     private JType javaType;
 
-    public Schema(URI id, JsonNode content, JsonNode parentContent) {
+    public Schema(URI id, JsonNode content, Schema parent) {
         this.id = id;
         this.content = content;
-        this.parentContent = parentContent;
+        this.parent = parent != null ? parent : this;
```
joelittlejohn: Could you help me understand why this is needed? If there is no parent, why is the parent 'this'?

joelittlejohn:
I think this is a good change and it can't have been easy to work out exactly how to tackle this. Your changes appear minimal and exact - thank you for submitting the PR. I've added a couple of questions I hope to merge this soon.

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/Schema.java
jrehwaldt: This is to mimic the logic you used before. Previously, the _parent_ in `parentContent` didn't really mean _parent or null if none_, but rather _parent or self aka "root"_. I adopted this style, because it meant minimal code changes and frees the caller of `getParent` from performing null-checks. I have no personal opinion whether `null` or `this` is better or worse.

PR #964, opened 2019-03-26
Check schema required array before is required flag

newmen:
Hi,

The generated code is incorrect when a required field is an object, and it has required internal fields. It because of wrong algorithm checks is required flag of the field previously than tests a top defined required fields list. I've changed the checking order.

I mean the case where is the top schema follows:

```json
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "type": ["object", "null"],
  "required": [
    "payment_type",
    "change_alignment",
    "billing_alignment"
  ],
  "properties": {
    "payment_type": {
      "type": "object",
      "$ref": "payment_type.json"
    },
    "change_alignment": {
      "type": "object",
      "$ref": "change_alignment.json"
    },
    "billing_alignment": {
      "type": "object",
      "$ref": "billing_alignment.json"
    }
  }
}
```

and `billing_alignment.json` is:

```json
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "type": "object",
  "required": [
    "date_type"
  ],
  "properties": {
    "date_type": {
      "type": "string",
      "enum": [
        "activation",
        "anniversary"
      ]
    },
    "day": {
      "type": "integer",
      "minimum": 1,
      "maximum": 28
    }
  }
}
```

For this example, the wrong plugin with `useOptionalForGetters=true` generates `Optional<BillingAlignment>` type of `getBillingAlignment` method instead of just `BillingAlignment`.
I've fixed it and now the type is `BillingAlignment`.

### DIFF
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/PropertyRule.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/PropertyRule.java
@@ -112,16 +112,10 @@ public class PropertyRule implements Rule<JDefinedClass, JDefinedClass> {
         return jclass;
     }
 
-    private boolean isRequired(String nodeName, JsonNode node, Schema schema) {
-        if (node.has("required")) {
-            final JsonNode requiredNode = node.get("required");
-            return requiredNode.asBoolean();
-        }
-
-        JsonNode requiredArray = schema.getContent().get("required");
-
-        if (requiredArray != null) {
-            for (JsonNode requiredNode : requiredArray) {
+    private boolean hasEnumerated(Schema schema, String arrayFieldName, String nodeName) {
+        JsonNode array = schema.getContent().get(arrayFieldName);
+        if (array != null) {
+            for (JsonNode requiredNode : array) {
                 if (nodeName.equals(requiredNode.asText()))
                     return true;
             }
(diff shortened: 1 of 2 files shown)


jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/PropertyRule.java
```diff
@@ -130,22 +122,25 @@ private boolean isRequired(String nodeName, JsonNode node, Schema schema) {
...
+            final JsonNode requiredNode = node.get(fieldName);
             return requiredNode.asBoolean();
         }
 
-        JsonNode javaOptionalArray = schema.getContent().get("javaOptional");
+        return false;
+    }
+
+    private boolean isDeclaredAs(String type, String nodeName, JsonNode node, Schema schema) {
+        return hasEnumerated(schema, type, nodeName) || hasFlag(node, type);
```
newmen: You can change the checking order here if you feel the increasing performance here.
We can do that because `hasFlag` is returning boolean independently and if it returns `false` then we check the top `required` array. In the previous implementation, we did not check the top `required` array if the test of flag returned `false.

newmen:
@joelittlejohn, when do you want to release `1.0.1`?

joelittlejohn:
It should be available on Maven Central now. Have a try.

newmen:
Awesome!

Issue #977, opened 2019-05-02
Add support for @Digits JSR-303 annotation


john-tipper:
Currently there is no support for the `@Digits` annotation (https://javaee.github.io/javaee-spec/javadocs/javax/validation/constraints/Digits.html), which is part of the JSR-303 standard (example here: https://beanvalidation.org/1.0/spec/#validationapi-message-examples).

It would be good to be able to validate BigDecimal values, which is a common requirement for financial services applications.

I'd propose having a `digits` field that is associated with a property, which in turn would require `integerDigits` and `fractionalDigits` to be set within that `digits` field.

For example (this assumes that the config option of `isUseBigDecimals()` returns true).
```
{
    "type" : "object",
    "properties" : {
        "decimal" : {
            "type" : "number",
            "digits" : {
                "integerDigits": 5,
                "fractionalDigits": 10
            }
        }
    }
}
```

john-tipper:
PR implementing the above is at #978.

PR #978, opened 2019-05-02
977 add digits rule

john-tipper:
Fix to #977.

Implementation notes:

Requires use of a `digits` field that is associated with a property, which in turn requires `integerDigits` and `fractionalDigits` to be set within that `digits` field.

For example (this assumes that the config option of `isUseBigDecimals()` returns true).

```
{
    "type" : "object",
    "properties" : {
        "decimal" : {
            "type" : "number",
            "digits" : {
                "integerDigits": 5,
                "fractionalDigits": 10
            }
        }
    }
}
```
Validation of `integerDigits` and `fractionalDigits` is currently restricted to only checking that they exist.  In the event that either does not, DigitsRule will throw an `NoSuchElementException`.

### DIFF
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/DigitsRule.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/DigitsRule.java
@@ -36,17 +36,12 @@ public class DigitsRule implements Rule<JFieldVar, JFieldVar> {
     public JFieldVar apply(String nodeName, JsonNode node, JsonNode parent, JFieldVar field, Schema currentSchema) {
 
         if (ruleFactory.getGenerationConfig().isIncludeJsr303Annotations()
-            && node.has("digits")) {
+            && node.has("integerDigits") && node.has("fractionalDigits")) {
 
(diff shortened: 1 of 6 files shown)


joelittlejohn:
I like this addition, but the `"digits"` object seems unnecessary to me (an extra layer of nesting with no particular benefit). Can you just support integerDigits and fractionDigits at the property level?

john-tipper:
Sure, no problem - how would you prefer to treat the 2 properties with respect to validation of their values? @Digits requires both values, so if one is present but not another then throw?

Sent from my iPhone

On 2 May 2019, at 13:39, Joe Littlejohn <notifications@github.com<mailto:notifications@github.com>> wrote:


I like this addition, but the "digits" object seems unnecessary to me (an extra layer of nesting with no particular benefit). Can you just support integerDigits and fractionDigits at the property level?

joelittlejohn:
Yes, sounds good. They must both be omitted or both be included :+1:

john-tipper:
Updated PR to require 2 properties in order to introduce a `@Digits` constraint, as per @joelittlejohn's suggestion.
Failure mode in the event only one property is specified is to do nothing.

PR #981, opened 2019-05-13
bugfix: 'java.lang.IllegalArgumentException: protocol = http host = n…

colinstonema:
java.lang.IllegalArgumentException: protocol = http host = null' will throw when source path is a http/https url


joelittlejohn:
It looks like this fix will silently ignore these paths. Is that right? That doesn't seem like a good solution.

ctrimble:
I took a quick look at this.  Here are my thoughts:

What is needed is a guard around the call to `FilenameUtils.normalize`, so that only elements of `sourcePaths[i]` that do not have a protocol will be normalized.  Something like:

```
if( URLUtil.parseProtocol(sourcePaths[i]) == URLProtocol. NO_PROTOCOL ) {
  sourcePaths[i] = FilenameUtils.normalize(sourcePaths[i]);
}
```

This will only normalize local file system paths, but it seems like the intended behavior.  `sourcePaths` is treated as an array of URLs/file system paths in the next block of code.  It may be even better to normalize it right in the [URLUtil.parseUrl](https://github.com/joelittlejohn/jsonschema2pojo/blob/master/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/util/URLUtil.java#L38) method.

PR #1005, opened 2019-07-12
Features/copy constructor

dewthefifth:
Added finer control to which constructors get generated, and added a copy constructor that does a shallow copy of the object.

These changes are fully backwards compatible.


dewthefifth:
@joelittlejohn does this need anything else to be taken into consideration?

dewthefifth:
All pending conflicts have been resolved, and master has been resynced with the branch.

joelittlejohn:
Thanks Duane. I've had a look through and this looks good to me. One change though, can you remove the word 'constructors' from the beginning of the three new config properties? They should be called e.g. `includeAllPropertiesConstructor`.

dewthefifth:
I originally named them that way so that they'd fit together with the existing "isConstructorsRequiredPropertiesOnly" property, and so that all four of the "constructor" configurations would be grouped together when sorted alphabetically.

However, you're right that the names felt clunky, and I've updated them as you indicated.

rafaelcoelho:
@joelittlejohn can you apprise us about what is the expect date to roll-out it? I''m looking forward to this feature.

PR #1008, opened 2019-07-19
support inner classes generating

zhouzhipeng:
Useful for generating all subtypes as inner classes in one main class. The default is one class corresponding to one subtype.

### DIFF
--- a/jsonschema2pojo-cli/src/main/java/org/jsonschema2pojo/cli/Arguments.java
+++ b/jsonschema2pojo-cli/src/main/java/org/jsonschema2pojo/cli/Arguments.java
@@ -69,6 +69,10 @@ public class Arguments implements GenerationConfig {
     @Parameter(names = { "--use-inner-class-builders" }, description = "Generate an inner class with builder-style methods")
     private boolean useInnerClassBuilders = false;
 
+    @Parameter(names = { "--gen-inner-classes" }, description = "Generate all subtypes as inner classes in one main class. The default is one class corresponding to one subtype.")
+    private boolean genInnerClasses = false;
+
+
     @Parameter(names = { "-c", "--generate-constructors" }, description = "Generate constructors")
     private boolean generateConstructors = false;
 
@@ -559,6 +563,11 @@ public class Arguments implements GenerationConfig {
     @Override
     public Language getTargetLanguage() {
         return targetLanguage;
+    }
+
+    @Override
+    public boolean isGenInnerClasses() {
+        return genInnerClasses;
     }
 
     @Override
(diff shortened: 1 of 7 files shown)


joelittlejohn:
Thanks @zhouzhipeng. A few comments on this:

1. To be consistent with the other configuration properties of this plugin, could you call your new option `useInnerClasses` instead of `genInnerClasses`?
2. You don't yet have any integration tests for this new feature. Every new feature needs integration tests so that we can be sure the feature is not lost in the future. Can you turn `GenInnerClassExample.java` into some integration tests? You should think of some different scenarios in which this will be used and write tests for these. 
3. This part:

        newType = _package.classes().next()._class(...);

    Could you explain how it works? `_package.classes().next()` will just get the first class in the package. How does this work for the very first class we generate? I wonder if this will always be a good class to use :thinking:

bradleydeutsch:
Hey guys, is there any update on this feature please?  It's something I need :)

Thanks

dougsatch:
I need this feature too - is there anything I can do to help get it merged in?

vicnevicne:
@zhouzhipeng Thanks for the great work already, but could you please take a look at the review comments ? 
They seem to be rather minor and this feature would be a real plus... 
Would be great to integrate the change before the PR's anniversary :-)

joelittlejohn:
If anyone is keen to use this, feel free to pick up the review comments yourself and created a new branch based on the work that's already been done by @zhouzhipeng.

shubhamoscar:
What can we do to get this changes in? Im dependent on this change as well.

alersrt:
Any updates with it?

bertramn:
Question, how is this useful? What would be the use case that requires to nest all types inside a main type?

ls-daniil:
> Question, how is this useful? What would be the use case that requires to nest all types inside a main type?

Pretty useful in our case: 

We keep our schema versioned with backwards compatibility to old schema versions a requirement (the data is financial and is stored in a retention-locked bucket for 7 years). I.e.: to make an incompatible change, we have to duplicate the whole schema. 

This makes a lot of manual work to keep the nested class names from colliding. We do this manually, by appending "M<major>m<minor>p<patch>" suffixes to all internal objects, as in

```json
"nestedField": {
   "$ref": "#/definitions/NestedStructureM1m2p3"
}
```

This versioning has to be mirrored in the java code that uses the new schema versions.  As you might imagine, this is a pain.

ctrimble:
I reviewed this PR today for possible inclusion in a 1.3.0 release.  It does not appear to have proper test cases documenting the functionality.

PR #1014, opened 2019-08-09
Feature/no more casting for builders

dewthefifth:
The previous implementation required every call to build() on an inherited builder to cast the results or else store them in the base class. This change will add an overridden build() method to the children which casts the result of the original parent's build method.


joelittlejohn:
Nice change, thanks!

PR #1017, opened 2019-08-16
Generate anonymous enum into its parent class.

### LINKED ISSUE #1012: Naming conflict, generated file is overwritten
Hi,

With the following schema, the generated `Kind` class is a enum; i expect a regular POJO matching the `Kind` definition.

If i rename the `MyItem#kinds` property, it works. (in the real life, i cannot rename the property, the schema is provided by a partner)

Here is the offending schema:
```json
{
  "$schema": "http://json-schema.org/draft-03/schema#",
  "javaType": "com.example.Response",
  "type": "object",
  "properties": {
    "myItems": {
      "type": "array",
      "items": {
        "$ref": "#/definitions/MyItem"
      }
    }
  },
  "definitions": {
    "Kind": {
      "$schema": "http://json-schema.org/draft-03/schema#",
      "javaType": "com.example.Kind",
      "type": "object",
      "properties": {
        "kindCode": {
          "type": "string"
        }
      }
    },
    "MyItem": {
      "$schema": "http://json-schema.org/draft-03/schema#",
      "javaType": "com.example.MyItem",
      "type": "object",
      "properties": {
        "kinds": {
          "type": "array",
          "items": {
            "enum": [
              "KIND1",
              "KIND2"
            ]
          }
        },
        "kindWithRequiredData": {
          "$ref": "#/definitions/KindWithRequiredData"
        }
      }
    },
    "KindWithRequiredData": {
      "$schema": "http://json-schema.org/draft-03/schema#",
      "javaType": "com.example.KindWithRequiredData",
      "type": "object",
      "properties": {
        "kinds": {
          "type": "array",
          "items": {
            "$ref": "#/definitions/Kind"
          }
        }
      }
    }
  }
}
```
Regards,

ghostd:
This avoids some potential naming conflicts with classes having the same name in the same package.

Fixes #1012


ghostd:
I added @Ignore for the test `multipleEnumArraysWithSameName`, since the enum is now an inner class. This is also more stable across the re-generation when the new schema insert a new similar enum.

PR #1026, opened 2019-09-24
Implement optional generation of the JDK ConstructorProperties …

ddcruver:
…annotation on constructors to aide in serialization wherever supported.  Jackson/Spring Framework support it for various things whereas Android omits the actual @ConstructorProperties class.


joelittlejohn:
Thanks for submitting this. You've included tests too which is great! :+1: 

One change though (hopefully can be a few easy find/replaces), I think this property should be named `includeConstructorPropertiesAnnotation` for consistency with the other properties.

ddcruver:
How does that look now?

PR #1032, opened 2019-10-21
Optionally exclude the timezone from the generated @JsonFormat annotation for date-time fields

jevring:
By default, when enabling formatting for fields with "type: date-time", the time-zone is *always* forced to UTC. This feature adds a flag where this can be disabled, which means that there will be no time-zone specified in the `@JsonFormat` annotation, which means that when we produce json from an object with jackson, it uses whatever timezone the date was in already. This is only really worthwhile when not using java.util.Date, such as for example java.time.ZonedDateTime.

NOTE: The integration tests I wrote work fine in IntelliJ IDEA, but they don't generate any annotations when run through maven. I'm not quite sure why this is, or if it is just a local issue for me. It would be nice if someone else could verify this and possibly help me resolve it, if it is indeed an issue for others.

### DIFF
--- a/jsonschema2pojo-ant/src/main/java/org/jsonschema2pojo/ant/Jsonschema2PojoTask.java
+++ b/jsonschema2pojo-ant/src/main/java/org/jsonschema2pojo/ant/Jsonschema2PojoTask.java
@@ -127,6 +127,8 @@ public class Jsonschema2PojoTask extends Task implements GenerationConfig {
 
     private String outputEncoding = "UTF-8";
 
+    private boolean excludeTimezoneFromDateTimeFormat = false;
+
     private boolean useJodaDates = false;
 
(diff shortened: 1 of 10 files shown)


jevring:
@joelittlejohn Any idea on why that test failure happens? Why the annotations aren't being generated when I run it through maven, whereas they do when I run it through IntelliJ IDEA?

AntonovDmitriy:
The problem is still valid. Please raise the priority. I have to delete the time zone with my hands

ctrimble:
@jevring it looks like this PR stalled out.  Is this something you still need?  Also, were you ever able to get the tests working?

jevring:
@ctrimble I no longer work at the company where we used this, so this isn't very relevant to me anymore.

Issue #999, opened 2019-06-26
Enumeration name fallback for JSON serialization


ddcruver:
The use case is I would like to use enumerations in JSON that follow the "human-readable" convention but also except the generate numeration name during JSON deserialization.  This is for legacy reasons and completeness.  i.e. we might be coming from an existing POJO or developers that may be to use to using the enumeration name in JSON instead of a better "human-readable" version.

In my example Digit POJO I would like the user to define a digit as "One", "Two" ... "Nine" in JSON but also support using the generated names "ONE", "TWO", "THREE" ... "NINE" as well.

Example Schema Definition:
```json
{
  "$id": "https://example.com/schemas/math/Digit.schema.json",
  "$schema": "http://json-schema.org/draft-07/schema#",
  "type": "string",
  "javaType": "com.example.api.math.Digit",
  "description": "All the digits.",
  "enum": [
    "One",
    "Two",
    "Three",
    "Four",
    "Five",
    "Six",
    "Seven",
    "Eight",
    "Nine"
  ],
  "required": [],
  "definitions": {},
  "additionalProperties": false
}
```

Example Generated POJO:
```java

package com.example.api.math;

import java.util.HashMap;
import java.util.Map;
import com.fasterxml.jackson.annotation.JsonCreator;
import com.fasterxml.jackson.annotation.JsonValue;


/**
 * All the digits.
 * 
 */
public enum Digit {

    ONE("One"),
    TWO("Two"),
    THREE("Three"),
    FOUR("Four"),
    FIVE("Five"),
    SIX("Six"),
    SEVEN("Seven"),
    EIGHT("Eight"),
    NINE("Nine");
    private final String value;
    private final static Map<String, Digit> CONSTANTS = new HashMap<String, Digit>();
    private final static Map<String, Digit> NAMES = new HashMap<String, Digit>();

    static {
        for (Digit c: values()) {
            CONSTANTS.put(c.value, c);
        }
        for (Digit c: values()) {
            NAMES.put(c.name(), c);
        }
    }

    private Digit(String value) {
        this.value = value;
    }

    @Override
    public String toString() {
        return this.value;
    }

    @JsonValue
    public String value() {
        return this.value;
    }

    @JsonCreator
    public static Digit fromValue(String value) {
        Digit constantValue = CONSTANTS.get(value);
        if (constantValue == null) {
            {
                Digit nameValue = NAMES.get(value);
                if (nameValue == null) {
[...]

PR #1040, opened 2019-11-05
Add Rule Logger

ddcruver:


ddcruver:
Opps .. tests need logger now...

jsonschema2pojo-ant/src/main/java/org/jsonschema2pojo/ant/AntRuleLogger.java
```diff
@@ -0,0 +1,81 @@
...
+  public void trace(String msg) {
+    log(msg, Project.MSG_VERBOSE);
+  }
+
+  @Override
+  public void warn(String msg) {
+    log(msg, Project.MSG_WARN);
+  }
+
+  private void log(String msg, int level) {
```
joelittlejohn: I think pass a `levelName` in here then there's no need for the switch.

ddcruver:
@joelittlejohn 
It looks like the Gradle version (1.6) it builds the code does not match the version (2.3) that is run on the build server.  Which version of Gradle should we target?  Both versions are ancient.

jsonschema2pojo-cli/pom.xml
```diff
@@ -50,8 +50,8 @@
                 <configuration>
                     <archive>
                         <manifest>
-                            <classpathPrefix>./</classpathPrefix>
                             <addClasspath>true</addClasspath>
+                            <classpathPrefix>lib/</classpathPrefix>
```
joelittlejohn: What's the reason for this change?

jsonschema2pojo-cli/pom.xml
```diff
@@ -72,6 +69,7 @@
                         </goals>
                         <configuration>
                             <includeScope>runtime</includeScope>
+                            <outputDirectory>${project.build.directory}/lib</outputDirectory>
```
joelittlejohn: What's the reason for this change?

jsonschema2pojo-cli/pom.xml
```diff
@@ -111,26 +109,44 @@
...
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-simple</artifactId>
+            <version>1.7.28</version>
+            <scope>compile</scope>
+        </dependency>
+
         <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
+            <scope>test</scope>
```
joelittlejohn: The scope is not needed for these three dependencies. It's included in `dependencyManagement` in the parent.

jsonschema2pojo-cli/src/main/scripts/jsonschema2pojo.bash
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
```
joelittlejohn: What's the reason for this script? We already have the `.sh` :thinking:

jsonschema2pojo-cli/src/main/scripts/jsonschema2pojo.bat
```diff
@@ -15,4 +15,7 @@
 @REM
 
 @echo off
-java -jar "%~dp0/../lib/${project.build.finalName}.jar" %*
+
+pushd %~dp0
```
joelittlejohn: What's the reason for this change? It looks like this will cause the script to always run in the working directory of the script itself, which doesn't seem correct.

jsonschema2pojo-cli/src/main/scripts/jsonschema2pojo.sh
```diff
@@ -18,4 +18,4 @@
 SCRIPT=$(readlink -f "$0")
 SCRIPTPATH=$(dirname "$SCRIPT")
 
-java -jar ${SCRIPTPATH}/../lib/${project.build.finalName}.jar "$@"
+java -jar "${SCRIPTPATH}"/"${project.build.finalName}".jar "$@"
```
joelittlejohn: What's the reason for this change?

jsonschema2pojo-cli/src/main/scripts/simplelogger.properties
```diff
@@ -0,0 +1,34 @@
+# SLF4J's SimpleLogger configuration file
```
joelittlejohn: This file should be coming from the classpath I think, not read as a file. It should be in the root of the classpath for this cli project e.g. `src/main/resources`.

joelittlejohn: If possible the CLI logging should be configured via command line switches.

joelittlejohn:
Strange, I have no idea why it says that this zip file was empty:

```
org.gradle.tooling.GradleConnectionException: Could not install Gradle distribution from 'http://services.gradle.org/distributions/gradle-2.3-bin.zip'.
```

When I download this using the same link I don't get an empty.

jsonschema2pojo-cli/src/main/scripts/jsonschema2pojo.bash
ddcruver: It uses pushd/popd. If you put those in the sh. IDE with inspections will warn that those commands are not POSIX compliment.

Could I just remove the sh one?

jsonschema2pojo-cli/src/main/scripts/jsonschema2pojo.bat
ddcruver: The manifest class path libs weren't resolving if I didn't do it.  If they provide a relative path as a parameter that may cause problems. Let me find a better solution.

jsonschema2pojo-cli/src/main/scripts/jsonschema2pojo.sh
ddcruver: The cli jar does not get built into the lib directory.

jsonschema2pojo-cli/src/main/scripts/simplelogger.properties
ddcruver: I will leave the command line switches for a future effort but agree.

jsonschema2pojo-cli/pom.xml
ddcruver: Thanks overlooked that. I normally don't dependency manage test artifacts but it is good.

ddcruver:
I think the stack trace above it gives more hint at what is going on but still not sure how to fix it.

Are you open to updating the groovy, Gradle, groovy maven plugin to resolve this issue? Or I could just not support logging under the Gradle plugin.

ddcruver:
The gradle plugin project failing is a weird problem because it fails only on the Travis CI server and not a local maven execution of it.

jsonschema2pojo-cli/src/main/scripts/jsonschema2pojo.bat
```diff
@@ -1,18 +1,20 @@
...
+@REM distributed under the License is distributed on an "AS IS" BASIS,
+@REM WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+@REM See the License for the specific language governing permissions and
+@REM limitations under the License.
+@REM
+
+@echo off
+
+java -jar "%~dp0\${project.build.finalName}.jar" %*
+
```
joelittlejohn: It looks like you've changed the CRLF line endings in this file to LF. Since this is a windows batch file, we need to keep the CRLF line endings. You could add this to you `.gitattributes` file, like:

```
*.bat text eol=crlf
```

jsonschema2pojo-gradle-plugin/src/test/resources/simplelogger.properties
```diff
@@ -0,0 +1,34 @@
...
+# Defaults to true.
+org.slf4j.simpleLogger.showThreadName=false
+
+# Set to true if you want the Logger instance name to be included in output messages.
+# Defaults to true.
+org.slf4j.simpleLogger.showLogName=false
+
+# Set to true if you want the last component of the name to be included in output messages.
+# Defaults to false.
+org.slf4j.simpleLogger.showShortLogName=false
```
joelittlejohn: Do we need this file?

.travis.yml
```diff
@@ -1,7 +1,20 @@
...
 cache:
   directories:
     - "$HOME/.m2"
+    - "$HOME/.gradle/caches/"
+    - "$HOME/.gradle/wrapper/"
+
+script:
+  - rm -rf ~/.gradle/wrapper/dists/gradle-2.3-bin
+  - mvn clean install -DskipTests
+  - mvn verify
```
joelittlejohn: What's the reason for `install` then `verify`?

ddcruver:
@joelittlejohn Since #1041 is also failing I believe this failure is not a result of my changes.  Something is broken with the build.

This is a far reach but @ben-manes do you have any insight into this problem?

ben-manes:
I believe it is due to not following a redirect. They enforce HTTPS now and their http client probably doesn't have redirect enabled. So maybe a 1 character fix! 😄 

```
$ curl -I http://services.gradle.org/distributions/gradle-2.3-bin.zip
HTTP/1.1 301 Moved Permanently
Date: Wed, 06 Nov 2019 18:58:00 GMT
Connection: keep-alive
Cache-Control: max-age=3600
Expires: Wed, 06 Nov 2019 19:58:00 GMT
Location: http://downloads.gradle-dn.com/distributions/gradle-2.3-bin.zip
Server: cloudflare
CF-RAY: 53194be1fed592c8-SJC

ben: lucene-solr $ wget -v http://services.gradle.org/distributions/gradle-2.3-bin.zip
--2019-11-06 10:58:10--  http://services.gradle.org/distributions/gradle-2.3-bin.zip
Resolving services.gradle.org (services.gradle.org)... 104.18.190.9, 104.18.191.9
Connecting to services.gradle.org (services.gradle.org)|104.18.190.9|:80... connected.
HTTP request sent, awaiting response... 301 Moved Permanently
Location: http://downloads.gradle-dn.com/distributions/gradle-2.3-bin.zip [following]
--2019-11-06 10:58:10--  http://downloads.gradle-dn.com/distributions/gradle-2.3-bin.zip
Resolving downloads.gradle-dn.com (downloads.gradle-dn.com)... 104.17.160.20, 104.17.159.20
Connecting to downloads.gradle-dn.com (downloads.gradle-dn.com)|104.17.160.20|:80... connected.
HTTP request sent, awaiting response... 301 Moved Permanently
Location: https://downloads.gradle-dn.com/distributions/gradle-2.3-bin.zip [following]
--2019-11-06 10:58:10--  https://downloads.gradle-dn.com/distributions/gradle-2.3-bin.zip
Connecting to downloads.gradle-dn.com (downloads.gradle-dn.com)|104.17.160.20|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 42475057 (41M) [application/zip]
Saving to: ‘gradle-2.3-bin.zip’

gradle-2.3-bin.zip                       100%[==================================================================================>]  40.51M  20.2MB/s    in 2.0s

2019-11-06 10:58:12 (20.2 MB/s) - ‘gradle-2.3-bin.zip’ saved [42475057/42475057]
```

ddcruver:
I literally asked you then though to do that myself. Lets see if it works.

jsonschema2pojo-cli/pom.xml
ddcruver: This was so that the jar can be run in the target directory and it can locate the jar files in the lib folder.

jsonschema2pojo-cli/pom.xml
ddcruver: Just condensing the two configuration sections together.  There is only one execution for this plugin so just moved all the configuration to a single configuration block.

jsonschema2pojo-cli/pom.xml
ddcruver: Fixed.

jsonschema2pojo-cli/src/main/scripts/jsonschema2pojo.bash
ddcruver: I have removed the bash version.

jsonschema2pojo-cli/src/main/scripts/jsonschema2pojo.bat
ddcruver: Remove all pushd and popd commands.

jsonschema2pojo-cli/src/main/scripts/jsonschema2pojo.sh
ddcruver: Even after I removed the pushd and popd and executed the script by ./target/jsonschema2pojo.sh it still works.

jsonschema2pojo-cli/src/main/scripts/jsonschema2pojo.bat
ddcruver: I believe I only did this for the .sh file and added a .gitattributes entry for it.

.travis.yml
ddcruver: I was trying to resolve the build issues and found this commit (81adfe76acdca7b5e965350d02b3bd91a1675461) and it mentioned that CI tests might be broken unless do a build then do a test build.

@ben-manes Is this unnecessary now?

ddcruver:
@ben-manes I also switch back to the groovy-eclipse-compiler, these days which groovy plugin should you use? 

I have read lots of write-ups on "which maven plugin/compiler for groovy should I use" but not being a groovy developer I am still not sure what this project should be using. I changed it because I thought it might fix my build/execution issues but it seems like it might of been unrelated now.

.travis.yml
ddcruver: This might be necessary with the eclipse-groovy-compiler. I will try removing it.

ben-manes:
I can't answer at that detail. I simply hacked together a few plugins for my needs and contributed them upstream (flyway, js2p, jooq, dependencyUpdates). I learned enough Groovy to write simple build scripts and plugins, but this was Gradle 1.x world (2012). I don't use Groovy otherwise and tried to handoff the plugins since this isn't my area of expertise. I'm happy to be a sounding board but my knowledge is fairly limited and outdated.

.travis.yml
ben-manes: That commit comment wrt CI failure was a one-time breakage and no longer relevant. Just required a clean if I recall correctly.

ddcruver: @joelittlejohn I have already removed the extra build and haven't notice any problems.
@ben-manes Thanks for your input.

ddcruver:
I read that the gmaven(plus) plugins had more features and may support more things but I believe thease are things we don't require, so since the eclipse plugin is an official plugin should leave it switched to that.  

I believe there is other improvements to the jsonschema2pojo-gradle-plugin and maybe it should actually be build with gradle (embedded in our maven) instead but that is out of scope with something I am willing to do right now.

.travis.yml
joelittlejohn: So I think all the changes in this file can be reverted, is that right?

ddcruver:
@joelittlejohn I believe I have responded and/or addressed to all your comments and the build looks good now.

jsonschema2pojo-cli/src/main/scripts/jsonschema2pojo.bat
joelittlejohn: You've actually swapped the CRLF line endings in this .bat file to LF as well (see the diff, all lines are changed).

Also, if you download a release from https://github.com/joelittlejohn/jsonschema2pojo/releases and look inside the archive, you'll see why this `../lib`` is required in these scripts. The scripts and the jars do not live alongside each other. We have brew integration that also depends on this layout.

jsonschema2pojo-cli/pom.xml
joelittlejohn: This will break our releases so I think we need to remove it.

jsonschema2pojo-cli/src/main/scripts/jsonschema2pojo.sh
joelittlejohn: This change will break our releases, we need to revert it.

joelittlejohn:
Great, thanks for looking at the Gradle build issues. I think we still have a few changes we need to make here, in particular any change that removes `../lib` from CLI scripts needs to be reverted. I've added some follow-up comments and unresolved a few conversations.

Also, is there any reason the CLI logger needs to use SLF4J rather than just `System.out` and `System.err`? This will never be part of a wider system with its own logging configuration.

ddcruver:
> Also, is there any reason the CLI logger needs to use SLF4J rather than just `System.out` and `System.err`? This will never be part of a wider system with its own logging configuration.

I guess if this was jsonschema2-pojo-embedded (a fictional module) that would be integrated in another application/system then it should use SLF4J but I agree.  Although there would be no way to turn off loggers for CLI if people started using info/debug/trace and it could get quit noisy.

jsonschema2pojo-ant/src/main/java/org/jsonschema2pojo/ant/AntRuleLogger.java
ddcruver: I am going to implement it as a levelPrefix like "DEBUG: " so no concatenation needs to be done, it may be optimized out but who knows.

jsonschema2pojo-cli/src/main/scripts/jsonschema2pojo.bat
ddcruver: Thank you for directing me to the release download. I was fixing it for how the maven build would place it before you make your release but now it make much more sense.

It won't run as is in the build but that is not as important as the release.  I will revert these changes.

jsonschema2pojo-ant/src/main/java/org/jsonschema2pojo/ant/AntRuleLogger.java
ddcruver: Done.

jsonschema2pojo-cli/src/main/scripts/jsonschema2pojo.bat
```diff
@@ -1,18 +1,18 @@
...
+@REM
+@REM Unless required by applicable law or agreed to in writing, software
+@REM distributed under the License is distributed on an "AS IS" BASIS,
+@REM WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+@REM See the License for the specific language governing permissions and
+@REM limitations under the License.
+@REM
+
+@echo off
+java -jar "%~dp0/../lib/${project.build.finalName}.jar" %*
```
joelittlejohn: Your branch is still showing a change in this file from CRLF line endings to LF. Could you maybe `git checkout origin/master jsonschema2pojo-cli/src/main/scripts/jsonschema2pojo.bat` and commit that?

jsonschema2pojo-cli/pom.xml
ddcruver: Fixed

jsonschema2pojo-cli/src/main/scripts/jsonschema2pojo.sh
ddcruver: Fixed.

jsonschema2pojo-gradle-plugin/src/test/resources/simplelogger.properties
ddcruver: Nope, we are using defaults anyways.

.travis.yml
ddcruver: I have removed the hacky gradle bin and double building stuff from Travis CI configuration but left in the caching.

jsonschema2pojo-cli/src/main/scripts/jsonschema2pojo.bat
ddcruver: I have added .bat to the .gitattributes file and renormlized that bat file. It may still show changes because it was original in the git repo as LF.

https://stackoverflow.com/a/47580886/45708

ddcruver:
@joelittlejohn I am ready for you to take another pass of seeing if this pull request is ready to merge.

jsonschema2pojo-cli/src/main/scripts/jsonschema2pojo.bat
joelittlejohn: Hi Dan. Everything looks good but this is the last problem in the PR. In `master` this file uses CRLF (Windows) line endings not LF:

```
$ file jsonschema2pojo-cli/src/main/scripts/jsonschema2pojo.bat
jsonschema2pojo-cli/src/main/scripts/jsonschema2pojo.bat: DOS batch file, UTF-8 Unicode text, with CRLF line terminators
```

You can also try `vim -b jsonschema2pojo-cli/src/main/scripts/jsonschema2pojo.bat` to see the `^M` characters.

In your branch (this PR) changes the line endings in this file to LF.

I've actually changed my mind about this `.gitattributes` file, it's adding more confusion and I don't want git doing anything to change the files under version control. Please remove `.gitattributes` and also remove your changes to this file (`git checkout origin/master jsonschema2pojo-cli/src/main/scripts/jsonschema2pojo.bat`). Then we're good to go.

ddcruver: Done.

ddcruver:
@joelittlejohn I would like to wrap up this pull request and possibly #1041.  I would like to write a RuleFactory with a custom EnumRule for my project that has the name fallback resolution as described in #999 and I want it to extend EnumRule instead of re-implementing it.

PR #1041, opened 2019-11-06
Implement javaEnums extension 

ddcruver:
Implement javaEnums extension to extend on what javaEnumNames provided.


ddcruver:
@joelittlejohn I have additionally changed some of the methods to protected to allow a custom EnumRule to be able to extend the rule in multiple different discrete ways instead of copy-n-pasting or reimplementing it all.

For example in a custom EnumRule I would like to write for my project I would like to even override the Constructor that is generated and add additional fields to the enum.  The actual use case is obscure so I am implementing it outside of this project.

ddcruver:
@joelittlejohn can we get this pull request in?

joelittlejohn:
If you could add something to https://github.com/joelittlejohn/jsonschema2pojo/wiki/Reference#extensions that would be great :+1:

ddcruver:
I will put it on my todo list.

PR #1054, opened 2019-12-05
Add package when javaType is not a fully qualified name

eltabo:
I am working in a POC with OCPP 1.6/2.0 that use json schemas with a not fully qualified javatype (example attacched). May be useful check if a specified javatype is fqn or not and hence add the package name if needed.

### DIFF
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/ObjectRule.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/ObjectRule.java
@@ -219,6 +219,11 @@ public class ObjectRule implements Rule<JPackage, JType> {
                 }
 
                 int index = fqn.lastIndexOf(".") + 1;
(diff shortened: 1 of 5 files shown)


joelittlejohn:
Hi @eltabo. Thanks for submitting this. I think this is a good change, but we need an integration test that checks this behaviour before we can merge this PR. Could you add one?

eltabo:
I have added some integration tests.

brydesigns:
So in 1.1.1, `"javaType":"Foo"` produces `com.example.Foo`, `"javaType":"subpackage.Foo"` and `"javaType":".subpackage.Foo"` both produce `subpackage.Foo`, and I still have no idea how to produce `com.example.subpackage.Foo`.... I'm using `<sourceSortOrder>FILES_FIRST</sourceSortOrder>` because I have subdirectory schema that is incorrectly packages a `"$ref":"../bar.json"` reference, so that might be affecting my results (OS or SUBDIRS_FIRST do at least create subpackage).

PR #1057, opened 2019-12-10
Add option to use java.util.Optional for non required reference fields

legoethals:
I have added an option to use Optional for non required fields, similar to the option that exists for returning Optional on the getters of non required fields.
Although it's not recommended to use Optional for fields, it's supported by default by Jackson >=2.8.5 and it's the easiest way to differentiate between a field not being present in the json document, and the field being present but with a value of null.


sharkvan:
I would like to have this feature too.

aslakhellesoy:
@legoethals looks like this is already implemented: https://github.com/joelittlejohn/jsonschema2pojo/pull/913

PR #1063, opened 2019-12-31
Restrict JSR-303 annotations to supported field types

rajpaulbagga:
I found that JSR-303 annotations were getting applied to types that didn't support them. For example, if a "pattern" was specified on a "string" property that also had a "format" of "uuid", then the generated field type is `UUID` class (rather than `String`) and `@Pattern` was still applied. Because that annotation isn't valid on a `UUID` type, then validators generate errors.

Continuing with this example, because a schema may be used to generate source classes in non-Java languages, there is still value to keep the "pattern" restriction on the source json schema itself for languages that might not have an in-built UUID type. 

This patch updates the various JSR-303 Rules classes to enhance them to check the generated class type of potentially annotated fields against the classes/types that are documented to support the relevant annotation (as documented in each annotation's JavaDoc).

In a second logical commit, Unit tests were added for thes JSR-303 rules classes. This required updating Mockito to a newer version that (properly) supports mocking final classes.

In a third logical commit, integration tests were updated with schemas that generate unsupported field types for the various cases, in addition to the supported types.

### DIFF
--- a/jsonschema2pojo-cli/pom.xml
+++ b/jsonschema2pojo-cli/pom.xml
@@ -130,7 +130,7 @@
         </dependency>
         <dependency>
             <groupId>org.mockito</groupId>
-            <artifactId>mockito-all</artifactId>
+            <artifactId>mockito-core</artifactId>
         </dependency>
     </dependencies>
 </project>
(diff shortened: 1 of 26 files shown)


rajpaulbagga:
@joelittlejohn ,  
Do you have any plans or notions when you might release 1.0.2?
Thank you.

Issue #1075, opened 2020-01-21
Schemas with relative paths can generate the same class twice with "__1" in the name


sharkvan:
When I generate schemas from places that will have a relative path for one instance of the referenced schema I get two classes that are the same except for their names.

For instance if I run the following command line I will get two class, Person.java, and Person__1.java. What I'm expecting is just one class of Person.java that references itself in the children property.

`./jsonschema2pojo -s ../../jsonschema2pojo-core/src/test/resources/schema/person.json -t .`

I will create a unit test in a new branch to show this behavior.

Json schema for test.
`{
  "$schema": "http://json-schema.org/draft-07/schema",
  "$id": "https://example.com/person.json",
  "type": "object",
  "title": "Person",
  "properties": {
    "firstName": {
      "type": "string",
      "minLength": 1
    },
    "lastName": {
      "type": "string",
      "minLength": 1
    },
    "middleInitial": {
      "type": "string"
    },
    "suffix": {
      "type": "string"
    },
    "children": {
      "$ref":"person.json"
    }
  }
}`

joelittlejohn:
Closed by #1077

PR #1077, opened 2020-01-22
Recursive schemas with relative paths can generate the same class twice with "__1" in the name

sharkvan:
Added a unit test to show how URIs with relative path segments will cause a recursive schema model to generate two classes with a difference in name only. Also added a fix for the code as well

 #1075


joelittlejohn:
Great, thanks!

PR #1078, opened 2020-01-22
Missing test case

sharkvan:
I apologize, but I missed a test case that exposed that fact that the fix was not complete. I wanted the SchemaStore to return the same object instance and it was not. I updated it to do so now and added the test to validate it.

#1077


sharkvan:
@joelittlejohn, I just wanted to see if you saw this?

Issue #1081, opened 2020-01-24
Feature to not create duplicate classes when using useTitleAsClassname


sharkvan:
I often get schemas that look like the one below where parts of the schema use duplicate subschemas. When using the config option to generate classes using the `title` properties as the classname, this causes me to get classes that end in "__1", however they are all basically the same thing.

With the useTitleAsClassname the following classes would be generated by the below schema
Car.java
Wheel.java
Wheel__1.java
Wheel__2.java
Wheel__3.java

What I would like to get (or would expect to get) is the following classes generated by the below schema
Car.java
Wheel.java

`{
  "$schema": "http://json-schema.org/draft-07/schema",
  "$id": "https://example.com/car.json",
  "type": "object",
  "title": "Car",
  "properties": {
    "frontRightWheel": {
      "$id": "#wheel.json",
      "title": "Wheel",
      "type": "object",
      "properties": {
        "pressure": {
          "$id": "#pressure",
          "title": "Pressure",
          "type": "integer"
        },
        "size": {
          "$id": "#size",
          "title": "Size",
          "type": "integer"
        }
      }
    },
    "frontLeftWheel": {
      "$id": "#wheel.json",
      "title": "Wheel",
      "type": "object",
      "properties": {
        "pressure": {
          "$id": "#pressure",
          "title": "Pressure",
          "type": "integer"
        },
        "size": {
          "$id": "#size",
          "title": "Size",
          "type": "integer"
        }
      }
    },
    "backRightWheel": {
      "$id": "#wheel.json",
      "title": "Wheel",
      "type": "object",
      "properties": {
        "pressure": {
          "$id": "#pressure",
          "title": "Pressure",
          "type": "integer"
        },
        "size": {
          "$id": "#size",
          "title": "Size",
          "type": "integer"
        }
      }
    },
    "backLeftWheel": {
      "$id": "#wheel.json",
      "title": "Wheel",
      "type": "object",
      "properties": {
        "pressure": {
          "$id": "#pressure",
          "title": "Pressure",
          "type": "integer"
        },
        "size": {
          "$id": "#size",
          "title": "Size",
          "type": "integer"
        }
      }
    }
  }
}`

PR #1082, opened 2020-01-24
Same class different name

sharkvan:
This is a fix for the linked feature request. Please give any feedback on the approach.

The basic concept is when using the config option for useTitleAsClassname, I will cache type objects by a sha key of their schema, then look up potential reuse by that sha key.

#1081


jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/model/EnumValueDefinition.java
```diff
@@ -1,3 +1,19 @@
+/**
```
sharkvan: I added this changes because everytime the build ran they came back, so I thought it would be nice for everyone to not have to exclude these again.

sharkvan:
Hi @joelittlejohn, not trying to be pushy, but I was wondering if you had a chance to review this PR?

joelittlejohn:
Hi Tim. I'm afraid I actually don't see us merging this one into the core of jsonschema2pojo, but thank you for sharing the code with others. We can't support every different idea that people have about the way they would like this plugin to act, because the maintenance burden (and multiplied complexity of all these different options) becomes huge.

If you want these various things to be understood as all referencing the same type, then you can use a $ref. This plugin understands that all refs to the same schema should reuse the same type.

sharkvan:
I understand. The reason I did this is because I cannot always control the schema I'm getting but they are essentially the same schema. There are tools out there the remove the `$ref` and replace it with the schema it references, removing any ability for tools like this one to do anything about it. But thank you for your consideration.

sharkvan:
@joelittlejohn, what would you think about merging the changes to the `Schema` class only so custom implementations of the the RuleFactory and the implementations it provides can have a hook into when the `JType` is set to the `Schema` class?

Issue #1100, opened 2020-02-24
Legal characters are limited to [0-9a-zA-Z_]


asympro:
Greetings @joelittlejohn,

First of all I would like to thank you for your excellent library. It is really simple to use.

Some time ago I started a project a small part of which is integration with governments' facilities.
The issue is that it is not English speaking which leads to the fact that JSON-schema contains non-latin characters (e.g. Cyrillic letters).

I cloned the repo and simply removed the part with removing illegal characters and it worked like a marvel (tests green).

Therefore I would like to ask 2 questions:

1)Why removing other characters in first place? Do I miss something or the getСтранныйНелатинскийЯзык looks funny? Actually it is but I cannot solve the problem by reworking the schema of government facility.

2)I would like to fix all this and contribute to the source code. As the owner of the repo I would like to ask if the fix should be optional (maven config disabled by default) or just remove the illegal characters' omitting?

Once again the library rocks, thank you.

Best regards, Asympro

ddcruver:
I'm not the owner but in general if you add tests he should let the code change in. There is a class NameHelper which is used throughout the code. I recommend extending that. You may have to update code to also handle unicode. Like of hand you may need to find unicode equivalent toUpper methods if they don't already support it.

ddcruver:
Also it needs to work on Android. So double check any APIs you use exists in the Android JavaDocs. They have surprising gaps sometimes.

joelittlejohn:
I think since this could completely transform the generated output I think we'd have to introduce a configuration option for this. I'd be happy to merge a PR.

Maybe `useAsciiJavaNames` where the default is `true` but it can be overridden to `false` (note: in the CLI you'll have to switch the logic and use something like `--allow-extended-java-characters`).

asympro:
> I think since this could completely transform the generated output I think we'd have to introduce a configuration option for this. I'd be happy to merge a PR.
> 
> Maybe `useAsciiJavaNames` where the default is `true` but it can be overridden to `false` (note: in the CLI you'll have to switch the logic and use something like `--allow-extended-java-characters`).

Thanks for your reply.

I have done as you suggested.
Changes are available in this PR:

https://github.com/joelittlejohn/jsonschema2pojo/pull/1105

Please comment if you see anything out of order.

Another thanks for your guidance!

Best regards, Dmitry.

PR #1105, opened 2020-03-06
#1100 Legal characters are now not limited to [0-9a-zA-Z_]

asympro:


jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/util/NameHelper.java
```diff
@@ -47,7 +47,11 @@ public String getBuilderTypeParameterName(JDefinedClass instanceClass) {
     }
 
     public String replaceIllegalCharacters(String name) {
-        return name.replaceAll(ILLEGAL_CHARACTER_REGEX, "_");
+        if (generationConfig.isAllowNonLatinNames()) {
+            return name;
```
eirnym: I bet we should add some restrictions here as `*` or whitespace is not allowed as well.

asympro: Nice one! Thanks for the comment.

Current approach:

Omit all starting characters until a valid java identifier start appears. 
Add all remaining valid java identifiers part characters.

eirnym: This [Stack Overflow](https://stackoverflow.com/a/45275027) answer has a good regex for the Java identifier characters. From this regex we can create a better way to replace invalid characters.

asympro: Unfortunately this regex is of no use here as none of them actually match this as an example:
"урл"
which is why this PR was created in the first place.
Also performance-wise approach in PR should be faster on the first glance and can be measured using JMH if this need any proof.

eirnym: \w is not only A-Z, but урл, łą and many other `word` symbols accepted by Java.
I'd separated first character and others to make sure everything is fine, 

Based on `java.util.Pattern` javadoc, you need following invalid characters RE for the first char: `[^$\p{IsAlphabetic}_]` and for other characters`[^\p{IsAlphabetic}\p{Digit}$_]*`

I don't know how to address other digits outside range 0-9, but I hope those restrictions would be enough to address the most. Obviously, you can use just any characters (except few may be) for literals in JVM itself, but I see this as a practical subset for names.

eirnym: Also we can use `Character.isJavaIdentifierStart()` and `Character.isJavaIdentifier` and replace letters, but this will cover only few more characters, like below. I don't know if you wish to enable them as well, but it's definitely possible. Also these schemas could be not  compatible between different generators and languages.

```
¢ £ ¤ ¥ ؋ ৲ ৳ ৻ ૱ ௹ ฿ ៛ ‿ ⁀ ⁔ ₠ ₡ ₢ ₣ ₤ ₥ ₦ ₧ ₨ ₩ ₪ ₫ € ₭ ₮ ₯ ₰ ₱ ₲ ₳ ₴ ₵ ₶ ₷ ₸ ₹ ꠸ ﷼ ︳ ︴ ﹍ ﹎ ﹏ ﹩ ＄ ＿ ￠ ￡ ￥ ￦
```

asympro: Both isJavaIdentifierStart and isJavaIdentifierPart are used in the PR suggested.
In result a valid java identifier is returned.
This is the fastest algorithm I could think of without much limitations.
As for particality - I highlighted the problem in the issue.
I cannot really do anything with the JSON schema provided by the government facilities for API integration.
I wish to enable any valid java identifier from the source JSON in case nothing can be done to convert to latin+numeric, so that others can gracefully avoid this problem with simple option.

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/util/NameHelper.java
```diff
@@ -47,7 +47,36 @@ public String getBuilderTypeParameterName(JDefinedClass instanceClass) {
...
+            return removeNotValidJavaIdentPartsStartingFrom(name, getValidJavaIdentifierStartIndex(name));
+        } else {
+            return name.replaceAll(ILLEGAL_CHARACTER_REGEX, "_");
+        }
+    }
+
+    private String removeNotValidJavaIdentPartsStartingFrom(String name, int validStartPosition) {
+        StringBuilder sb = new StringBuilder();
+        int length = name.length();
+        sb.append(name.charAt(validStartPosition));
```
eirnym: you all invalid symbols of the `name` while original algorithm replace all invalid characters with underscore. I'd like to preserve behaviour as it was, to make this one as an extended version of the previous

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/util/NameHelper.java
eirnym: if you use `IsAlphabetic` instead of `alpha`, you'll get all alphabetic symbols as well, not only from `[a-zA-Z]`

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/util/NameHelper.java
asympro: This behavior is maintained as an option, underscores are then removed.
I don't see much profit in maintaining this behavior as such.

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/util/NameHelper.java
asympro: I don't think that regexes are a good idea due to performance and/or compatibility and/or readability.
If anything changes in forward releases (though I really! doubt to happen) the behavior of open Char API will be updated making changes intact.

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/util/NameHelper.java
eirnym: Documentation for an option doesn't state, that behaviour is changed. Reading current documentation, I expect that more symbols would be accepted as valid, not only `[a-zA-Z0-9_$]`. Because of this I'd insist on maintaining the behaviour

asympro: Done, please check

asympro:
@joelittlejohn please take a look.

joelittlejohn:
Hi @asympro. Thanks for working on this. As per [CONTRIBUTING.md](https://github.com/joelittlejohn/jsonschema2pojo/blob/master/CONTRIBUTING.md#3-add-integration-tests) we do need an integration test for this.

joelittlejohn:
I don't actually understand why you've introduced a new approach to handling unsupported characters. Why is it that sometimes chars that can't be supported in the identifier are replaced with `_` and other times chars that can't be supported are in the identifier are omitted? 

Chars that are illegal (can't be supported) should always be replaced by underscores. When `allowNonLatinNames = false`, this will be any character outside [a-zA-Z0-9], when `allowNonLatinNames = true`, this will be any character where `isJavaIdentifierPart` returns false.

If `isJavaIdentifierStart` returns false for the first char, stick an underscore on the front of the name (prepend).

I'm finding the current logic hard to follow and complicated, I think it can be a lot simpler, and more consistent with how jsonschema2pojo currently works.

asympro:
You are totally right. Will implement anew like you suggested.
As for the coding style - I tried to optimize it (losing in readability though), possibly it is due to my "build is always taking too long" mental issue.

joelittlejohn:
Thanks @asympro! Let me know when you want me to take another look at your changes and I'll review 👍

asympro:
@joelittlejohn IT test written, reimplementation ready, please review

eirnym:
It's much better. I vote for this implementation.

asympro:
@joelittlejohn eirnym and I both prefer new implementation. Please review if you have time.

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/GenerationConfig.java
```diff
@@ -625,4 +625,23 @@ default boolean isUseInnerClassBuilders() {
...
+   *
+   * Defaults to 'false' meaning that if source JSON schema does contain any char other than [0-9A-Za-z] then those
+   *  characters will be omitted in target code.
+   *
+   * E.g. if source JSON schema contains a field like 'testПоле' then target code will contain field with name 'test'
+   *  and methods 'getTest'/'setTest' meaning 'Поле' part of source field will be omitted.
+   *
+   * Warning: if JSON schema contains a field which consists only of chars not in [0-9A-Za-z] and this setting is set
+   *  to 'false' then an exception can occur.
+   *
```
joelittlejohn, line 640: I think this part should be removed.

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/util/NameHelper.java
```diff
@@ -47,7 +47,28 @@ public String getBuilderTypeParameterName(JDefinedClass instanceClass) {
...
+        for (int i = 0; i < length; i++) {
+            char curChar = name.charAt(i);
+            if (Character.isJavaIdentifierPart(curChar)) {
+                sb.append(curChar);
+            } else {
+                sb.append("_");
+            }
+        }
+        if (!Character.isJavaIdentifierStart(sb.charAt(0))) {
+            return "_" + sb.toString();
```
joelittlejohn, line 69: Let's avoid two return statements:

```suggestion
            sb.insert(0, "_");
```

jsonschema2pojo-integration-tests/src/test/java/org/jsonschema2pojo/integration/config/AllowNonLatinIT.java
```diff
@@ -0,0 +1,49 @@
...
+import static org.jsonschema2pojo.integration.util.CodeGenerationHelper.config;
+import static org.junit.Assert.assertThat;
+
+public class AllowNonLatinIT {
+
+    @Rule
+    public Jsonschema2PojoRule schemaRule = new Jsonschema2PojoRule();
+
+    @Test
+    public void writeNonLatinSchema() throws IOException {
```
joelittlejohn, line 38: Could you add another test where you check that the non-latin name is escaped with `_` by default?

PR #1118, opened 2020-04-07
added private modifier to configuration options where missing

MCSafety:
The fields `includeRequiredPropertiesConstructor`, `includeAllPropertiesConstructor`, and `includeCopyConstructor` in the maven plugin cannot be configured in the pom. They are ignored by maven because they are missing the private modifier. This pull request adds this modifier.


joelittlejohn:
Thanks all! :+1:

PR #1122, opened 2020-04-08
Add configuration map (customRuleFactoryConfiguration) for Custom Rule Factories

ddcruver:
This will allow custom rule factories to have a simple string to string map of configuration properties.  For example in maven this allows a property such as this to be defined:

```
<plugin>
	<version>1.0.3-SNAPSHOT</version>
	<executions>
		<execution>
			<goals>
				<goal>generate</goal>
			</goals>
		</execution>
	</executions>
	<dependencies>
		<dependency>
			<groupId>org.example.jsonschema2pojo-extensions</groupId>
			<artifactId>org.example.jsonschema2pojo-extensions.custom-rule-factory</artifactId>
			<version>${project.version}</version>
		</dependency>
	</dependencies>

	<configuration>
		<customRuleFactory>org.example.extensions.rules.CustomRuleFactory</customRuleFactory>

		<customRuleFactoryConfiguration>
			<quantitiesAsString>true</quantitiesAsString>
		</customRuleFactoryConfiguration>
	</configuration>
</plugin>
```

This allows the custom rule factory to have configuration in the configuration section of the plugin without fear of it being an invalid configuration.

### DIFF
--- a/jsonschema2pojo-ant/src/main/java/org/jsonschema2pojo/ant/Jsonschema2PojoTask.java
+++ b/jsonschema2pojo-ant/src/main/java/org/jsonschema2pojo/ant/Jsonschema2PojoTask.java
@@ -114,6 +114,8 @@ public class Jsonschema2PojoTask extends Task implements GenerationConfig {
 
     private Class<? extends RuleFactory> customRuleFactory = RuleFactory.class;
 
+    private Map<String, String> customRuleFactoryConfiguration = new HashMap<>();
+
     private boolean includeJsr303Annotations = false;
 
     private boolean includeJsr305Annotations = false;
@@ -577,6 +579,10 @@ public class Jsonschema2PojoTask extends Task implements GenerationConfig {
         } else {
             this.customRuleFactory = RuleFactory.class;
         }
+    }
+
+    public void setCustomRuleFactoryConfiguration(Map<String, String> formatTypeMapping) {
+        this.formatTypeMapping = formatTypeMapping;
     }
 
     /**
@@ -1070,6 +1076,11 @@ public class Jsonschema2PojoTask extends Task implements GenerationConfig {
     @Override
     public Class<? extends RuleFactory> getCustomRuleFactory() {
         return customRuleFactory;
+    }
+
+    @Override
+    public Map<String, String> getCustomRuleFactoryConfiguration() {
+        return customRuleFactoryConfiguration;
     }
 
     @Override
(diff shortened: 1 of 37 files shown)


ddcruver:
For a little context without something like this I would have to do this to get a configuration value from a Maven POM:

```
protected Boolean getConfigurationValueAsBooleanParseMavenConfiguration(String configurationField, boolean defaultValue)
	{
		GenerationConfig generationConfig = this.getGenerationConfig();

		if (generationConfig instanceof AbstractMojo)
		{
			AbstractMojo abstractMojo = (AbstractMojo) generationConfig;

			Object pluginDescriptor = abstractMojo.getPluginContext().get("pluginDescriptor");

			if (pluginDescriptor instanceof PluginDescriptor)
			{
				Object configuration = ((PluginDescriptor) pluginDescriptor).getPlugin().getConfiguration();

				if (configuration instanceof Xpp3Dom)
				{
					Xpp3Dom xpp3Dom = (Xpp3Dom) configuration;
					Xpp3Dom configNode = xpp3Dom.getChild(configurationField);

					if (configNode != null && StringUtils.isNotBlank(configNode.getValue()))
					{
						getLogger().debug("Config Value for " + configurationField + ": " + configNode.getValue());
						return configNode.getValue().equalsIgnoreCase("true");
					}
				} else
				{
					getLogger().error(CustomRuleFactory.class.getCanonicalName() +
											  " found unexpected configuration object of type " + configuration.getClass().getCanonicalName() +
											  " expected " + Xpp3Dom.class.getCanonicalName() +
											  "; will not be able to resolve " + configurationField);
				}
			} else
			{
				getLogger().error(CustomRuleFactory.class.getCanonicalName() + " could not get Plugin Descriptor from Plugin Context; will not be able to resolve " + configurationField);
			}
		} else
		{
			getLogger().warn(CustomRuleFactory.class.getCanonicalName() + " only supports Maven execution; will not be able to resolve; will not be able to resolve  " + configurationField);
		}

		return defaultValue;
	}
```

This limits the custom rule factory to maven and is dependent on internals of Maven that may change between releases and will be marked as invalid XML in IDE that validate Maven POMs.

The commit solution will eliminate the need for any of this.

dewthefifth:
I'm a fan, and think it significantly reduces the barrier to entry for writing your own custom rule factory.

ddcruver:
@joelittlejohn Could we get this pull request in?  Any concerns or objections?

joelittlejohn:
I like this, and I think we should definitely take many parts of it, however I wonder about the implementation here. I think we might be able to do it in a less intrusive way, without introducing a new Abstract class which all rules must extend.

How about this: When we instantiate a custom rule factory class, if we find a constructor that takes a GenerationConfig, we can use it. If there is no constructor that takes a GenerationConfig, we can use the default constructor as we do now. The rule factory class can easily pass this GenerationConfig value on to any rules it creates.

All existing rule factories can easily support this feature by just adding a new constructor.

Another, even more 'future proof' and extensible version of this would be: when we instantiate a new custom rule factory, we look for a method like `setGenerationConfig(GenerationConfig)`. If this method exists, we set the generation config. This would allow more fields like this to be added in future.

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## 1.0.3
+* Add support for configuration maps for Custom Rule Factories ([#1122](https://github.com/joelittlejohn/jsonschema2pojo/pull/1122))
+
```
joelittlejohn, line 5: No need to add this btw. It gets generated automatically from GitHub issues when we release.

PR #1160, opened 2020-09-16
 Inner classes

hupfdule:
Classes can be generated as inner classes per Schema URI.
This avoids multiple classes with suffixes line `__1` due to name
clashes.

This PR is based on the work of @zhouzhipeng in #1008.
It fixes a few bugs, renames the property to `useInnerClasses`, changes the behaviour to generate the inner classes per schema URI (therefore creating one top-level class per schema-URI) and provides an integration test.


jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/ObjectRule.java
```diff
@@ -64,6 +65,9 @@
  */
 public class ObjectRule implements Rule<JPackage, JType> {
 
+    /** The toplevel classes per schema URI. Used when generating inner classes. */
+    private static final Map<URI, JDefinedClass> topLevelClasses= new HashMap<>();
```
hupfdule, line 69: I am not quite happy with having to use a static map here. But since there are several ObjectRule objects created in the generation process an instance field isn't suitable here.

jsonschema2pojo-integration-tests/src/test/java/org/jsonschema2pojo/integration/config/UseInnerClassesIT.java
```diff
@@ -0,0 +1,93 @@
...
+
+    Assert.assertFalse(canLoad(classLoader, "com.example.BaseType1$AmbiguousType"));
+    Assert.assertFalse(canLoad(classLoader, "com.example.BaseType1$One"));
+
+    Assert.assertTrue(canLoad(classLoader, "com.example.BaseType2"));
+    Assert.assertTrue(canLoad(classLoader, "com.example.AmbiguousType__1"));
+    Assert.assertTrue(canLoad(classLoader, "com.example.Two"));
+
+    Assert.assertFalse(canLoad(classLoader, "com.example.BaseType2$AmbiguousType"));
+    Assert.assertFalse(canLoad(classLoader, "com.example.BaseType2$Two"));
```
hupfdule, line 65: The helper method `canLoad` is a bit suboptimal. 

I would rather like some hamcrest matcher that such a line reads like

```
assertThat(classLoader, canLoad("com.example.ClassName"))
```

but I am not used enough to hamcrest to be able to provide such a matcher.

jsonschema2pojo-integration-tests/src/test/java/org/jsonschema2pojo/integration/config/UseInnerClassesIT.java
```diff
@@ -0,0 +1,93 @@
...
+
+  }
+
+  /**
+   * This test asserts that only one top level class per schema URI is generated.
+   */
+  @Test
+  public void useInnerClasses() throws IOException {
+    ClassLoader classLoader = schemaRule.generateAndCompile("/schema.useInnerClasses",
```
hupfdule, line 74: This currently fails.

If I comment out the other test, it is working as expected. So I assume that the `schemaRule` maintains some state that breaks these tests. Any help is appreciated.

jsonschema2pojo-integration-tests/src/test/java/org/jsonschema2pojo/integration/config/UseInnerClassesIT.java
```diff
@@ -0,0 +1,93 @@
...
+                                                            "com.example",
+                                                            config("useInnerClasses", Boolean.TRUE));
```
hupfdule, line 76: This currently fails.

If I comment out the other test, it is working as expected. So I assume that the `schemaRule` maintains some state that breaks these tests. Any help is appreciated.

hupfdule:
I just rebased the PR to the current HEAD.

I would be glad if someone could review it and especially tell what is the problem with the failing test.

ctrimble:
@hupfdule This is an interesting feature.  Is this something you are still interested in working on?

PR #1173, opened 2020-11-11
GitHub actions

HanSolo:
Added a github action to run mvn verify on each push on Ubuntu 18.04 using JDK 8

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-18.04]
-        java: [8, 11, 15, 16-ea]
+        java: [8]
       fail-fast: false
       max-parallel: 4
     name: Test JDK ${{ matrix.java }}, ${{ matrix.os }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,26 @@
+---
+name: Java CI
+
+on: [push]
+
+jobs:
+  test:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-18.04]
+        java: [8, 11, 15, 16-ea]
+      fail-fast: false
+      max-parallel: 4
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,26 @@
...
+
+on: [push]
+
+jobs:
+  test:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-18.04]
+        java: [8]
```
hannesa2, line 12: What's the point to have a matrix for `os` and `java` having only one single value ?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,26 @@
...
+      fail-fast: false
+      max-parallel: 4
```
hannesa2, line 14: Without a valid matrix this is pointless too. It seems it's a copy&paste leftover

.github/workflows/ci.yml
```diff
@@ -0,0 +1,26 @@
...
+    name: Test JDK ${{ matrix.java }}, ${{ matrix.os }}
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up JDK
```
hannesa2, line 19: As far as I know, `ubuntu-18.04` has already java 8 installed

hannesa2 (approved):
Anyway, I like to see a CI here.
Maybe you find some time to simplify and reduce to necessary commands ?

joelittlejohn:
We actually use Travis for CI already. It's there any benefit to moving to GitHub actions?

hannesa2:
> We actually use Travis for CI already. It's there any benefit to moving to GitHub actions?

It's faster, it can keep some artifacts, you have a lot of additional actions, you can split tasks and you can run Espresso tests with ease. 

Maybe more advantages are there too

hannesa2:
Give it a try, the one doesn't exclude the other

.github/workflows/ci.yml
HanSolo: Just left the matrix in, in case you would like to add other operating systems or jdk‘s to test against.

PR #1186, opened 2020-12-24
Initial protoype for custom hooks

shubhamoscar:
Sample explanation for custom hooks


joelittlejohn:
Thanks for sharing the code with us and others, however this isn't something we're going to add to the core, it's too niche. It sounds like something you need, so please feel free to fork and add this and build your own version.

Issue #1162, opened 2020-09-18
@Generated annotation is not added to the generated classes


sergeykad:
The [@Generated](https://docs.oracle.com/javase/10/docs/api/javax/annotation/processing/Generated.html) annotation is handy to quickly spot generated classes.
In addition, some tools like [Google Error Prone](https://github.com/google/error-prone), rely on it to distinguish between the regular and generated code. The missing annotation makes Error Prone practically unusable since hundreds of errors are reported for the code generated by jsonschema2pojo.

joelittlejohn:
You might be interested in this solution: https://github.com/joelittlejohn/jsonschema2pojo/issues/835#issuecomment-385551880. You can remove the paths that contain generated output from Error Prone's scanning.

We originally removed @Generated because it caused hassle for Android users:

https://github.com/joelittlejohn/jsonschema2pojo/issues/577

Their classes would not compile without adding an additional library and it caused a lot of questions.

I do agree though, it's a lot nicer to annotate these generated classes correctly because a lot of tools will read that annotation.

We still support Java 8 though so I think we need to add `javax.annotation.Generated` or `javax.annotation.processing.Generated` depending on what we find is available.

sergeykad:
Thanks for the detailed explanation. Looks like I inadvertently duplicated an issue. I am closing this one and will add comments to the #835

Issue #577, opened 2016-05-23
Remove javax.annotation.Generated as it doesn't ship with Android


joelittlejohn:
The goal of adding this annotation is to let tools know that this is generated source and hence probably shouldn't be considered for things like static code analysis. This annotation doesn't ship with Android's Java though, which causes [annoyance for Android developers](http://stackoverflow.com/questions/37046649/pojos-generated-by-jsonschema2pojo-have-annotation-which-android-studio-doesnt).

Let's stop adding `javax.annotation.Generated` entirely.

farmazon3000:
Having this annotation is fine in Android because I can mark (for example for future developers of my code) that they can simplify they lives with it so I'd live it - as you can see in SO it can be easily solved in gradle (with adding a dependency) - however it would be nice to have a checkbox to enable/disable generation of this annotation.

samskiter:
It would be removed for only Android projects by default by adjusting the configuration in `GenerateJsonSchemaAndroidTask`

joelittlejohn:
@farmazon3000 As you say, it's nice to advertise the existence of this project, but this error will waste a lot of engineering time. It's simply not worth sending people to Google and forcing them to add an extra dependency. This annotation hasn't proven useful enough IMO to warrant the annoyance, it's such a trivial thing to break people's projects over.

I like the idea of forcing this off for Android projects. I'm happy to introduce a configuration option here and set the default as described by @samskiter.

Clans:
+1 for removing this

joelittlejohn:
More problems caused by having this annotation: 

https://github.com/andrewkandzuba/jsonschema2pojo-plugin-jdk9

ankon:
FWIW: I actually appreciated the `Generated` annotation on the generated files, because it made Eclipse aware that these files should not be edited.

Issue #835, opened 2018-02-25
Add javax.annotation.processing.Generated for Java 9


ben-manes:
Java 9 replaces `javax.annotation.Generated` with `javax.annotation.processing.Generated`, with the former removed for compatibility. The former is not compatible with JDK9 due to split packages (Jigsaw modules).

Unlike other static analyzers, Google's ErrorProne does not support excluding directories from analysis. Instead it uses the above annotations as an ignore rule. Since neither are generated any longer, the rules are enforced. Other analyzers are likely to follow this idiom as they become Java 9 compatible.

It would be nice to have an option to enable this annotation, with the correct one chosen based on the Java version.

joelittlejohn:
Do you think it is acceptable to check the current runtime when generating classes, or should it be possible to e.g. target Java 8 even though your build is running with Java 9?

ben-manes:
In this case the runtime is fine, since it is a backwards compatibility issue with the jdk, but I don’t have a strong preference.

ben-manes:
Closing because I got this to work in ErrorProne. I must have missed the configuration option to exclude a director.


```java
tasks.withType(JavaCompile) {
  options.errorprone {
    disableWarningsInGeneratedCode = true
    excludedPaths = "${buildDir}/generated-sources/.*"
  }
```

Art-B:
Why was this closed? It seems like a valid feature even if there is a work around for a particular scenario? It seems like having the Java version being target passed in seems reasonable. It would allow for the generation of more idiomatic or faster code etc in the future.

I am currently just researching for an option to enable the annotation even if only for the IDE warning.

ben-manes:
Sorry, I closed it since the use-case was no longer pressing and, since I wasn't going to do the work, it seemed impolite to throw on @joelittlejohn's plate unnecessarily. You are welcome to reopen / refile.

joelittlejohn:
Happy to reopen this.

sergeykad:
There is an existing Google utility that returns a correct annotation. It can be used to add a correct one depending if it is Java 8, Java 9+ or Android.

https://github.com/google/auto/blob/master/common/src/main/java/com/google/auto/common/GeneratedAnnotations.java

cyclotomic18:
How come this issue was never addressed with enough importance? Imho this feature is a required one for a (Java) code-generation tool.
[...]

PR #1202, opened 2021-03-01
Add @Generated annotation to classes and includeGeneratedAnnotation config option

cyclotomic18:
Restores the feature which added @javax.annotation.Generated annotation to generated code. In addition, the feature becomes conditional on a boolean configuration property "markGenerated" (as in JAXB. openapi ecc...), defaulted to false.
Addresses #1162 and related, without falling back to #577.

### DIFF
--- a/jsonschema2pojo-ant/src/main/java/org/jsonschema2pojo/ant/Jsonschema2PojoTask.java
+++ b/jsonschema2pojo-ant/src/main/java/org/jsonschema2pojo/ant/Jsonschema2PojoTask.java
@@ -192,6 +192,8 @@ public class Jsonschema2PojoTask extends Task implements GenerationConfig {
     private SourceSortOrder sourceSortOrder = SourceSortOrder.OS;
 
     private Map<String, String> formatTypeMapping = new HashMap<>();
+    
+    private boolean includeGeneratedAnnotation = true;
 
     /**
      * Execute this task (it's expected that all relevant setters will have been
@@ -1312,4 +1314,7 @@ public class Jsonschema2PojoTask extends Task implements GenerationConfig {
     public boolean isIncludeConstructorPropertiesAnnotation() {
         return includeConstructorPropertiesAnnotation;
     }
+    
+    @Override
+    public boolean isIncludeGeneratedAnnotation() { return includeGeneratedAnnotation; }
 }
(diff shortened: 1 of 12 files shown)


cyclotomic18:
Imo, this feature is of extreme importance in a code-genereation tool like this, for the reason i mentioned in https://github.com/joelittlejohn/jsonschema2pojo/issues/835#issuecomment-785947066. Thus, i kindly ask for an asap minor including this feature, if and once it's approved and all good.

joelittlejohn:
I'm actually tempted to turn this back on by default, with logic like:

- add `javax.annotation.Generated` if it is available
- if not, add `javax.annotation.processing.Generated` if it is available

This doesn't specifically use the target 'source compliance' version but it's a good compromise IMO. We are already moving to 1.1.0 for other reasons (and the versioning scheme in this project means that changes like this are expected).

joelittlejohn:
Out of interest, are you using Java 8?

cyclotomic18:
No, I'm using java 11, but Spring Boot packs `jackarta.annotation-api` (which contains `javax.annotation.Generated`) so I have no issue about that,
however, I mostly agree with resolving the type of `Generated` the way you exposed,,,I just wanted to address the main issue, that is the current absence of a primary feature.
I'm not sure about turning it on by default, do you mean with no prop at all or by defaulting the prop to true? For sure I wouldn't go with the first solution; the second one is pretty much just a 50/50 choice, I think everyone is good as long as the feature is present in some way...maybe it's better to stay as clean as possible with a default/minimal configuration, though? Idk.

joelittlejohn:
Yeah, I think we could keep the flag and have it true by default. This means in the CLI using something like `--omit-generated-annotation` as the flag (the presence of the flag sets it to false).

I realise you have matched jaxb, but in this project we usually name flags like this `includeXXX` (these control what is added to the ouput), so I'd prefer `includeGeneratedAnnotation` (default true) for all the other plugins.

Also, the `value` of the `Generated` annotation should be `"jsonschema2pojo"`. The name of the generator.

cyclotomic18:
I made the changes you mentioned in your last comment

joelittlejohn:
Awesome. Could you combine your commits? I'll make some further changes to detect which generated annotation is available before 1.1.0 is released.

cyclotomic18:
I rebased with master and force-pushed.
When's 1.1.0 going to be released approximately?

joelittlejohn:
@cyclotomic18 Soon I'd say. I expect it will be this week.

cyclotomic18:
Awesome, thank you for the support!

joelittlejohn:
@cyclotomic18 1.1.0 is released now 👍

cyclotomic18:
Ye, already included in my project, all good to go now. Thanks!

PR #1211, opened 2021-03-12
issue 1197 - Turns fields in the json-schema into javadoc comments. Similar to how the description field works.

markusjevringsesame:


joelittlejohn:
Thanks for implementing this Markus. There's one more thing we need before merging which is an integration test. If you check out the integration tests module you will see a load of examples. We _must_ have these to ensure that behaviour is not lost (and to prove that an example schema input _does_ produce the expected output).

markusjevringsesame:
@joelittlejohn I added the tests. None of the tests build for me in either Java 1.8 or Java 15. Not sure why. Hopefully CI will run the checks for me (and hopefully I didn't mess anything up =))

PR #1258, opened 2021-05-20
Switched from javax.validation to jakarta.validation

gerbenk:
We needed this fix for a project where we had to move from the javax to jakarta namespace. Maybe it's useful for others as well.


joelittlejohn:
Superseded by #1280

PR #1283, opened 2021-07-09
Issue 1152: existingJavaType takes precedence over $ref

jfgaretSymphony:
In the scenario where both existingJavaType and $ref are set on the same schema node, the creator of the schema intends to rely on an existing Java type for the code generation, but still wants to refer a schema for documentation and interoperability purposes.
Therefore, from code generation standpoint, existingJavaType takes precedence, and $ref is ignored.
Resolution of ref at property level is not needed in that scenario.

### DIFF
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/SchemaRule.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/SchemaRule.java
@@ -63,7 +63,7 @@ public class SchemaRule implements Rule<JClassContainer, JType> {
     @Override
     public JType apply(String nodeName, JsonNode schemaNode, JsonNode parent, JClassContainer generatableType, Schema schema) {
 
-        if (schemaNode.has("$ref")) {
+        if (schemaNode.has("$ref") && !schemaNode.has("existingJavaType")) {
             final String nameFromRef = nameFromRef(schemaNode.get("$ref").asText());
 
             schema = ruleFactory.getSchemaStore().create(schema, schemaNode.get("$ref").asText(), ruleFactory.getGenerationConfig().getRefFragmentPathDelimiters());
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/PropertyRule.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/PropertyRule.java
@@ -199,7 +199,7 @@ public class PropertyRule implements Rule<JDefinedClass, JDefinedClass> {
     }
 
     private JsonNode resolveRefs(JsonNode node, Schema parent) {
-        if (node.has("$ref")) {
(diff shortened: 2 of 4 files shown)


jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/SchemaRule.java
```diff
@@ -63,7 +63,7 @@ protected SchemaRule(RuleFactory ruleFactory) {
     @Override
     public JType apply(String nodeName, JsonNode schemaNode, JsonNode parent, JClassContainer generatableType, Schema schema) {
 
-        if (schemaNode.has("$ref")) {
+        if (schemaNode.has("$ref") && !schemaNode.has("existingJavaType")) {
```
unkish, line 66: Can it be that schema node would have both `$ref` and `existingJavaType` ?
Perhaps a test case for such scenario could be added

jfgaretSymphony: I will add a test.
I am not sure however of the desired behavior when it comes to having the root object in the schema with a $ref (and could not find a test for that).
Today, there is no class generated, other than the class corresponding to the referenced schema. Is that the desired behavior @unkish ?

unkish: > I am not sure however of the desired behavior when it comes to having the root object in the schema with a $ref (and could not find a test for that).

You're right - at present there is no test for that.

> Today, there is no class generated, other than the class corresponding to the referenced schema. Is that the desired behavior ?

It seems that it is the desired behavior as it is how the code currently works (perhaps @joelittlejohn could prove that assumption to be wrong)

joelittlejohn: In the past, the JSON Schema specification dictated that when $ref is used, all other properties *must* be ignored. The referenced schema replaces the schema node in its entirety. This was a rule in the spec up to (and including) draft-07.

For schema release '2019-09' (a.k.a. draft-08) this rule was dropped from the JSON Schema specification.

I think it's useful to allow `existingJavaType` to be used in this way. Now that it is not explicitly disallowed by the spec, I'm happy to make this change. It's a minor breaking change, it's true, but unlikely to affect anyone tbh. I don't like doing it, but it's pretty useful.

joelittlejohn: There is a strong case to allow `javaType` too IMO.

eirnym:
It depends on a schema version. Most versions don't accept any overrides along `$ref`.

ctrimble:
@jfgaretSymphony it looks like this PR has stalled out.  I am trying to get a 1.3.0 release together.  Is this something you are interested in picking back up?

Issue #1310, opened 2021-08-27
Enum not getting generated correctly with CamelCase values


iuvrayan:
I have given the following enum mapping in json schema

```json
{
   "$schema": "http://json-schema.org/draft-04/schema#",
   "additionalProperties": false,
   
   "properties": {
		"Person": {
			"$ref": "#/definitions/Person"
		}
	}, 
	"type": "object",
	
   "definitions": {
      "Person": {
         "properties": {
            "name": {"type": "string"},
            "age": {"type": "number", "minimum": 18, "maximum": 64},
            "gender": {"$ref": "#/definitions/GenderType"}
         }, 
         "required": ["name", "age", "gender"], 
         "type": "object"
      },
      
      "GenderType": {
         "enum": [
           "Male",
           "Female",
           "FeMale",
           "She_Male",
           "HE_MALE",
           "HUMAN"
         ],
         "javaEnums": [
            {"name": "Male"},
            {"name": "Female"},
            {"name": "FeMale"},
            {"name": "She_Male"},
            {"name": "HE_MALE"},
            {"name": "HUMAN"}
         ],
         "javaType": "com.example.GenderType"
      }
   }
}
```

and the produced class has the enum structure as the following:

```java
@Generated("jsonschema2pojo")
public enum GenderType {

    Male("Male"),
    Female("Female"),
    FeMale_("FeMale"),
    She_Male("She_Male"),
    HE_MALE("HE_MALE"),
    HUMAN("HUMAN");
```

I expected the FeMale to match exactly but it got generated as FeMale_

Kindly request you to fix this to obey the name given in the javaEnums for the generated classes.

joelittlejohn:
Fixed by #1329

PR #1329, opened 2021-10-08
Fix generating of enum constant names with mixed case values

unkish:
Addressing issue #1310

Please note that this might be a potential breaking change for cases when enums are defined with `javaEnumNames` or `javaEnums` and had enum alike names defined.
As outlined in the issue, current behavior is to generate:
```
    Female("Female"),
    FeMale_("FeMale"),
```
whilst new behavior would be to generate:
```
    Female("Female"),
    FeMale("FeMale"),
```


joelittlejohn:
Your contributions here have been very valuable @unkish, thank you! 

I thought I'd mention that for this kind of change, please do feel free to make your commit message look like this:

````
Fix generating of enum constant names with mixed case values

Closes #1310 
````

GitHub will then close the issue when your PR is merged.

PR #1341, opened 2021-11-29
Using TCK Tested JDK builds of OpenJDK 

carldea:
The AdoptOpenJDK has been discontinued since July 2021 (https://adoptopenjdk.net). Switching the distribution to Azul Zulu. When using Zulu you get all the latest updated (TCK Tested) builds for all versions of OpenJDK. 

Please see more details at Foojay.io:
https://foojay.io/today/github-actions-with-java-part-2/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-18.04]
-        java: [8]
+        java: [ 8.0.192, 8 ]
       fail-fast: false
       max-parallel: 4
     name: Test JDK ${{ matrix.java }}, ${{ matrix.os }}
@@ -27,7 +27,7 @@ jobs:
         uses: actions/setup-java@v2
         with:
           java-version: ${{ matrix.java }}
-          distribution: adopt
+          distribution: zulu
       - name: Cache local Maven repository
         uses: actions/cache@v2.1.6
         with:
--- a/.github/workflows/android-ci.yml
+++ b/.github/workflows/android-ci.yml
@@ -21,7 +21,7 @@ jobs:
         uses: actions/setup-java@v2
         with:
           java-version: ${{ matrix.java_version }}
-          distribution: adopt
+          distribution: zulu 
       - name: Install Android SDK
         uses: malinskiy/action-android/install-sdk@release/0.1.1
       - name: Build project


unkish:
As per information in provided link (https://adoptopenjdk.net):

> 24th July 2021: AdoptOpenJDK is moving to the Eclipse Foundation and rebranding.
Our July 2021 and future releases will come from Adoptium.net

Wouldn't it be logical to switch to `temurin` (Eclipse Temurin), which is also stated in [Supported distributions](https://github.com/actions/setup-java#supported-distributions) section of GitHub Actions ?

> NOTE: Adopt OpenJDK got moved to Eclipse Temurin and won't be updated anymore. It is highly recommended to migrate workflows from adopt to temurin to keep receiving software and security updates. See more details in the Good-bye AdoptOpenJDK post.

carldea:
> As per information in provided link (https://adoptopenjdk.net):
> 
> > 24th July 2021: AdoptOpenJDK is moving to the Eclipse Foundation and rebranding.
> > Our July 2021 and future releases will come from Adoptium.net
> 
> Wouldn't it be logical to switch to `temurin` (Eclipse Temurin), which is also stated in [Supported distributions](https://github.com/actions/setup-java#supported-distributions) section of GitHub Actions ?
> 
> > NOTE: Adopt OpenJDK got moved to Eclipse Temurin and won't be updated anymore. It is highly recommended to migrate workflows from adopt to temurin to keep receiving software and security updates. See more details in the Good-bye AdoptOpenJDK post.


It is suggested, not recommended.

Well it depends on your needs. I wrote an article above describing the scenario.
Temurin doesn’t support archived JDK versions such as `8.0.192`.
The original `setup-java@v1` used to be Zulu.
If you want Temurin then I can remove the fixed version. This would decrease your test coverage.

.github/workflows/ci.yml
```diff
@@ -16,7 +16,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-18.04]
-        java: [8]
+        java: [ 8.0.192, 8 ]
```
joelittlejohn, line 19: Why specify a specific version here?

carldea: > Why specify a specific version here?

Glad you asked! It’s often good practice. Sometimes the latest JDK version can **fail** the tests/build process and the fixed(major) version will **pass** which indicates something in the latest JDK caused the failure and not your code.

Here’s an example:
https://github.com/HdrHistogram/HdrHistogram/blob/master/.github/workflows/maven.yml

More coverage on testing goes a long way.🙂
-Carl

unkish: - But why `8.0.192` specifically, why not `8.0.292` or  `8.0.302` ?
- What about fixed versions in other PR's like:
    - https://github.com/UweTrottmann/SeriesGuide/pull/827/files
    - https://github.com/marytts/marytts/pull/1060/files

joelittlejohn:
Thanks for raising this @carldea and bringing the whole area to our attention. I like your article. I'm going to switch to temurin for now though.

eirnym:
@carldea is there difference in generated byte code between 2 different vendors with the same release versions?

carldea:
> Thanks for raising this @carldea and bringing the whole area to our attention. I like your article. I'm going to switch to temurin for now though.

Not a problem👍🏻. 


Temurin doesn’t support older fixed releases. So you may need to remove.

.github/workflows/ci.yml
carldea: Sure you can add those too🙂 . Those 8.0.192 is a known major releases for many organizations. Often there are hardened baselines in IT orgs, where major fixed releases are deployed to production, if you build on latest and don’t test for them you can’t guarantee your library will work in production.

eirnym:
By now I observe the same patch level for Java 1.8 and Java 11 from temurin and Zulu

.github/workflows/ci.yml
joelittlejohn: I see. I actually haven't heard of an open source library building against a specific JVM patch version, or organisation being stuck on a specific patch version for a very long time. It sounds like a bad thing to encourage if I'm honest, since security updates are important. 

I think those orgs will run their own tests, and I also because of the nature of our project it is very unlikely that our features will break on a different JVM patch version.

Thanks again for creating the discussion though.

carldea:
> @carldea is there difference in generated byte code between 2 different vendors with the same release versions?

JDK language byte code yes, built runtime can differ. Some vendors will backport features or have security fixes, and additional command line  switches.

PR #1354, opened 2021-12-17
Replaced deprecated mockito method calls, minor other improvements

unkish:
Replaced deprecated mockito methods that were removed in `4.0.0` (https://github.com/joelittlejohn/jsonschema2pojo/pull/1345)
Additionally:

- removed unused import statements
- removed redundant casts
- removed redundant `throws IOException`
- added `final` modifier to fields

### DIFF
--- a/jsonschema2pojo-core/src/test/java/org/jsonschema2pojo/SchemaMapperTest.java
+++ b/jsonschema2pojo-core/src/test/java/org/jsonschema2pojo/SchemaMapperTest.java
@@ -18,7 +18,6 @@ package org.jsonschema2pojo;
 
 import static org.hamcrest.MatcherAssert.*;
 import static org.hamcrest.Matchers.*;
-import static org.mockito.ArgumentMatchers.*;
 import static org.mockito.Mockito.*;
 
 import java.io.IOException;
@@ -41,7 +40,7 @@ import com.sun.codemodel.JPackage;
 public class SchemaMapperTest {
 
     @Test
-    public void generateReadsSchemaAsObject() throws IOException {
+    public void generateReadsSchemaAsObject() {
 
         final SchemaRule mockSchemaRule = mock(SchemaRule.class);
 
@@ -64,7 +63,7 @@ public class SchemaMapperTest {
     }
 
     @Test
-    public void generateCreatesSchemaFromExampleJsonWhenInJsonMode() throws IOException {
+    public void generateCreatesSchemaFromExampleJsonWhenInJsonMode() {
 
         URL schemaContent = this.getClass().getResource("/schema/address.json");
 
--- a/jsonschema2pojo-core/src/test/java/org/jsonschema2pojo/SourceSortOrderTest.java
+++ b/jsonschema2pojo-core/src/test/java/org/jsonschema2pojo/SourceSortOrderTest.java
@@ -22,29 +22,28 @@ import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.*;
 
 import java.io.File;
-import java.io.IOException;
 import java.util.Comparator;
 
 import org.junit.Test;
 
 public class SourceSortOrderTest {
     @Test
-    public void testTwoFilesAreCompared_FILES_FIRST() throws IOException {
+    public void testTwoFilesAreCompared_FILES_FIRST() {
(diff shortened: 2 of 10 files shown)


unkish:
Side question - are there any potential future plans to:

- switch to Java 11/17 ?
- switch to JUnit 5 / TestNG / ... ?
- start using `AssertJ` for assertions ?
- replacing other deprecated method invocations ?

joelittlejohn:
@unkish Thanks for this, I love it! There are a lot of things that have evolved in the testing ecosystem (and elsewhere) and you're right we should make some sweeping upgrades.

- I think it it's still important that we support Java 8 (e.g. check out the [Snyk JVM ecosystem report for 2021](https://snyk.io/jvm-ecosystem-report-2021/)), but there are a lot of improvements we can make by using Java 8 features.
- Happy to upgrade to JUnit 5 (I would prefer that over TestNG). 
- I don't think I see much benefit to moving from Hamcrest to AssertJ. Do you?

joelittlejohn:
I also did some experimentation recently using Java 11 for local development on this project. There is a problem with that upgrade that relates to the maven javadoc plugin where I had to change the configuration, and some other simple changes that I can't remember. It would be nice if the build did not fail on Java 11 as it will be increasingly annoying for people that come to contribute something to this project.

unkish:
> I don't think I see much benefit to moving from Hamcrest to AssertJ. Do you?

You're right, it wouldn't add much benefit.

I should have expressed myself better - I've tried to provide some examples of possible changes/improvements to gather thoughts.
Maybe a better example would have been question whether there have been considerations for alternatives of `com.sun.codemodel:codemodel` eg:

- https://github.com/eclipse-ee4j/jaxb-ri
- https://github.com/phax/jcodemodel
- https://github.com/forge/roaster
- https://github.com/square/javapoet
- ...
- making a fork

eirnym:
@joelittlejohn what benefits of Hamcrest/AssertJ do you see in comparison to JUnit5? In other words, which kind of assertions from additional libraries are crucial for this project?

joelittlejohn:
@unkish I do think moving to an actively maintained CodeModel would be a great idea, and relatively easy (and may fix some minor annoyances in the output). Moving to Roaster or Java Poet would be a big job, and I like how JAXB has forced CodeModel to be quite complete. I'd also be concerned about unexpected changes to output in moving to roaster or poet.

So maybe a better CodeModel is the easiest route. I think phax's CodeModel has many years of fixes and improvements, although again I think it has probably gone dormant. The Eclipse fork is active, but probably not much has changed there since the original. It looks like mostly their changes have been related to getting the library to build and integrate correctly with Java 11.

I think on balance, phax would be the best route, and we could just fork if there are any improvements left to make. I'm interested to see whether he has kept the API basically the same or introduced any breaking changes.

joelittlejohn:
@eirnym I prefer `assertThat` with matchers. There are all kinds of useful matchers and they can be composed nicely too, for very good error messages on failure.

It looks like this has been a bit of a debacle with JUnit 5 though, since `assertThat` was the 'modern' approach but then in order to avoid having any external dependencies it was removed from JUnit 5. Now only the specific assert methods remain. Since we can get assertThat from hamcrest, I don't really see a reason to go back to all these specific assert methods that JUnit provides (although I understand why they removed hamcrest as a required dependency).

joelittlejohn:
@unkish I'm definitely interested in your thoughts regarding the different code-generation options though.

eirnym:
I see your point. There's often used `assertThat(var, is("value"))` instead of `assertThat(var, equalsTo("value"))`. As per my personal experience with tests, such behavior could lead to some weird problems later on.

unkish:
@joelittlejohn my thoughts were identical

"easy" route(s):

- jaxb-ri
    - backed by strong community (eclipse)
    - seems to be mostly driven by jaxb needs, thus at present switching to it would be basically equivalent to one nail driving out another
    - possibly new "features" might start appearing in near? future
- phax's CodeModel
    - can't really say whether it's dormant - maybe it's due not being as popular/weaker community ?
    - there's might be a risk when switching to it in case it will go different paths with "historical successor" jaxb-ri

"hard" route(s):

- more effort to produce similar output as consequence likelihood of breaking something is higher
- might offer more "flexibility"
- java poet
    - appears to be dormant

Issue #1087, opened 2020-02-03
"extends" not supported in array item type


ogerardin:
**Testcase:**
flower.json:
```
{
  "type" : "object"
}
```

flowerarray.json:
```
{
  "type": "object",
  "properties": {
    "flowers": {
      "type": "array",
      "items": {
        "type": "object",
        "extends": {
          "$ref": "flower.json"
        }
      }
    }
  }
}

```

**Expected result:**
Following classes are generated:
- Flower, 
- FlowerArray with property "flowers" as array of Flower__1,
- Flower__1 extends Flower

**Actual result:**
Fails with: `Path not present: extends`
If I remove the "extends" part, it doesn't fail but Flower__1 does not extend Flower...

**Version**
1.0.2

joelittlejohn:
I don't have a solution for this but it looks like it was caused in some way by https://github.com/joelittlejohn/jsonschema2pojo/issues/408 and introduced in 0.4.15.

ogerardin:
> I don't have a solution for this but it looks like it was caused in some way by #408 and introduced in 0.4.15.

I don't really see how it's related to #408 but I trust you :)
An obvious workaround is to define the array's item type as a top-level schema, but I wanted to avoid that.

unkish:
It looks like given bug is now fixed with https://github.com/joelittlejohn/jsonschema2pojo/commit/a0bf7f57b8bb5076e321d89f8f0f25ae1d419a57 / https://github.com/joelittlejohn/jsonschema2pojo/commit/a0bf7f57b8bb5076e321d89f8f0f25ae1d419a57

joelittlejohn:
Closed as fixed :+1:

PR #1355, opened 2021-12-17
Regression test for support of 'extends' in array item type

unkish:
Regression test for #1087


joelittlejohn:
Thanks @unkish, you are a star!

Issue #1370, opened 2022-01-27
Jackson2Annotator should add JsonClassDescription for type description


BigBeaule:
Just like JsonPropertyDescription annotation is added for properties description, JsonClassDescription annocation should be added for the classes description.

Use case: Extracting all the information of a schema to generate automatically documentation for a project is easy when scanning generated classes with reflection. However, the Classes description are currently not available runtime.

N.B. A PR fixing this issue will soon be proposed.

BigBeaule:
Proposed fix: https://github.com/joelittlejohn/jsonschema2pojo/pull/1371

PR #1371, opened 2022-01-27
Added class description annotation for Jackson2

BigBeaule:
See issue #1370 for description of the problem.

Resolution is simple enough and only Jackson2 seems to have an annotation to support this feature.


ctrimble:
@BigBeaule I know this is an old PR, but I would be happy to get it into 1.3.0 if you want.  I will need it rebased on master if it is going to move forward.

BigBeaule:
@ctrimble, I missed your notification last year. The code has been rebased on master and ready to be reviewed/approved.

PR #1380, opened 2022-02-27
add support to generate subschemas as nested classes

wigbam:


wigbam:
This PR supersedes https://github.com/joelittlejohn/jsonschema2pojo/pull/1008 and https://github.com/joelittlejohn/jsonschema2pojo/pull/1160

hupfdule:
As you are writing this PR supersedes #1160, what is the reason you created this PR?
What does it do differently than #1160? What additional functionality does it provide to #1160?

wigbam:
> As you are writing this PR supersedes #1160, what is the reason you created this PR?

Not sure I fully understand the question. I have added this PR in order to add the support for the feature in question.

> What does it do differently than #1160?
- The implementation logic is different and subjectively more elegant
- It does not rely on static fields or any other questionable methods (https://github.com/joelittlejohn/jsonschema2pojo/pull/1160#discussion_r489386985)

What additional functionality does it provide to #1160?
- full unit test coverage for all affected classes
- full integration test coverage for all functionality with multiple trivial and not so trivial cases

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/ObjectRule.java
```diff
@@ -187,39 +184,37 @@ private void addParcelSupport(JDefinedClass jclass) {
...
+        if (node.has("existingJavaType")) {
+            String fqn = substringBefore(node.get("existingJavaType").asText(), "<");
 
-                JClass existingClass = resolveType(_package, fqn + (node.get("existingJavaType").asText().contains("<") ? "<" + substringAfter(node.get("existingJavaType").asText(), "<") : ""));
-                throw new ClassAlreadyExistsException(existingClass);
+            if (isPrimitive(fqn, container.owner())) {
+                throw new ClassAlreadyExistsException(primitiveType(fqn, container.owner()));
             }
 
-            boolean usePolymorphicDeserialization = annotator.isPolymorphicDeserializationSupported(node);
```
wigbam, line 217: It seems as this branch had no difference in outcome from the other one, hence it was removed.

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/ObjectRule.java
```diff
@@ -229,24 +224,21 @@ private JDefinedClass createClass(String nodeName, JsonNode node, JPackage _pack
...
+                    fqn = container.getPackage().name() + "." + fqn;
                     index = fqn.lastIndexOf(".") + 1;
                 }
 
-                if (index >= 0 && index < fqn.length()) {
+                if (index < fqn.length()) {
                     fqn = fqn.substring(0, index) + ruleFactory.getGenerationConfig().getClassNamePrefix() + fqn.substring(index) + ruleFactory.getGenerationConfig().getClassNameSuffix();
                 }
 
-                if (usePolymorphicDeserialization) {
```
wigbam, line 240: It seems as this branch had no difference in outcome from the other one, hence it was removed.

hupfdule:
Hello,

and thanks for your fast answer!

> > As you are writing this PR supersedes #1160, what is the reason you created this PR?
> 
> Not sure I fully understand the question. I have added this PR in order to add the support for the feature in question.

Well, there already was an active PR for supporting that feature. Therefore I am asking. There is not much reason to create a PR that does exactly the same thing again. So there must be some differences.
 
> > What does it do differently than #1160?
> 
> * The implementation logic is different and subjectively more elegant

I can't decide whether that's the case. Elegance is quite subjective. So that is something that the repository owner must decide.

> * It does not rely on static fields or any other questionable methods ([Generate only one top-level class per schema URI by using inner classes  #1160 (comment)](https://github.com/joelittlejohn/jsonschema2pojo/pull/1160#discussion_r489386985))

Well, that's nice. I am not happy with the fact that a static map is used in the other PR.
 
I found the following other differences between this PR and #1160:

- Usage of the term "nested class" instead of "inner class".

  I am fine with that. "Nested class" seems a bit more correct. Could be applied to the other PR as well, if desired.

- Providing a short option `-nc` additionally to the long option `--use-nested-classes`.

  Seems fine, too. Can also be applied to the other PR if desired.

- Bigger changes in `org.jsonschema2pojo.rules.ObjectRule`, `org.jsonschema2pojo.rules.TypeRule` and `org.jsonschema2pojo.util.NameHelper`. This is the main difference between both PRs.

  I didn't dive deep enough to understand why all of the changes were done, but I can directly spot these differences.

  1. This PR is more intrusive than #1160 (in regards to changed code lines). More intrusive usually means less likely to get accepted by the repository owner.
  2. This PR seems to not respect different schema files. What I mean is that it does not separate the generation of nested classes between different Schema URIs.
  3. It doesn't seem to fully support the use case it is supposed to support. See the two example JSON files from #1160:

     ```json
     {
       "type" : "object",
       "properties" : {
         "ambiguousType" : {
           "type" : "object",
           "properties" : {
             "name" : "string"
           }
         },
         "One" : {
           "type" : "object",
           "properties" : {
             "pattern" : "string"
           }
         }
       }
     }
     ```
     and
     ```json
     {
       "type" : "object",
       "properties" : {
         "ambiguousType" : {
           "type" : "object",
           "properties" : {
             "length" : "string"
           }
         },
         "Two" : {
           "type" : "object",
           "properties" : {
             "forecast" : "string"
           }
         }
       }
     }
     ```

     Both define the type `ambiguousType`. This should be generated as a nested type and therefore do not create a name clash. However with your implementation, the `ambiguousType` is getting generated as a top-level type and therefore is subject to a name clash.


I don't care much about which PR gets approved and integrated into jsonschema2pojo as long as both provide the same functionality. At the moment I see a small advantage of yours (not needing a static map) and a bigger disadvantage (generating top-level classes where nested classes should be generated and not respecting schema URI boundaries).

Unfortunately both PRs are ignored by the repository owner up until now (ignored in that there was no reaction to them; not even a negative one). So we do not know what needs to be changed to get approved (apart from the comments on #1008 which was abandoned by the author).

wigbam:
@hupfdule wow, many thanks for such a thorough review and your comments! I will address all of them individually soon, but the main priority is of course the functionality deficiency you have pointed out. Not sure how it got slipped through, I thought I had all corner cases covered. Again, thank you for pointing it out.

wigbam:
@hupfdule, I have tried to reproduce the issue with the generation you have described, but I was unable to. I have added more explicit integration tests now, specifically [canGenerateTheWholeDirectory](https://github.com/joelittlejohn/jsonschema2pojo/pull/1380/files#diff-d847115fce1a24348f604276cd49b8122a9071288ebbcb338c35dd50051593e3R62) which generates all schemas in a given directory. It passes fine and does not result in schemas defined inline being generated as top-level classes. Not sure if I am missing something?

hupfdule:
@wigbam Sorry for the late reply. I didn't notice have answered.
I tried with your current version and it still does not work for the example files above.
Did you try it with those files? Please use the test UseInnerClassesIT from my PR and get it to work with your code.

hupfdule:
Yes, it seems I was wrong. I didn't test it thoroughly, but it seems to be working.

But you should definitely not introduce changes unrelated to the actual Code. For example https://github.com/joelittlejohn/jsonschema2pojo/pull/1380/files#diff-e8700089f1b3fd621a3f5dc397def9056c689291769b3aa00222bdde3c8ff846L240. Is it necessary to remove an unused branch to fulfil the purpose of your PR? If not, then don't change it.

Also your changes to the pom should be reverted. They have nothing to do with the actual functionality you want to introduce it should not be part of this PR (you can file a second PR for that purpose if you want).

Now you should raise awareness on this PR. @joelittlejohn Can you give us some hints on whether you would accept either this PR or #1160? Would you consider to merge one of them (and which one) and what changes are necessary to get approved?

wigbam:
> Yes, it seems I was wrong. I didn't test it thoroughly, but it seems to be working.

Great! Thanks for confirming!

> But you should definitely not introduce changes unrelated to the actual Code. For example https://github.com/joelittlejohn/jsonschema2pojo/pull/1380/files#diff-e8700089f1b3fd621a3f5dc397def9056c689291769b3aa00222bdde3c8ff846L240. Is it necessary to remove an unused branch to fulfil the purpose of your PR? If not, then don't change it.

I don't agree and the reasoning is as follows:
- The class in question had to be modified for the purpose of introducing the new functionality
- The class in question did not have any unit tests to begin with
- In order for the class to be properly unit tested it would have been necessary create unit tests for both branches which would have been pointless since they are equivalent
- No integration tests were broken

> Also your changes to the pom should be reverted. They have nothing to do with the actual functionality you want to introduce it should not be part of this PR (you can file a second PR for that purpose if you want).

I can agree here, those were not related to the PR. I will revert.

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/util/NameHelper.java
```diff
@@ -286,13 +289,22 @@ private String createFullFieldName(String nodeName, String prefix, String suffix
...
-            JDefinedClass _class = _package._class(className);
-            _package.remove(_class);
+            JDefinedClass _class = container._class(className);
+            if (container instanceof JPackage) {
+                ((JPackage) container).remove(_class);
+            } else {
+                Iterator<JDefinedClass> iter = container.classes();
+                while (iter.hasNext()) {
+                    if (iter.next().equals(_class)) {
+                        iter.remove();
```
wigbam, line 301: Not completely happy about it but, unfortunately, JClass does not expose a method to remove a class.

jsonschema2pojo-ant/src/main/java/org/jsonschema2pojo/ant/Jsonschema2PojoTask.java
```diff
@@ -71,6 +71,8 @@ public class Jsonschema2PojoTask extends Task implements GenerationConfig {
 
     private boolean includeTypeInfo = false;
 
+    private boolean useNestedClasses = false;
```
wigbam, line 74: The proper full name would be `useStaticNestedClasses`, but since it is rather long, so I have shortened it.

wigbam:
I am going to keep this up-to-date in case it might get another look one day.

hupfdule:
I just noticed that your PR does not generate the nested class as `static`. Is there some reason for that? It makes using a builder more complicated as the nested class cannot be instantiated without the enclosing class. Therefore it cannot be generated in a single step.

wigbam:
> I just noticed that your PR does not generate the nested class as `static`. Is there some reason for that? It makes using a builder more complicated as the nested class cannot be instantiated without the enclosing class. Therefore it cannot be generated in a single step.

Oh boy 🤦‍♂️ somehow it slipped through the cracks. I have it working locally. Will update today/tomorrow. Many thanks for the good spot!

RanABI:
Hi guys :) Anything new with this PR? This would be a super useful addition.

hupfdule:
> Oh boy 🤦‍♂️ somehow it slipped through the cracks. I have it working locally. Will update today/tomorrow. Many thanks for the good spot!

Hi, any update on it? I’d like to use your patch for a current project.

wigbam:
> > Oh boy 🤦‍♂️ somehow it slipped through the cracks. I have it working locally. Will update today/tomorrow. Many thanks for the good spot!
> 
> Hi, any update on it? I’d like to use your patch for a current project.

Things been quite hectic over the past month, but all is fixed and updated now. Please give it a try.

hupfdule:
Sorry that it took such a long time to answer, but I didn’t find the time to check earlier.

Basically it looks good. The generated files seem to correct.

Only one small remark. In https://github.com/wigbam/jsonschema2pojo/blob/6060219f977ead894f25b6d6b2c098ef7195074d/jsonschema2pojo-maven-plugin/src/main/java/org/jsonschema2pojo/maven/Jsonschema2PojoMojo.java#L143 the annotation 

```java
@Parameter(property = "jsonschema2pojo.useNestedClasses", defaultValue = "false") 
```

should be added to let maven generate the help information for it. I don’t know why this in not generated automatically, but apparently the annotation is necessary.

wigbam:
> ```java
> @Parameter(property = "jsonschema2pojo.useNestedClasses", defaultValue = "false") 
> ```

Thanks for spotting. I have added the annotation as per above now.

doxsch:
We are very interested in this feature. Is there an update on it?

wigbam:
Even though initially this feature seemed like something that was of interest, @joelittlejohn 's latest [comment](https://github.com/joelittlejohn/jsonschema2pojo/issues/125#issuecomment-1536530017) doesn't seem to suggest that it will ever be merged, although I do not believe neither a code review nor a final decision has been made yet. So all I can do is keep this PR updated in the meantime

Alternatively, I guess forking the project and having a separate release cycle and artifact might be a more practical option. If we get a NO then that will be the way to go.

alersrt:
Looks like it can be moved to custom `RuleFactory` as an addon.

PR #1385, opened 2022-03-17
Added support for JSON-B

TheTrueDentist:
My suggestion for JSON-B-support. I have provided two annotators for JSON-B 1 and 2. They differ only in the package of the annotations though. Alternatively, one could provide a baseclass, too, and let the two extending classes only provide the annotation classes. Thiw, however, would lead to a close coupleing of both versions. Another alternative would be to have the subversion be specified by the configuration of the plugin and provide one class that handles both versions by respecting the configuration. Thought I leave that decision up to you.

### DIFF
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/Jsonb1Annotator.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/Jsonb1Annotator.java
@@ -1,0 +1,138 @@
+/**
+ * Copyright © 2010-2020 Nokia
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jsonschema2pojo;
+
+import static org.apache.commons.lang3.StringUtils.isNotEmpty;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.sun.codemodel.JAnnotationArrayMember;
+import com.sun.codemodel.JDefinedClass;
+import com.sun.codemodel.JFieldVar;
+import com.sun.codemodel.JMethod;
+
+import org.jsonschema2pojo.rules.FormatRule;
+
+import javax.json.bind.annotation.JsonbDateFormat;
+import javax.json.bind.annotation.JsonbProperty;
+import javax.json.bind.annotation.JsonbPropertyOrder;
+import javax.json.bind.annotation.JsonbTransient;
+
+import java.util.Iterator;
+
+/**
+ * Annotates generated Java types using the Jackson 2.x mapping annotations.
+ *
+ * @see <a
+ *      href="https://github.com/FasterXML/jackson-annotations">https://github.com/FasterXML/jackson-annotations</a>
+ */
+public class Jsonb1Annotator extends AbstractAnnotator {
+
+    public Jsonb1Annotator(GenerationConfig generationConfig) {
+        super(generationConfig);
+    }
+
+    @Override
+    public void propertyOrder(JDefinedClass clazz, JsonNode propertiesNode) {
+        JAnnotationArrayMember annotationValue = clazz.annotate(JsonbPropertyOrder.class).paramArray("value");
+
+        for (Iterator<String> properties = propertiesNode.fieldNames(); properties.hasNext();) {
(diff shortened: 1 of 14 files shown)


winterkid:
Excuse my impatience. Is there any chance that this pull request will be confirmed/merged soon? I can't wait bringing JSON-B POJO generation into action.

joelittlejohn:
Do we actually need to support JSONB1 at this point or is it already effectively obsolete?

If we support the annotator it needs to be changed and supported over time. Makes me think that maybe we should just support `JSONB` (which means JSONB 2).

TheTrueDentist:
I see your point and would wish I could say JSON-B 2 would be enough. Unfortunately, JSON-B 2 seems to be supported by Jakarta EE 9 on only. Jakarta EE 9 in turn is supported by Payara 6 on, which is not even available at this point. So, I don't think we come around JSON-B 1 at this point.
As the only difference in the implementation ist the package name, one could imagine a config parameter to override the package name, but that would seem be rather hacky, I'm afraid.

joelittlejohn:
Makes sense @TheTrueDentist, thanks!

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/Jsonb1Annotator.java
```diff
@@ -0,0 +1,138 @@
...
+
+import javax.json.bind.annotation.JsonbDateFormat;
+import javax.json.bind.annotation.JsonbProperty;
+import javax.json.bind.annotation.JsonbPropertyOrder;
+import javax.json.bind.annotation.JsonbTransient;
+
+import java.util.Iterator;
+
+/**
+ * Annotates generated Java types using the Jackson 2.x mapping annotations.
```
unkish, line 37: Looks like a copy/paste from Jackson annotator

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/Jsonb2Annotator.java
```diff
@@ -0,0 +1,138 @@
...
+
+import java.util.Iterator;
+
+import jakarta.json.bind.annotation.JsonbDateFormat;
+import jakarta.json.bind.annotation.JsonbProperty;
+import jakarta.json.bind.annotation.JsonbPropertyOrder;
+import jakarta.json.bind.annotation.JsonbTransient;
+
+/**
+ * Annotates generated Java types using the Jackson 2.x mapping annotations.
```
unkish, line 37: Looks like copy/paste from Jackson annotator

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/Jsonb1Annotator.java
TheTrueDentist: Yeah, I based my work on the Jackson annotator as the annotations and handling are quite similar, although not entirely the same. At the end, i copied JSON-B-annotator v1 for v2. This ways, I forgot to fix thecomments and then copied the errors from one version to the other, sorry. I guess the iterator could also be improved by using for-each.

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/Jsonb2Annotator.java
TheTrueDentist: See above.

TheTrueDentist:
I have provided ITs as suggested. Note that I have splitted the test class into two, one per version of JSON-B to decrease confusion. Please also note, that I had to use two different JSON-B implementations. This is because I would have Yasson 1 for JSON-B 1 and Yasson 2 for JSON-B 2 (Yasson 2 is seemingly not backwards compatible). When I tried to set that up, things did not work, wrong version of yasson was used. So I tricked it out and used Johnzon for JSON-B 1.

winterkid:
So, everythings fine now?

TheTrueDentist:
There seems to be a status report required about test result on Java 8 under Ubuntu 18.04 before this can be pulled. I have no Ubuntu 18.04 at hand, but under 20.04 with Java 8 tests are all green. Of course, if Ubuntu 18.04 is the reference here, tests should be run there as well. Also, I see that you want to check the results yourself. Still, the setup I have used makes me confident to state: I would not expect any problems under Ubuntu 18.04 with Java 8.

joelittlejohn:
Just had another look through this one and I think the last thing that needs to be done is an update to some docstrings:

* https://github.com/joelittlejohn/jsonschema2pojo/blob/master/jsonschema2pojo-ant/src/site/Jsonschema2PojoTask.html#L36
* https://github.com/joelittlejohn/jsonschema2pojo/blob/master/jsonschema2pojo-maven-plugin/src/main/java/org/jsonschema2pojo/maven/Jsonschema2PojoMojo.java#L245
* https://github.com/joelittlejohn/jsonschema2pojo/blob/master/jsonschema2pojo-gradle-plugin/README.md?plain=1#L109

TheTrueDentist:
You're right, of course. Documentation should be fixed now.

winterkid:
Is milestone 1.1.3 about to be released this year?

Issue #166, opened 2014-03-26
Lombok support


cchacin:
Less boilerplate code.

http://projectlombok.org/

DanielWeigl:
+1

joelittlejohn:
Sounds like Lombok is getting more and more traction.

lucasmorano:
+1

guptaavi:
+1

yaitskov:
must be

akashborkar:
+1

volkert-fastned:
What I miss in particular with the builder methods generated by jsonschema2pojo is that they don't seem to process the `@Nullable` annotations.

When you let Lombok generate the builder methods in the POJOs instead (through Lombok's `@Builder` annotation), you get explicit null-checks for free, since Lombok checks for `@Nullable` field annotations when it generates builder methods. This provides a nice fail-fast mechanism for fields that are never allowed to be null. The builder's `.build()` method will fail immediately with an `IllegalArgumentException` and a clear message indicating which required field was omitted. I love that in Lombok!

Right now with the current internal builder implementation of jsonschema2pojo, required fields with null values are not noticed until they are deserialized on the other end, or when explicitly validated against a schema. 😞 Please let Lombok provide the builder methods instead without reinventing the wheel. Thanks for considering this! 🙂

lindsaykwardell:
Having Lombok support probably won't change whether I use this or not (as you noted the boilerplate doesn't matter as much), but I think it would certainly be helpful for reading the auto-generated files. I found this project looking through auto-generated objects from this tool, and not having to visually parse the get/set methods would be wonderful.

CayoSouza:
+1

KaiStapel:
+1

ogerardin:
What is the status of this ? I already use lombok throughout the project so for the sake of consistency I'd love to see this implemented.

sridivakar:
I prefer creating Immutable classes with lombok's `@Value` and `@Builder(toBuilder = true)`, its simple way to achieve thread-safety of POJOs.

With these annotations, if there is a way to annotate the Java POJOs generated from from json schema, and support to serialize/deserialize to/from json then it will be of great value for achieving thread-safety for generated Java POJOs .

Alternatively can we use `org.jsonschema2pojo.Annotator`/`customAnnotator`  to achieve the same, will serialization/deserialization to/from json supported with such lombok annotated POJOs ?

eirnym:
[...]

PR #1394, opened 2022-04-12
Add config "extraClassAnnotations" (enabler for Lombok usage)

goekay:
this PR introduces a new config `extraClassAnnotations` that allows the user to specify fully qualified class names of annotations that should be put on _all_ classes during code generation.

its primary motivation is to enable the usage of lombok without this library depending on lombok. and hence the generic solution approach. if wanted, the user can extend the use case to other annotations.

the main work is in core module. maven, cli, ant and gradle modules are adjusted accordingly to enable this config with all entry points. the following examples are for the maven plugin.

as an example, in order to use it with lombok, the following configurations can be included:

```
<configuration>
    ....
    <!-- either getters or setters have to be enabled, because jsonschema2pojo makes the fields public otherwise -->
    <includeGetters>true</includeGetters>
    <includeSetters>false</includeSetters>
    <extraClassAnnotations>
        <annotation>lombok.NoArgsConstructor</annotation>
        <annotation>lombok.Setter</annotation>
        <annotation>lombok.ToString</annotation>
        <annotation>lombok.EqualsAndHashCode</annotation>
    </extraClassAnnotations>
    ....
</configuration>
```

if having public fields is not an issue, another example would be:

```
<configuration>
    ....
    <includeGetters>false</includeGetters>
    <includeSetters>false</includeSetters>
    <extraClassAnnotations>
        <annotation>lombok.Data</annotation>
    </extraClassAnnotations>
    ....
</configuration>
```

keep in mind that lombok (or the annotation the user is trying to add) should be included in the project executing this plugin.

References:
- https://github.com/joelittlejohn/jsonschema2pojo/issues/166
- https://github.com/joelittlejohn/jsonschema2pojo/issues/166#issuecomment-1079443937

### DIFF
--- a/jsonschema2pojo-ant/src/main/java/org/jsonschema2pojo/ant/Jsonschema2PojoTask.java
+++ b/jsonschema2pojo-ant/src/main/java/org/jsonschema2pojo/ant/Jsonschema2PojoTask.java
@@ -113,6 +113,8 @@ public class Jsonschema2PojoTask extends Task implements GenerationConfig {
 
     private Class<? extends RuleFactory> customRuleFactory = RuleFactory.class;
 
+    private String[] extraClassAnnotations = new String[] {};
+
     private boolean includeJsr303Annotations = false;
 
     private boolean includeJsr305Annotations = false;
@@ -577,6 +579,16 @@ public class Jsonschema2PojoTask extends Task implements GenerationConfig {
         } else {
             this.customRuleFactory = RuleFactory.class;
         }
+    }
+
+    /**
+     * Sets the 'extraClassAnnotations' property of this class
+     *
+     * @param extraClassAnnotations
+     *            A list of fully qualified annotation class names to be applied to Java classes.
+     */
+    public void setExtraClassAnnotations(String[] extraClassAnnotations) {
+        this.extraClassAnnotations = extraClassAnnotations;
     }
 
     /**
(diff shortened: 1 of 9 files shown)


migroskub:
@joelittlejohn Why doesn't this getting merged already?

princefisher:
@goekay the PR shows as out of sync with base branch, any chance you can update the PR?
@joelittlejohn is there anything else required to get this feature merged?

I really like the concept of Lombok integration and our team would greatly benefit from it as well.

goekay:
@princefisher done, thanks for the heads up.

octgsoftware:
Any update on this feature?  I would also like to add a property that will override the field access regardless if 					
includeGetters and includeSetters is false.

Currently I had to use an RuleFactory and PropertyRule just to have private fields if I disabled the getters and setters.  I also had to implement  an AbstractAnnator.

In my case I would want JMod.PRIVATE since Lombok would be generating the accessors.

Jmod.NONE could also be used with @FieldDefaults(level=AccessLevel.PRIVATE)

octgsoftware:
@goekay Will this PR handle annotations with properties like @Accessors(fluent = true, chain = true) or @FieldDefaults(level=AccessLevel.PRIVATE)?

goekay:
@octgsoftware no. the properties of the annotations are not handled.

goekay:
this is where i give up.

darkmastermindz:
@joelittlejohn @goekay Is it possible if we take another go at this?

PR #1397, opened 2022-04-30
Use LinkedHashMap for additional properties, so the original ordering in the JSON is preserved

jdimeo:
Original GitHub issue:
> It's quite annoying that additionalProperties values lose their initial array order when json is parsed.
> Since I use the same schema classes to generate json after tweaking, I can't sort them too.
> I guess it will be easy to add an option to generate LinkedHashMap instead of HashMap for additionalProperties field.

### DIFF
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/AdditionalPropertiesRule.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/AdditionalPropertiesRule.java
@@ -16,7 +16,7 @@
 
 package org.jsonschema2pojo.rules;
 
-import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Spliterator;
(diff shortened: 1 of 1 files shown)


rohanKanojia:
@jdimeo @joelittlejohn : Hello, Is there any way to revert to the old `java.util.HashMap` type for `additionalProperties`? We're trying to migrate to v1.2.1 and facing this issue.

joelittlejohn:
@rohanKanojia what's the issue you're facing?

rohanKanojia:
@joelittlejohn :  Sorry for the wording, it's not really an issue but a change. I'm wondering if any KubernetesClient user could complain about this.

We use JsonSchema2Pojo to generate POJOs for Kubernetes objects in  https://github.com/fabric8io/kubernetes-client

After the bump to 1.2.1 in https://github.com/fabric8io/kubernetes-client/pull/4895 , we have all the generated POJOs modified with `additionalProperties` type update. I just wanted to check if there is some way to avoid this.

rohanKanojia:
@jdimeo : Thanks for the clarification!  I'll push for this change in fabric8 k8s client.

logeshwaranRR:
> Original GitHub issue:
> 
> > It's quite annoying that additionalProperties values lose their initial array order when json is parsed.
> > Since I use the same schema classes to generate json after tweaking, I can't sort them too.
> > I guess it will be easy to add an option to generate LinkedHashMap instead of HashMap for additionalProperties field.

@jdimeo is there any way to use  java.util.HashSet or java.util.TreeSet in Json schema? because the user would want to use any collection type they want

jdimeo:
Not for `additionalProperties` no, because it's specifically a map of keys to values

PR #1401, opened 2022-05-05
Adds builder method when using innerBuilder

### LINKED ISSUE #1396: Add `builder` method to create a new instance of the builder
When using (in the context of the maven builder) `useInnerClassBuilders` and `generateBuilders` set to `true` it would be nice to have a `builder` method to generate a new builder. 

i.e.
Instead of:
```java
new MyGeneratedObject.MyGeneratedObjectBuilder()
    .withFoo("bar")
    .build();
```

Have something like:
```java
MyGeneratedObject.builder()
    .withFoo("bar")
    .build();
```

It could be an opt-in by a new setting e.g. `addBuilderMethodForInnerClassBuilders`.

I'm supposing the same applies for the core, but haven't checked all the details other thant the [BuilderRule](https://github.com/joelittlejohn/jsonschema2pojo/blob/master/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/BuilderRule.java) file where I suppose it needs to be modified.

josejulio:
This PR updates the code to add a `public static` method called `builder` to all the classes that are created with the `generateBuilders` and `useInnerClassBuilders`.

This is a sample of how the builder method looks like:

```
public static Child.ChildBuilderBase builder() {
    return new Child.ChildBuilder();
}
```

Fixes #1396

### DIFF
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/BuilderRule.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/BuilderRule.java
@@ -110,6 +110,10 @@ public class BuilderRule implements Rule<JDefinedClass, JDefinedClass> {
       generateNoArgsBuilderConstructors(instanceClass, builderClass, concreteBuilderClass);
     }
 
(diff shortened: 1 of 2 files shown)


joelittlejohn:
Just closing/opening to trigger checks.

PR #1406, opened 2022-05-15
Support - as reading from input stream

maxandersen:
idea I had was to enable using - as reading from standard input so you can do something like:

```
curl https://api.github.com/users/maxandersen | jbang --debug org.jsonschema2pojo:jsonschema2pojo-cli:1.1.3-SNAPSHOT --source - --target out -T json
```

would that be interesting ? (if yes I'll investigate adding tests for it)

things I noticed doing this that i'm not sure is expected or bugs:

when treating input as schema the same url is requested multiple times in ContentResolver requiring me to add a cache for standard input...maybe should instead do it generically instead of reading url multiple times?

when treating input as json then content resolvers are NOT used but instead read directly so had to add - handling of urls redundantly.

wdyt ?

### DIFF
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/ContentResolver.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/ContentResolver.java
@@ -51,6 +51,8 @@ public class ContentResolver {
                 .enable(DeserializationFeature.USE_BIG_DECIMAL_FOR_FLOATS);
     }
 
+    static JsonNode cachedStdIn;
+
     /**
      * Resolve a given URI to read its contents and parse the result as JSON.
      * <p>
@@ -73,7 +75,14 @@ public class ContentResolver {
         }
 
         try {
-            return objectMapper.readTree(uri.toURL());
+            if(uri.getPath().endsWith("/-")) {
+                if(cachedStdIn==null) {
+                    cachedStdIn = objectMapper.readTree(System.in);
+                }
+                return cachedStdIn;
+            } else {
+                return objectMapper.readTree(uri.toURL());
+            }
         } catch (JsonProcessingException e) {
             throw new IllegalArgumentException("Error parsing document: " + uri, e);
         } catch (IOException e) {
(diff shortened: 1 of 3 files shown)


maxandersen:
Anyone with a comment ?

joelittlejohn:
This is a nice little addition max, thanks for submitting.

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/ContentResolver.java
```diff
@@ -51,6 +51,8 @@ public ContentResolver(JsonFactory jsonFactory) {
                 .enable(DeserializationFeature.USE_BIG_DECIMAL_FOR_FLOATS);
     }
 
+    static JsonNode cachedStdIn;
```
joelittlejohn, line 54: Could you explain why you introduced a static member here?

Why not just:

```java
            if(uri.getPath().endsWith("/-")) {
                return objectMapper.readTree(System.in);
            } else {
                return objectMapper.readTree(uri.toURL());
            }
```

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/Jsonschema2Pojo.java
```diff
@@ -168,6 +168,10 @@ public static String getNodeName(URL file, GenerationConfig config) {
 
     public static String getNodeName(String filePath, GenerationConfig config) {
         try {
+            if(filePath.endsWith("/-")) {
+                // if standard input '-' use base directory as name.
+                filePath=filePath.substring(0,filePath.length()-2);
```
joelittlejohn, line 173:
```suggestion
                filePath = filePath.substring(0, filePath.length() - 2);
```

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/SchemaGenerator.java
```diff
@@ -53,8 +53,14 @@ public SchemaGenerator(JsonFactory jsonFactory) {
     public ObjectNode schemaFromExample(URL example) {
 
         try {
-            JsonNode content = this.objectMapper.readTree(example);
+            JsonNode content = null;
```
joelittlejohn, line 56:
```suggestion
            JsonNode content;
```

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/ContentResolver.java
maxandersen: Because its called multiple times when dealing with a Schema. Can only read the input once.

maxandersen:
It's been a while since I replied and updated based on comments. 

Any chance or merging this?

Issue #1242, opened 2021-04-22
Parent properties not generated directly on Child inner class builders.


Woodham:
When using inner-class builders with a Child schema extending a Parent schema, builder methods for properties on the Parent aren't directly generated on the Child's builder. 

This means that with `withX()` methods on the Child builder return the Parent's builder type and make it difficult to construct an instance of the Child without casting. 

e.g. (Using examples from the inner builder tests)

Child:
```
{
  "type" : "object",
  "extends" : {
    "$ref" : "parent.json"
  },
  "properties" : {
    "childProperty" : { "type" : "integer"},
    "sharedProperty" : { "type" : "string" }
  }
}
```

Parent:
```
{
  "type" : "object",
  "properties" : {
    "parentProperty" : { "type" : "string" },
    "sharedProperty" : { "type" : "string" }
  },
  "required" : ["parentProperty"]
}
```
  
This won't work:

```
Child child = new Child.ChildBuilder()
    .withChildProperty(1)
    .withSharedProperty("test")
    .withParentProperty("test2")
    .build();
```

A Cast is required, since `withParentProperty()` returns `Parent.ParentBuilder`:

```
Child child = (Child) new Child.ChildBuilder()
    .withChildProperty(1)
    .withSharedProperty("test")
    .withParentProperty("test2")
    .build();
```

It looks like similar functionality was added for non inner-class style builders in #458.

PR #1416, opened 2022-07-15
Cast parent builder to this builder return type

nielspardon:
This PR adds wrapper methods to the builder class calling the `withXXX` builder method of the parent builder and casting the result to the class of the current builder, e.g.:

```java
public static abstract class ChildBuilderBase<T extends Child> extends Parent.ParentBuilderBase<T> {
  ...
  @Override
  public Child.ChildBuilderBase<T> withTimestamp(Instant timestamp) {
    return ((Child.ChildBuilderBase<T> ) super.withTimestamp(timestamp));
  }
  ...
}
```

This allows to actually use inherited builders with a fluent API.

Fixes https://github.com/joelittlejohn/jsonschema2pojo/issues/1242


joelittlejohn:
Looks like there is some problem with the way this interacts with the 'inner builder' option.

nielspardon:
> Looks like there is some problem with the way this interacts with the 'inner builder' option.

Sorry for that. I think I found a better solution that avoids this issue.

PR #1442, opened 2022-10-17
correct gradle default values for includeGetters/includeSetters

nils-a:
The default for both values is `true`, according to
 https://github.com/joelittlejohn/jsonschema2pojo/blob/2d55620ac925ff7406f2982a2c37d49a6697cd20/jsonschema2pojo-gradle-plugin/src/main/groovy/org/jsonschema2pojo/gradle/JsonSchemaExtension.groovy#L146-L147

### DIFF
--- a/jsonschema2pojo-gradle-plugin/README.md
+++ b/jsonschema2pojo-gradle-plugin/README.md
@@ -218,10 +218,10 @@ jsonSchema2Pojo {
   serializable = false
 
   // Whether to include getters or to omit these accessor methods and create public fields instead.
-  includeGetters = false
+  includeGetters = true
 
   // Whether to include setters or to omit these accessor methods and create public fields instead.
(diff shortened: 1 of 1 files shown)


unkish:
Hi

I'm not sure whether the intent was to express that:

_default values will be used if not set (following "example" configuration)_
or
_below is list of default configuration values_

If it's former - perhaps the wording of it could/should be changed (native English speakers input would be nice in such case).
If it's the latter, then there are more "invalid" values than the ones proposed to be fixed, listing some of them:

- `source`
- `targetDirectory`
- `includeAdditionalProperties`
- `propertyWordDelimiters`
- `customAnnotator`
- ...

joelittlejohn:
It says in the README that `// Each configuration is set to the default value`. We just make that true I think..

PR #1456, opened 2022-12-21
Use Object as the type in case schema property contains mixed types

### LINKED ISSUE #1425: Generated attribute takes first type on types list.
Consider the following jsonSchema:
```json
  {
        "type" : "object",
        "properties" : {
            "foo" : {
                "type" : ["boolean", "string","number","integer"]
            }
        }
    }
```

In this case generated attribute `foo` will have as type `Boolean`:

```java
package com.example;

import javax.annotation.Generated;

@Generated("jsonschema2pojo")
public class Example {

public Boolean foo;

}
```

I noticed that, if type is not specified in jsonSchema, foo takes the type `Object`:

```json
  {
        "type" : "object",
        "properties" : {
            "foo" : {
            }
        }
    }
```
Result:

```java
package com.example;

import javax.annotation.Generated;

@Generated("jsonschema2pojo")
public class Example {

public Object foo;

}
```
There's a way of having the last result (without changing the jsonSchema) when a list of types is passed ?

unkish:
Closes #1425


jsonschema2pojo-integration-tests/src/test/java/org/jsonschema2pojo/integration/TypeIT.java
```diff
@@ -282,13 +282,13 @@ public void useDoubleNumbersFalseCausesPrimitiveNumbersToBecomeFloats() throws C
     }
 
     @Test
-    public void unionTypesChooseFirstTypePresent() throws ClassNotFoundException, SecurityException, NoSuchMethodException {
+    public void unionTypesChooseFirstTypePresent() throws ReflectiveOperationException {
```
joelittlejohn, line 285: Could we add one more check here, where the `"null"` comes first in the array?

So `["null", "boolean"]` should produce a java.lang.Boolean property.

unkish: Sure

joelittlejohn:
Yes, you're right.

PR #1464, opened 2023-01-18
Use decorated JsonFactory backed by SchemaGenerator when resolving JSON/YAML source inputs

### LINKED ISSUE #1427: "Path not present" error when processing multiple JSON files having objects with the same name
jsonschema2pojo 1.1.2 fails with "Path not present: watcher" error when processing multiple json files having objects with the same name, e.g.:

- JiraCreateIssueRequest.json https://docs.atlassian.com/software/jira/docs/api/REST/8.1.0/#api/2/issue-createIssue
- JiraIssue.json https://docs.atlassian.com/software/jira/docs/api/REST/8.1.0/#api/2/issue-getIssue

... and the following configuration:

        <plugins>
            <plugin>
                <version>1.1.2</version>
                <configuration>
                    <generateBuilders>true</generateBuilders>
                    <annotationStyle>none</annotationStyle>
                    <includeAdditionalProperties>false</includeAdditionalProperties>
                    <initializeCollections>false</initializeCollections>
                </configuration>
                <executions>
                    <execution>
                        <id>execution-with-error</id>
                        <goals>
                            <goal>generate</goal>
                        </goals>
                        <configuration>
                             <sourceDirectory>${basedir}/src/main/resources/schema</sourceDirectory>
                             <sourceType>json</sourceType>
                        </configuration>
                    </execution>
                </executions>
            </plugin>
        </plugins>

The error persists if I move json-files into different subdirectories of "sourceDirectory". I expect that json-files in different subdirectories will result in POJOs with different packages to avoid naming conflicts.

unkish:
Closes #1427


unkish:
Some notes:

- removed `YamlRefIT` as it was empty (dead code)
- removed `@Ignore` from `ConstructorsIT` as it caused a false claim of test being ignored
- fixed some typos in `ConstructorsIT`
- rewritten `SelftRefIT::nestedSelfRefsInStringContentWithoutParentFile` such that it would act as IT rather than unit test (`schemaRule.generateAndCompile(...)` vs. `new SchemaMapper().generate(...)`)
- couldn't spot any tests for https://github.com/joelittlejohn/jsonschema2pojo/blob/8e772064776c51376d63c35674e6dd5a0d5e123c/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/SchemaMapper.java#L113-L114
- spotted several places creating `ObjectMapper` with some features enabled https://github.com/joelittlejohn/jsonschema2pojo/blob/8e772064776c51376d63c35674e6dd5a0d5e123c/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/SchemaMapper.java#L140-L142 Would it make sense to extract that into separate class ?

Feedback is welcome

joelittlejohn:
@unkish Thanks for submitting this.

I'm struggling to understand exactly what's going on in #1427 and why your changes fix the problem. Maybe you could give me a summary of the problem you found and what you had to do to fix it?

It looks like you are wanting to unify the way that inputs are read using the JsonFactory abstraction, so that the source type applies correct **whenever** any file is read, not just at the start. Is that right?

unkish:
Hi

Should have provided additional justification/clarification it in the first place as I've struggled to understand what is the cause myself :/

It's related to the way how schemas are getting stored/retrieved in/from `SchemaStore` and manifests itself if there are several files processed.
Following assumes that we have 2 files: `AResponse.json` and `BResponse.json` which do not reference each other but have a object property with same name in both of them.
When files are passed as yaml/json schema, `SchemaMapper` wraps paths to them into `$ref` after that they end up being added to `SchemaStore` through https://github.com/joelittlejohn/jsonschema2pojo/blob/fccc0161afd787ece9f1e25e9e6f6638d8630e8e/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/SchemaStore.java#L140 https://github.com/joelittlejohn/jsonschema2pojo/blob/fccc0161afd787ece9f1e25e9e6f6638d8630e8e/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/SchemaStore.java#L54 with id's having URI as prefix and path appended to them eg.:

- `"file:/exampleSchemasWithSameName/AResponse.json" -> {Schema@2479}`
- `"file:/exampleSchemasWithSameName/AResponse.json#/properties/fieldA" -> {Schema@3503}`
- `"file:/exampleSchemasWithSameName/AResponse.json#/properties/common" -> {Schema@3504}`
- `"file:/exampleSchemasWithSameName/AResponse.json#/properties/common/properties/diffA" -> {Schema@3505}`
- `"file:/exampleSchemasWithSameName/BResponse.json" -> {Schema@3513}`
- `"file:/exampleSchemasWithSameName/BResponse.json#/properties/fieldB" -> {Schema@3519}`
- `"file:/exampleSchemasWithSameName/BResponse.json#/properties/common" -> {Schema@3526}`
- `"file:/exampleSchemasWithSameName/BResponse.json#/properties/common/properties/diffB" -> {Schema@3576}`

That way they can be/are uniquely referenced/resolved.

However when files are passed as yaml/json examples content is first transformed via `SchemaGenerator` and then stored/resolved in `SchemaStore` through https://github.com/joelittlejohn/jsonschema2pojo/blob/fccc0161afd787ece9f1e25e9e6f6638d8630e8e/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/SchemaStore.java#L128-L138 with id's being:

- `"#/properties/common/properties/diffA" -> {Schema@3393} `
- `"#/properties/fieldB" -> {Schema@3417}`
- `"#/properties/fieldA" -> {Schema@2436}`
- `"#/properties/common" -> {Schema@3384}`

As id's are no longer unique across files then when `#/properties/common` is being resolved/looked up for `BResponse.json` it ends up getting "cached" value `"#/properties/common" -> {Schema@3384}` of `AResponse.json`.


\
So when we're dealing with input from URI's, it seemed that if we won't treat schemas different from samples (logic processing schemas shouldn't care whether input was from sample or not) then they'd be processed in similar way.
`JsonFactoryProvider` was supposed to provide `JsonFactory` to be used by `ContentResolver`/`SchemaMapper` such that it would look to latter as if they're working with json schema input (example transformation to schema would happen _ad hoc_).
I've left some of the `public` methods in place in an attempt to maintain backwards compatibility.
It should be noted though that this fix won't help if someone uses own `JsonFactory` that won't act as instance provided by `JsonFactoryProvider` whilst passing multiple example files with similar property names - issue would re-surface itself.

joelittlejohn:
I'm going to release 1.1.3 before merging this, as it's quite substantial and I'd like to get a 'good' release out that resolves the 'path not present' problem before we make any further fundamental change. Nothing stopping us releasing 1.1.4 soon though.

unkish:
Sure, no problem.
I won't mind if someone will find/propose another way of fixing issue and we'll drop this PR :)

unkish:
Added also copyright headers to java files (how embarrassing that I didn't notice their absence previously 🤦‍♂️)

unkish:
Superseded by #1484.

PR #1465, opened 2023-01-21
Encode property names when building path to property

### LINKED ISSUE #1402: Properties with special character in the name (#, /, etc) cause IllegalArgumentException 'Path not present' when using 1.1.2
After updating to version [org.jsonschema2pojo:jsonschema2pojo-gradle-plugin](https://togithub.com/joelittlejohn/jsonschema2pojo) `1.1.2` of the plugin, we are no longer able to generate classes when the attributes contain `#`
This was possible on version `1.1.1`
For example `VERSV#`

```
{
  "$schema": "http://json-schema.org/draft-06/schema#",
  "$ref": "#/definitions/BusinessData",
  "definitions": {
    "BusinessData": {
      "type": "object",
      "properties": {
        "VERSV#": {
          "type": "string"
        },
        "HEADER": {
          "$ref": "#/definitions/Header"
        }
      },
      "title": "BusinessData"
    }
}
```

joelittlejohn:
When creating a fragment with JSON Pointer path, we now encode property names using the rules from RFC 6901. This means that the property name can be treated as a single token, even if it contains special characters like # or /.

Closes #1402

### DIFF
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/JsonPointerUtils.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/JsonPointerUtils.java
@@ -1,0 +1,52 @@
+/**
+ * Copyright © 2010-2020 Nokia
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jsonschema2pojo;
+
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.ArrayList;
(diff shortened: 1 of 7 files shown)


unkish:
I'll try to have a look at it during next week if that's OK.

Off topic: `Android Sample / Build JDK 8 (pull_request)`  seems to fail due to [`moshi` dependency upgrade](https://github.com/joelittlejohn/jsonschema2pojo/commit/4c9bde2008c3264088f9f49058d5474163bcd99b#r97666212)

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/JsonPointerUtils.java
```diff
@@ -0,0 +1,52 @@
...
+    }};
+
+    public static String encodeReferenceToken(final String s) {
+        String encoded = s;
+        for (Map.Entry<String,String> sub : SUBSTITUTIONS.entrySet()) {
+            encoded = encoded.replace(sub.getKey(), sub.getValue());
+        }
+        return encoded;
+    }
+    public static String decodeReferenceToken(final String s) {
```
unkish, line 41: Just a nitpick: a new line between L40-L41 might increase readability

joelittlejohn: I need to turn off all the extra annotations that Intellij does in the editor. It makes it really confusing to see where the actual new lines are. I'm trying to ween myself off Eclipse you see :disappointed:

PR #1466, opened 2023-01-22
Add a simple build.gradle to allow publishing to plugins.gradle.org

### LINKED ISSUE #282: Add to Gradle Plugin Portal
This should be done for discoverability to promote js2p and has no functionality changes. It is required if using the new plugin DSL, but that's not used in practice yet.

The [steps](https://plugins.gradle.org/docs/submit) appear minimal. It does require changing to the new Gradle plugin naming convention, so there is a little upgrade pain.

JCenter is a two-way mirror of Maven Central. Ideally this requires a small change to the build while still publishing as usual. If the new attributes can only be provided if publishing to JCenter directly, then this won't be worth the effort.

joelittlejohn:
The important file here is `jsonschema2pojo-gradle-plugin/build.gradle`. We'll give some thought to how to keep the version number up to date, but my plan is to add a step in the release process to run `./gradlew publishPlugins`.

Closes #282.


unkish:
I wonder if git tag's could be used for that 🤔

joelittlejohn:
@unkish Excellent point. Ideally, we should just be able to tag a new release, and all activities (including everything that happens in the [upload release script](https://github.com/joelittlejohn/jsonschema2pojo/blob/82a1352024f0cbeb5d95aa09b39dd36f363e2f4b/jsonschema2pojo-upload-release)) should be triggered be the creation of a new tag.

unkish:
Another option might be to do similar solution as Rafael Winterhalter has in [`byte-buddy-gradle-plugin`](https://github.com/raphw/byte-buddy/tree/master/byte-buddy-gradle-plugin).
It looks like the version is being read from `pom.xml` (in `bytebuddy.gradle` which is loaded/applied in `build.gradle`) and publishing is happening via `exec-maven-plugin` by invoking `gradle` and passing params to it.

joelittlejohn:
Great plan @unkish. I was thinking exactly the same, and I have already made some local changes this week to take version as an arg on the command line. Seems to work well.

joelittlejohn:
We can either call from mvn during release, or just invoke as a release step from bash.

Issue #632, opened 2016-10-14
Object name of "_" causes exception


Smitty010:
I have some json I'm trying to process that uses the single character "_" as the name (not the value) of one of the fields in the json.  If I had the option of changing the name of the field I would.  But, it's a third party library (apache solr).  At any rate, the normalizeName() function throws an exception.

Exception in thread "main" java.lang.StringIndexOutOfBoundsException: String index out of range: 0
        at java.lang.String.charAt(Unknown Source)
        at org.jsonschema2pojo.util.NameHelper.normalizeName(NameHelper.java:45)

        at org.jsonschema2pojo.util.NameHelper.getPropertyName(NameHelper.java:76)

joelittlejohn:
I guess we can fix this quite easily. Do you have any thoughts about what you'd _like_ to see generated when we encounter a property called `_`?

Smitty010:
First, to be clear, I won't tell you I'm very animated about getting this fixed.  I'm ok if you ignore it or wait and fix it when there are other more important fixes.  

The workaround is straight forward.  Change the "_" in the example json to something like "underscore", generate the POJOs, and then go fix the annotations for this field.  Not that hard.  I've already done that.

Second, if it had been almost anyone else other than someone big like apache solr generating this, I wouldn't have said anything.  People who use names like this or (my bigger pet peave) use object names to provide some of the information about the object (like use the name of a server as the name of an object instead of having a string named "server" in the object with the name of the server as a value) are IMHO misusing the standard.  It makes parsing much harder as you either have to write a custom deserializer or use a map and lose the safety of having a real pojo with types for the values and a list of getters/setters.  OK, I feel better now.

If you do want to do something, I'd probably have the name of the field as "underscore" or "hyphen" if the name is "-".  But, again, I'm ok if you don't fix it.

Finally, I wanted to say this.  I have to generate parsers for some company's json on a regular basis.  Being able to take their example json (nobody gives me schemas), run it through your code and generate pojos has saved me a lot of time.  So, thank you for doing this program.

joelittlejohn:
Thanks @Smitty010, it's nice to get this kind of insight into how you're thinking about this change.

I think if the string contains no valid characters we should use the unicode name for the chars within it. This could produce a long name, but it's an edge-case so as long as we have _some_ way to support this I think that's okay. We can prefix field name with something to indicate that it's a special translation (e.g. with `__` or `$`).

In this case, your underscore  field would have been translated to a field called:

[...]

PR #1467, opened 2023-01-31
Fix: fixed underscore management to JSON schema

simo385:
Hi everybody,

We are working with the UBL Invoices JSON schema (http://docs.oasis-open.org/ubl/UBL-2.1-JSON/). 

During class generation using the _jsonschema2pojo-maven-plugin_, we get the following exception: _"Exception in thread "main" java.lang.StringIndexOutOfBoundsException: String index out of range: 0" (https://github.com/joelittlejohn/jsonschema2pojo/issues/632)

After code investigation, we found out the character "_" was identified as illegal, then it has removed from the name.

We cannot modify the UBL schema, because we need to build a maintainable application. For this reason, we propose substituting "_" with "underscore" in this fix to avoid a library broke down.

Thanks to everybody.
Simone

### DIFF
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/util/NameHelper.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/util/NameHelper.java
@@ -61,6 +61,9 @@ public class NameHelper {
     public String capitalizeTrailingWords(String name) {
         char[] wordDelimiters = generationConfig.getPropertyWordDelimiters();
 
+        if(name.equals("_")){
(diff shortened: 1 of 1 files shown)


joelittlejohn:
I think best if we fix this by replacing `"_"` with `"__"` (two underscores). It's a valid name and less intrusive.

joelittlejohn:
I have changes (and a test) that I can push.

simo385:
Dear @joelittlejohn,

Ok. Did u schedule the new release? When?

Thanks,
Simone

PR #1474, opened 2023-02-02
Let AnnotationHelper respect targetVersion

XN137:
this should avoid the problem of using the wrong annotation when building on jdk 11 but targeting java 8


XN137:
(in case you consider squash merging this, please use the my same email as on the commit, thanks)

eirnym:
Do I need to attach any additional libraries if I generate on Java 17 targeting for Java 8?

joelittlejohn:
Superseded by #1482.

PR #1475, opened 2023-02-04
Add Email bean validation annotation for "format":"email"

joelittlejohn:
Things to consider if are using the includeJsr303Annotations feature:

* If you have useJakartaValidation turned on, you'll now see `jakarta.validation.constraints.Email` added to a property when using `"format":"email"`.
* If don't have useJakartaValidation turned on, you'll now see [`javax.validation.constraints.Email`](https://javaee.github.io/javaee-spec/javadocs/javax/validation/constraints/Email.html) added to a property when using `"format":"email"`.
  * The Email annotation is introduced in the bean validation spec 2.0, so you will need to upgrade to javax.validation:validation-api 2.x (2.0.1.Final is a good choice). Note, all old annoations are maintained (with the same fully-qualified name) so upgrading this should not break anything.
  * If email validation is still not happening, your validator might only be doing bean validation 1.0 or 1.1 validations. You would then need to upgrade your validator (e.g. with bval, [upgrade to 2.0.5 or later](https://bval.apache.org/downloads.html)).

### DIFF
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/PropertyRule.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/PropertyRule.java
@@ -16,7 +16,6 @@
 
 package org.jsonschema2pojo.rules;
 
-import org.apache.commons.lang.StringUtils;
 import org.jsonschema2pojo.GenerationConfig;
 import org.jsonschema2pojo.JsonPointerUtils;
 import org.jsonschema2pojo.Schema;
@@ -31,6 +30,7 @@ import com.sun.codemodel.JMethod;
 import com.sun.codemodel.JMod;
 import com.sun.codemodel.JType;
 import com.sun.codemodel.JVar;
+import org.apache.commons.lang.StringUtils;
 
 /**
  * Applies the schema rules that represent a property definition.
@@ -201,6 +201,12 @@ public class PropertyRule implements Rule<JDefinedClass, JDefinedClass> {
             ruleFactory.getAnnotator().dateField(field, clazz, node);
         } else if ("time".equalsIgnoreCase(format)) {
             ruleFactory.getAnnotator().timeField(field, clazz, node);
+        } else if ("email".equalsIgnoreCase(format) && ruleFactory.getGenerationConfig().isIncludeJsr303Annotations()) {
+            if (ruleFactory.getGenerationConfig().isUseJakartaValidation()) {
+                field.annotate(jakarta.validation.constraints.Email.class);
+            } else {
+                field.annotate(javax.validation.constraints.Email.class);
+            }
         }
     }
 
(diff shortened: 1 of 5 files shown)


joelittlejohn:
If you receive compilation errors because javax.validation.constraints.Email cannot be found, upgrade to javax.validation/validation-api 2.0.1.Final.

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/PropertyRule.java
```diff
@@ -201,6 +202,12 @@ private void formatAnnotation(JFieldVar field, JDefinedClass clazz, JsonNode nod
             ruleFactory.getAnnotator().dateField(field, clazz, node);
         } else if ("time".equalsIgnoreCase(format)) {
             ruleFactory.getAnnotator().timeField(field, clazz, node);
+        } else if ("email".equalsIgnoreCase(format) && ruleFactory.getGenerationConfig().isIncludeJsr303Annotations()) {
```
unkish, line 205: Would be nice to have some tests as well

unkish (changes requested):

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/PropertyRule.java
```diff
@@ -16,6 +16,7 @@
 
 package org.jsonschema2pojo.rules;
 
+import jakarta.validation.constraints.Email;
```
unkish, line 19: Nitpick: currently this is unused import, since fully-qualified class name is used on L207.

jsonschema2pojo-integration-tests/src/test/java/org/jsonschema2pojo/integration/config/IncludeJsr303AnnotationsIT.java
```diff
@@ -182,6 +182,23 @@ public void jsr303SizeValidationIsAddedForSchemaRuleMinItemsAndMaxItems() throws
 
     }
 
+    @Test
+    public void jsr303EmailValidationIsAddedForFormatEmailSchemaRule() throws ClassNotFoundException {
```
unkish, line 186: A non-related "issue": would it make sense to fix typo by renaming `jar303AnnotationsValidatedForAdditionalProperties` to `jsr303AnnotationsValidatedForAdditionalProperties` ?

unkish (commented):
Added few comments

unkish:
> **If you receive compilation errors because javax.validation.constraints.Email cannot be found, upgrade to javax.validation/validation-api 2.0.1.Final.**

Note that given annotation can't be used when `targetVersion` is less that `1.8` as it has:

- `@Retention(RUNTIME)`
- `@Target` contains `TYPE_USE` which is available since `1.8`

Those who target older version might need to make sure that version of `javax.validation/validation-api` being used is compatible with `targetVersion`

unkish:
Nice work!

Issue #1473, opened 2023-02-02
Annotation `javax.annotation.processing.Generated` is used when building with jdk11 despite `targetVersion=8`


XN137:
we are building on jdk11 (but target a java 8 runtime) with the maven plugin and setting `<targetVersion>8</targetVersion>`

but getting the following compilation errors:
```
error: cannot find symbol
    symbol:   class Generated
    location: package javax.annotation.processing
```

so we have to add `jsonschema2pojo-jdk-annotation` to the classpath to workaround the problem.

however: when targetting java 8 it should not try to add the java9 annotation here imo:
https://github.com/joelittlejohn/jsonschema2pojo/blob/284c1e81556eaa0bd4515a186f25fc875384eef9/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/util/AnnotationHelper.java#L42-L46

i am guessing `Class.forName` works for the java9 classname because we are building on jdk11.

a more general solution that would also cover https://github.com/joelittlejohn/jsonschema2pojo/issues/1461 is to make the annotation class freely configurable with another mojo property like `generatedAnnotationClass`

it could default to `"auto"` and keep running the existing logic if desired

joelittlejohn:
One issue with this having this as a configurable property is that most people don't care about this annotation, they just want something that works without receiving an error on compilation the first time they use this jsonschema2pojo. That said, there are a lot of benefits to having a Generated annotation - I removed it for a while because of these issues, but found that there's quite a lot of tooling that uses that annotation to decide how to treat our generated source code, so I added it back.

So assuming we keep it, it would be good to have both an unobtrusive and sensible default, as well as a config option for those that care to use it.

So I guess the logic should be:

1. `generationAnnotationClass` is specified: use that
2. `generationAnnotationClass` is not specified:
   * targetVersion = 8 or lower, use `javax.annotation.Generated`
   * targetVersion = 9 or higher, use `jakarta.annotation.Generated` is it is found, or if not then use `javax.annotation.processing.Generated` if it is found, or if not then use `javax.annotation.Generated` if it is found.

I think I'll add `jakarta.annotation.Generated` to jsonschema2pojo-jdk-annotation as well, as it can make life easier.

It's so frustrating that something so basic is not part of the Java's core.

XN137:
@joelittlejohn thanks for the quick reply, i've crafted a small PR that would fix the intermediate problem:
[...]

PR #1478, opened 2023-02-08
Choose default targetVersion using Gradle/Maven configuration

joelittlejohn:
Relates to #1473.


jsonschema2pojo-maven-plugin/src/main/java/org/jsonschema2pojo/maven/Jsonschema2PojoMojo.java
```diff
@@ -1114,6 +1117,31 @@ public boolean isIncludeAdditionalProperties() {
     @Override
     public boolean isIncludeSetters() { return includeSetters; }
 
+    private void setTargetVersion() {
+        if (isNotBlank(this.targetVersion)) {
+            return;
+        }
+
+        if (project.getProperties() != null && project.getProperties().containsKey("maven.compiler.target")) {
```
unkish, line 1125: Wouldn't that be provided by `defaultValue = "${maven.compiler.target}"` ?

joelittlejohn: I thought best to centralise this logic in one place, so we can easily see what the order of precedence is and we are not relying on Maven machinery.

joelittlejohn: I had originally implemented this inside getTargetVersion, but since this becomes quire a complex process I think best to trigger it once. I realise now that I replaced `return` with `this.targetVersion =` so I need to adjust this to ensure this process stops when it should.

jsonschema2pojo-gradle-plugin/src/main/groovy/org/jsonschema2pojo/gradle/GenerateJsonSchemaAndroidTask.groovy
```diff
@@ -57,4 +59,12 @@ class GenerateJsonSchemaAndroidTask extends SourceTask {
 
     Jsonschema2Pojo.generate(configuration, new GradleRuleLogger(logger))
   }
+
+  void setTargetVersion(JsonSchemaExtension configuration) {
+    if (!configuration.targetVersion) {
+      def compileJavaTask = project.getTasksByName("compileJava", false).first()
+      configuration.targetVersion = compileJavaTask.getProperties().get("sourceCompatibility")
```
XN137, line 66: should this be `targetCompatibility` ?
seems to disagree with the log output below

same problem in other file

jsonschema2pojo-maven-plugin/src/main/java/org/jsonschema2pojo/maven/Jsonschema2PojoMojo.java
```diff
@@ -1114,6 +1118,40 @@ public boolean isIncludeAdditionalProperties() {
     @Override
     public boolean isIncludeSetters() { return includeSetters; }
 
+    private void setTargetVersion() {
+        if (isNotBlank(this.targetVersion)) {
+            return;
+        }
+
+        if (project.getProperties() != null && project.getProperties().containsKey("maven.compiler.source")) {
+            this.targetVersion = project.getProperties().get("maven.compiler.source").toString();
```
XN137, line 1127: could it make sense to log how the value was determined?

jsonschema2pojo-maven-plugin/src/main/java/org/jsonschema2pojo/maven/Jsonschema2PojoMojo.java
```diff
@@ -1114,6 +1118,40 @@ public boolean isIncludeAdditionalProperties() {
...
+        }
+
+        if (project.getProperties() != null && project.getProperties().containsKey("maven.compiler.release")) {
+            this.targetVersion = project.getProperties().get("maven.compiler.release").toString();
+            return;
+        }
+
+        for (Plugin p : (List<Plugin>) project.getBuildPlugins()) {
+            if (p.getKey().equals("org.apache.maven.plugins:maven-compiler-plugin") && p.getConfiguration() instanceof Xpp3Dom) {
+                final Xpp3Dom compilerSourceConfig = ((Xpp3Dom) p.getConfiguration()).getChild("source");
```
XN137, line 1138: should we prefer the `target` property first?
i.e. same as in the property `defaultValue`

jsonschema2pojo-gradle-plugin/src/main/groovy/org/jsonschema2pojo/gradle/GenerateJsonSchemaAndroidTask.groovy
joelittlejohn: Ah yes, well spotted, thanks. I will update the log message. I changed my mind about which property was more appropriate, but didn't update the log message.

jsonschema2pojo-maven-plugin/src/main/java/org/jsonschema2pojo/maven/Jsonschema2PojoMojo.java
joelittlejohn: and yep, I need to update `defaultValue = "${maven.compiler.target}"`. I had missed this after bringing in the recent changes to move to the `Parameter` annotation. Thanks for mentioning!

XN137: i'll trust you to make the right choice between target or source, if we are generating code then yeah source seems better

my point was only that it seemed inconsistent to default to `target` but not use `target` from `maven-compiler-plugin` here as well :sweat_smile:
if the checking order is "source, release" in both then that seems fine :+1:

jsonschema2pojo-gradle-plugin/src/main/groovy/org/jsonschema2pojo/gradle/GenerateJsonSchemaAndroidTask.groovy
```diff
@@ -57,4 +59,12 @@ class GenerateJsonSchemaAndroidTask extends SourceTask {
...
+      logger.warn 'Using Gradle targetCompatibility as targetVersion for jsonschema2pojo: ' + configuration.targetVersion
```
XN137, line 67: @joelittlejohn you have now merged this with the inconsistent log message, it should mention `sourceCompatibility` here, no?

same problem in `GenerateJsonSchemaJavaTask.groovy`

joelittlejohn: Oops, thanks I will fix

jsonschema2pojo-gradle-plugin/src/main/groovy/org/jsonschema2pojo/gradle/GenerateJsonSchemaAndroidTask.groovy
```diff
@@ -57,4 +59,12 @@ class GenerateJsonSchemaAndroidTask extends SourceTask {
 
     Jsonschema2Pojo.generate(configuration, new GradleRuleLogger(logger))
   }
+
+  void setTargetVersion(JsonSchemaExtension configuration) {
+    if (!configuration.targetVersion) {
+      def compileJavaTask = project.getTasksByName("compileJava", false).first()
```
unkish, line 65: Could this be the cause for [actions failure](https://github.com/joelittlejohn/jsonschema2pojo/actions/runs/4206974193/jobs/7301154742)

>  Execution failed for task ':app:generateJsonSchema2PojoForDebug'.
> Cannot access first() element from an empty Iterable

Neither gradle nor android expert, but could it be that something similar to
```groovy
  void setTargetVersion(JsonSchemaExtension configuration) {
    if (!configuration.targetVersion) {
      configuration.targetVersion = project.android.compileOptions.sourceCompatibility
      logger.info 'Using Gradle sourceCompatibility as targetVersion for jsonschema2pojo: ' + configuration.targetVersion
    }
```
should be used for android (see [Android Gradle plugin API reference: CompileOptions](https://developer.android.com/reference/tools/gradle-api/7.4/com/android/build/api/dsl/CompileOptions) )?

joelittlejohn: Yes, clearly the way the detection is working here is not right.

I have replicated the problem locally and I guess a combination of using a different property, and some more defensive coding, would be best.

joelittlejohn: I don't have time to look this evening but can hopefully fix tomorrow (and release 1.2.1).

joelittlejohn: We also need to get these CI jobs running against HEAD, not the latest release! :)

unkish: > I don't have time to look this evening but can hopefully fix tomorrow (and release 1.2.1).

Whilst error is certainly not a pleasant one, there might be several workarounds (maybe not the nicest ones) eg.: provide target version in configuration, create/register a dummy `compileJava` task, appy `java` plugin.

I don't know whether it is a critical/big pain point for android developers or not, if not - perhaps it would make sense to take some time and not rush ?

PR #1481, opened 2023-02-10
Add Javadoc to constructors parameters for properties that have title/description/$comment

### LINKED ISSUE #1463: Javadoc for constructors missing "description" on parameters?
I'm using the jsonschema2dataclass gradle plugin (issue clone from https://github.com/jsonschema2dataclass/js2d-gradle/issues/604) but I think this issue is coming from the core library here.

for data like:

```
...
                "hash": {
                    "description": "Specifies the hash algorithm and value for the content",
                    "type": "object",
                    "properties": {
                        "algorithm": {
                            "description": "The hashing function used to compute the hash value",
                            "type": "string",
                            "enum": [ "sha256" ]
                        },
                        "value": {
                            "description": "The hash value for the content",
                            "type": "string"
                        }
                    },
                    "required": [ "algorithm", "value" ]
                }
...
```
a constructor is created like:
```
     /**
      *
      * @param value
      * @param algorithm
      */
     public Hash(Hash.Algorithm algorithm, String value) {
         super();
         this.algorithm = algorithm;
         this.value = value;
     }
```

Would it make sense to inject the description into the generated javadoc here so we don't get javadoc generation warnings (happens with Temurin Java11) like:
```
/home/.../Hash.java:44: warning: no description for @param
     * @param value
       ^
```

I was hoping the javadoc could include descriptions from the json spec itself and be generated something like:
```
     /**
      *
      * @param value      the hash value for the content
      * @param algorithm  the hashing function used to compute the hash value
      */
     public Hash(Hash.Algorithm algorithm, String value) {
         super();
         this.algorithm = algorithm;
         this.value = value;
     }
```

unkish:
Add Javadoc to constructors parameters which have corresponding documented properties.
Properties that are not documented (do not have `title` or `description`) won't be added to constructors Javadoc.

Following format will be used:
```java
/**
 * {@param {title. }?{description.}?}?
 */
```

Closes #1463

### DIFF
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/ConstructorRule.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/ConstructorRule.java
@@ -21,13 +21,15 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
-import java.util.LinkedHashSet;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Objects;
 import java.util.Set;
+import java.util.StringJoiner;
 
+import org.apache.commons.lang3.StringUtils;
 import org.jsonschema2pojo.GenerationConfig;
 import org.jsonschema2pojo.Schema;
 import org.jsonschema2pojo.util.NameHelper;
@@ -69,10 +71,10 @@ public class ConstructorRule implements Rule<JDefinedClass, JDefinedClass> {
     return instanceClass;
   }
 
-  private JDefinedClass handleLegacyConfiguration(JsonNode node, JDefinedClass instanceClass, Schema currentSchema) {
+  private void handleLegacyConfiguration(JsonNode node, JDefinedClass instanceClass, Schema currentSchema) {
     // Determine which properties belong to that class (or its superType/parent)
-    LinkedHashSet<String> requiredClassProperties = getConstructorProperties(node, true);
-    LinkedHashSet<String> requiredCombinedSuperProperties = getSuperTypeConstructorPropertiesRecursive(node, currentSchema, true);
+    Map<String, String> requiredClassProperties = getConstructorProperties(node, true);
+    Map<String, String> requiredCombinedSuperProperties = getSuperTypeConstructorPropertiesRecursive(node, currentSchema, true);
 
     // Only generate the constructors if there are actually properties to put in them
     if (!requiredClassProperties.isEmpty() || !requiredCombinedSuperProperties.isEmpty()) {
@@ -84,9 +86,6 @@ public class ConstructorRule implements Rule<JDefinedClass, JDefinedClass> {
       // Generate the actual constructor taking in only the required properties
       addFieldsConstructor(instanceClass, requiredClassProperties, requiredCombinedSuperProperties);
     }
-
-    // Return the original class we modified
-    return instanceClass;
   }
 
   private void handleMultiChoiceConstructorConfiguration(JsonNode node, JDefinedClass instanceClass, Schema currentSchema) {
@@ -94,10 +93,10 @@ public class ConstructorRule implements Rule<JDefinedClass, JDefinedClass> {
     boolean requiresConstructors = false;
 
     // Use these lists to keep track of the properties on the class, but we'll only populate them if we need to
-    LinkedHashSet<String> requiredClassProperties = null;
-    LinkedHashSet<String> classProperties = null;
-    LinkedHashSet<String> requiredCombinedSuperProperties = null;
-    LinkedHashSet<String> combinedSuperProperties = null;
+    Map<String, String> requiredClassProperties = null;
+    Map<String, String> classProperties = null;
+    Map<String, String> requiredCombinedSuperProperties = null;
+    Map<String, String> combinedSuperProperties = null;
 
     GenerationConfig generationConfig = ruleFactory.getGenerationConfig();
     boolean includeCopyConstructor = generationConfig.isIncludeCopyConstructor();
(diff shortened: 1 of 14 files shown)


joelittlejohn:
I know I recommended this, but I think with this change we will see a new javadoc warning:

```
warning: no @param for somearg
```

For any argument that doesn't have `@param`.

So we get a warning for no param, and we get a warning for param without description. So we're a bit stuck if we don't have a description :thinking: Making up some junk (like `"The somearg value"`) seems pointless too.

unkish:
I've been thinking about it and couldn't come up with any "ideal" solutions, so far following "ideas" where on the table:
- accept the fact that we can't do much and add junk description eg. `@param somearg somearg`
    - Drawback: junk in Javadoc comments
- having more complicated wording there `"The somearg value"`
    - Drawback: wouldn't improve much over previous solution whilst also might be not the best solution for projects that might be using non-english descriptions ?
- have a configuration property for missing property descriptions
    - Drawback: one more config property to support that doesn't add much value?
- have a configuration property to adding of disable Javadoc params
    - Drawback: one more config property to support
- try to complicate logic a bit - apply "all or nothing" strategy: eg. check if number of Javadoc descriptions for constructor matches the number of arguments - add Javadoc, else skip it entirely
    - Drawback: might not be so obvious what is going on for users, why in certain cases Javadoc is generated and in the other cases is not
- Some sort of combination of above

joelittlejohn:
On reflection I think the way you have implemented this is correct. We should omit any param that has no description.

If you see a warning telling you that you have not documented some parameters, then this warnings are accurate, and you can fix it by documenting the parameter (add title and/or description).

unkish:
For a moment I've thought that people might start complaining about schema being an external to their project and them being unable/not allowed to modify it.
But change wouldn't make it worse (granted that warning and not error will be emitted) as currently we produce Javadoc without `@param` description that also generates warning.

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/ConstructorRule.java
```diff
@@ -200,38 +198,50 @@ private LinkedHashSet<String> getConstructorProperties(JsonNode node, boolean on
...
+    private String getJavadocForProperty(JsonNode propertyObj) {
+        final boolean hasDescription = propertyObj.has("description") && !propertyObj.get("description").isNull();
+        final String title;
+        if (propertyObj.has("title") && !propertyObj.get("title").isNull()) {
+            title = StringUtils.appendIfMissing(propertyObj.get("title").asText(), ".") + " ";
+        } else {
+            title = "";
+        }
+        final String description = hasDescription ? StringUtils.appendIfMissing(propertyObj.get("description").asText(), ".") : "";
+        return title + description;
```
joelittlejohn, line 228: Sorry to be pedantic, but if we have title only and no description, then we should not have a space after the `.`.

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/ConstructorRule.java
```diff
@@ -200,38 +198,50 @@ private LinkedHashSet<String> getConstructorProperties(JsonNode node, boolean on
...
       } else {
-        rtn.add(nameHelper.getPropertyName(property.getKey(), property.getValue()));
+        rtn.put(nameHelper.getPropertyName(property.getKey(), property.getValue()), javadoc);
       }
     }
     return rtn;
   }
 
+    private String getJavadocForProperty(JsonNode propertyObj) {
+        final boolean hasDescription = propertyObj.has("description") && !propertyObj.get("description").isNull();
```
joelittlejohn, line 220: I'd use `&& isNotBlank(propertyObj.get("description"))` here.

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/ConstructorRule.java
```diff
@@ -301,14 +311,13 @@ private void generateFieldsConcreteBuilderConstructor(JDefinedClass baseBuilderC
...
-  private JMethod generateCopyConstructor(JDefinedClass jclass, Set<String> classProperties, Set<String> combinedSuperProperties) {
+  private JMethod generateCopyConstructor(JDefinedClass jclass, Map<String, String> classProperties, Map<String, String> combinedSuperProperties) {
 
     // Create the JMethod for the copy constructor
     JMethod copyConstructorResult = jclass.constructor(JMod.PUBLIC);
 
     // Add a parameter to the copyConstructor for the actual object being copied
-    copyConstructorResult.javadoc()
-        .addParam("source");
+    copyConstructorResult.javadoc().addParam("source").append("an actual object being copied");
```
joelittlejohn, line 320: s/an actual/the/

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/ConstructorRule.java
unkish: Same comment would be applicable for `propertyObj.get("title").isNull()` I would assume ?

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/ConstructorRule.java
unkish: Replaced with `StringJoiner`

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/ConstructorRule.java
unkish: Replaced

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/ConstructorRule.java
unkish: Changed, thank You for noticing

joelittlejohn:
@unkish This is the last one for 1.2.0 I think. Could you rebase and I will merge?

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/ConstructorRule.java
joelittlejohn: I didn't know about StringJoiner, nice!

Issue #1255, opened 2021-05-03
Unable to generate sources for multiple version of java


sjmittal:
I am using the `jsonschema2pojo-maven-plugin` to generate java sources via maven.
I have java version 11 installed on my PC but I need to generate the sources for both java8 and java11 versions.
I have tried configuring the `targetVersion` attribute in the plugin but it still seems to be generating the source for java 11. Basically it is picking up the java version of the installed JRE and not the `targetVersion` property.

So as next step when I try to compile the sources as java8 it fails.
The reason is because then import statements are generated as 
`import javax.annotation.processing.Generated;` and not
`import javax.annotation.Generated;`

Any idea if this is an issue or how this can be worked around.

joelittlejohn:
I think we can fix this, but I'm interested to know if you are able to try adding this dependency to your project:

https://search.maven.org/artifact/org.jsonschema2pojo/jsonschema2pojo-jdk-annotation/0.0.1/jar

In theory it includes only the annotation that you need, and is intended to be used when javax.annotation.processing.Generated is not available.

joelittlejohn:
Closed by #1482.

PR #1482, opened 2023-02-14
Use targetVersion to decide which Generated annotation to add

### LINKED ISSUE #1361: @Generated annotation generator ignores target Java version
When use jsonschema2pojo generator with **runtime** java version 9+ it's impossible to generate classes for older versions of java, even target version is set to 1.7 or 1.8. 

It's quite unfortunate for our project (which uses java 11+) to use detached module only to publish models compatible with java 1.8

Both maven and gradle generators are affected.

### LINKED ISSUE #1473: Annotation `javax.annotation.processing.Generated` is used when building with jdk11 despite `targetVersion=8`

joelittlejohn:
Closes #1474, closes #1473, closes #1361

### DIFF
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/DefaultGenerationConfig.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/DefaultGenerationConfig.java
@@ -25,6 +25,7 @@ import java.util.Iterator;
 import java.util.Map;
 
 import org.jsonschema2pojo.rules.RuleFactory;
+import org.jsonschema2pojo.util.JavaVersion;
 
 /**
  * A generation config that returns default values for all behavioural options.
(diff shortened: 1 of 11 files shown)


unkish:
Would it possibly close https://github.com/joelittlejohn/jsonschema2pojo/issues/1255 as well ?

PR #1483, opened 2023-02-15
Add useful debug logging when reading schemas and creating classes

### LINKED ISSUE #863: Limited information in Maven logs
Hello, when using this plugin I noticed that the information in Maven log is very limited:

> [INFO] 
> [INFO] --- jsonschema2pojo-maven-plugin:0.5.1:generate (default) @ SportDeerClient ---
> [INFO] 

Is it possible to specify somehow that the plugin would write additional information like what POJOs were generated and so on? This would be handy when using external building with CI servers.

joelittlejohn:
Closes #863

### DIFF
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/SchemaStore.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/SchemaStore.java
@@ -31,13 +31,16 @@ public class SchemaStore {
 
     protected final FragmentResolver fragmentResolver = new FragmentResolver();
     protected final ContentResolver contentResolver;
+    protected final RuleLogger logger;
 
     public SchemaStore() {
         this.contentResolver = new ContentResolver();
+        this.logger = new NoopRuleLogger();
     }
 
-    public SchemaStore(ContentResolver contentResolver) {
+    public SchemaStore(ContentResolver contentResolver, RuleLogger logger) {
         this.contentResolver = contentResolver;
+        this.logger = logger;
     }
 
     /**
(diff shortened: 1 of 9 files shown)


jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/SchemaStore.java
```diff
@@ -18,26 +18,28 @@
 
 import static org.apache.commons.lang3.StringUtils.*;
 
+import com.fasterxml.jackson.databind.JsonNode;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.HashMap;
 import java.util.Map;
 
-import com.fasterxml.jackson.databind.JsonNode;
```
unkish, line 26: Out of curiosity: did some setting change or it is due to switch from Eclipse to IDEA that order of import rows is sometimes "jumping".
Maybe it would be possible to share idea Code Style (Settings -> Editor -> Code Style -> Java | Json) scheme to be used similarly to [jsonschema2pojo-eclipse-formatter.xml](https://github.com/joelittlejohn/jsonschema2pojo/blob/master/jsonschema2pojo-eclipse-formatter.xml) and [jsonschema2pojo-eclipse-preferences.epf](https://github.com/joelittlejohn/jsonschema2pojo/blob/master/jsonschema2pojo-eclipse-preferences.epf) ?

joelittlejohn: Yes, I'm still getting used to the options in IDEA. I think I've set things up now so that it will be the same as Eclipse.

I'm a reluctant IDEA user tbh :smile:

I've added a jsonschema2pojo-intellij-formatter.xml now too.

joelittlejohn:
Note, in Maven if you want to enable debug logging for this plugin only, add a system property to configure the SimpleLogger, like this:

```
mvn -Dorg.slf4j.simpleLogger.log.org.jsonschema2pojo=debug test
```

This is useful for CI jobs where you want to see some output written to the log for debugging if the build fails.

PR #1484, opened 2023-02-15
Clear schema cache between JSON files

### LINKED ISSUE #1427: "Path not present" error when processing multiple JSON files having objects with the same name

joelittlejohn:
When using JSON or YAML files (not schemas, but example data) the schemas will not have a proper file path in their ID, as there is no actual schema file. All schema IDs will be fragments only (relative references within the document). Once generation is complete for a file, these schemas should be thrown away and not reused.

Closes #1427


jsonschema2pojo-integration-tests/src/test/java/org/jsonschema2pojo/integration/json/JsonTypesIT.java
```diff
@@ -210,4 +210,29 @@ public void arrayAtRootProducesNoJavaTypes() throws Exception {
 
     }
 
+    @Test
+    @SuppressWarnings("rawtypes")
```
unkish, line 214: Nitpick: it looks like `@SuppressWarnings("rawtypes")` is redundant here

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/Jsonschema2Pojo.java
```diff
@@ -127,6 +127,10 @@ private static void generateRecursive(GenerationConfig config, SchemaMapper mapp
 
         for (File child : schemaFiles) {
             if (child.isFile()) {
+                if (config.getSourceType() == SourceType.JSON || config.getSourceType() == SourceType.YAML) {
+                    // any cached schemas will have ids that are fragments, relative to the previous document (and shouldn't be reused)
+                    mapper.getRuleFactory().getSchemaStore().clearCache();
```
unkish, line 132: Before implementing https://github.com/joelittlejohn/jsonschema2pojo/pull/1464, I have also thought about clearing cache but for some reason it felt somewhat weird to have source type checks.
Now that there is a nice comment it does not feel so "alien" anymore 👍

unkish:
Would it make sense to take some of the changes from https://github.com/joelittlejohn/jsonschema2pojo/pull/1464 in particular:

- remove `YamlRefIT` as it is empty (dead code)
- rewrite `SelftRefIT::nestedSelfRefsInStringContentWithoutParentFile` such that it would act as an IT rather than unit test (`schemaRule.generateAndCompile(...)` vs. `new SchemaMapper().generate(...)`)
- add `|| ruleFactory.getGenerationConfig().getSourceType() == SourceType.YAML` to https://github.com/joelittlejohn/jsonschema2pojo/blob/65e1fdbcf4a4fd7298ea7ede0dc9c3e1651e9035/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/SchemaMapper.java#L127

Or separate PR would be better ?

jsonschema2pojo-integration-tests/src/test/java/org/jsonschema2pojo/integration/json/JsonTypesIT.java
```diff
@@ -210,4 +210,29 @@ public void arrayAtRootProducesNoJavaTypes() throws Exception {
...
+    public void propertiesWithSameNameOnDifferentObjects() throws Exception {
```
unkish, line 215: Would it make sense to add similar test also to `PlainYamlTypesIT` ?

jsonschema2pojo-integration-tests/src/test/java/org/jsonschema2pojo/integration/json/JsonTypesIT.java
```diff
@@ -210,4 +210,29 @@ public void arrayAtRootProducesNoJavaTypes() throws Exception {
...
+
+        ClassLoader resultsClassLoader = schemaRule.generateAndCompile("/json/propertiesSameName", "com.example",
+                config("sourceType", "json"));
```
unkish, line 218: Question: was the line split here on purpose eg. there's a "rule" to try keep line length under some threshold ?
For example L161, L225-226 are pretty long also

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/Jsonschema2Pojo.java
joelittlejohn: Yes, I must admit, it feels like a bit of a hack.

I implemented a couple of alternatives before landing on the solution you see in this PR. First I thought lets put the JSON filename into the ID, however this means that we will try to resolve content from it (and it led me down the same path as you). Next I thought let's not cache the schemas at all in this case (why bother?) but then found that this breaks recursive references since you get into an infinite loop of resolving schemas if you don't remember (and reuse) what you have resolved in the past (yay for extensive integration tests!).

After these two failures, I decided that if we are building a cache of schemas that must refer only to one document, because all references are root-relative, then just clear them out between files. Hopefully the comment stops the reasoning being completely lost to the mists of time.

It could be better to 'qualify' the schemas in the cache in a better way, or to unify everything as you did in your PR, however I like how little additional change and conceptual complexity my change brings in, so it felt like the best of all available solutions.

jsonschema2pojo-integration-tests/src/test/java/org/jsonschema2pojo/integration/json/JsonTypesIT.java
joelittlejohn: I think I will merge these two files into one parameterised test.

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/Jsonschema2Pojo.java
unkish: Given approach indeed is far simpler than unifying one.
Even though latter would have made `json` and `yaml` source types "fisrt class citizens", clearing cache makes sense as well.
Thank you for providing "insights" 👍

jsonschema2pojo-integration-tests/src/test/java/org/jsonschema2pojo/integration/json/JsonTypesIT.java
joelittlejohn: I generally prefer long lines to splitting (we have lots of horizontal space now). I think this came from me copying code from elsewhere. Sometimes I split for readability. I have removed some more newlines where I think they did not improve readability.

joelittlejohn:
- [x] remove YamlRefIT as it is empty (dead code)
- [ ] rewrite SelftRefIT::nestedSelfRefsInStringContentWithoutParentFile such that it would act as an IT rather than unit test (schemaRule.generateAndCompile(...) vs. new SchemaMapper().generate(...))
- [x] add || ruleFactory.getGenerationConfig().getSourceType() == SourceType.YAML

I've ticked off two of the above as part of this change.

PR #1485, opened 2023-02-15
Avoid output to System.err when finding unique Enum name

joelittlejohn:


unkish (approved):
New logged out message makes more sense now also (sample from `EnumIT::enumWithNoValidCharactersForIdentifier` output):
```
Enum name __EMPTY__ already used; trying to replace it with __EMPTY___
Enum name __EMPTY___ already used; trying to replace it with __EMPTY____
Enum name __EMPTY__ already used; trying to replace it with __EMPTY____
```
vs.
```
[warn] Enum name __EMPTY__ already used; trying to replace it with __EMPTY___
[warn] Enum name __EMPTY__ already used; trying to replace it with __EMPTY___
[warn] Enum name __EMPTY___ already used; trying to replace it with __EMPTY____
```

PR #1487, opened 2023-02-18
Fix detection of targetVersion for Android gradle builds

joelittlejohn:
This fixes the broken targetVersion detection for Gradle Android builds, and also updates the Android Sample GitHub action so that it actually runs against the HEAD code (rather than the latest release).


unkish (approved):
Worked like a charm locally 👍

Issue #1497, opened 2023-03-05
Support custom Generated annotation that uses runtime retention policy required by JaCoCo


marioja:
The javax.annotation.Generated or javax.annotation.processing.Generated annotation all use SOURCE retention policy which is not compatible with JaCoCo code coverage which works at the byte code level.  We need to have the ability to use a custom generated annotation instead of or in addition to the javax annotation.

marioja:
I am working on a pull request.

joelittlejohn:
I think we can use an existing pull request for this.

What annotation does JaCoCo expect?

marioja:
JaCoCo requires a runtime retention policy annotation.  What existing pull request are you talking about?

marioja:
Have a look at PR #1499 and let me know if you need anything.

marioja:
Hi @joelittlejohn, is there any issue with my PR #1499?  You might have been waiting until I was done and I think I am and just need your review.

marioja:
Could this PR #1499 be merged or do you require something more from me? I would like to use it in our builds and would appreciate if this could be accepted. Thanks

marioja:
@joelittlejohn gentle ping.  Do you think PR #1499 could be merged?

marioja:
@billkoch Well I pinged joelittlejohn, so all we can do is wait for him to respond on the PR.

PR #1499, opened 2023-03-06
Add/custom/generated/annotation

marioja:
This pull request provides support for generating a custom annotation type called JaCoCoGenerated and add this as an annotation to all the classes generated


marioja:
Are there any issues with this pull request?

marioja:
This is referenced in issue #1497.  Could this PR be merged or do you require something  more from me? I would like to use it in our builds and would appreciate if this could be accepted.  Thanks

PR #1515, opened 2023-04-25
Avoid NPE when setting additionalProperties path

MikeEdgar:
A schema's `id` is not guaranteed to be non-null. E.g.

https://github.com/joelittlejohn/jsonschema2pojo/blob/80d827db93e4cc3848c4f297ebbfe4b3fd936272/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/SchemaMapper.java#L89-L91

### DIFF
--- a/jsonschema2pojo-core/src/test/java/org/jsonschema2pojo/rules/AdditionalPropertiesRuleTest.java
+++ b/jsonschema2pojo-core/src/test/java/org/jsonschema2pojo/rules/AdditionalPropertiesRuleTest.java
@@ -1,0 +1,52 @@
+/**
+ * Copyright © 2010-2020 Nokia
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jsonschema2pojo.rules;
+
+import static org.junit.Assert.assertEquals;
+
+import org.jsonschema2pojo.Schema;
+import org.junit.Test;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.sun.codemodel.JClass;
+import com.sun.codemodel.JCodeModel;
+import com.sun.codemodel.JDefinedClass;
+import com.sun.codemodel.JMethod;
+import com.sun.codemodel.JType;
+
+public class AdditionalPropertiesRuleTest {
+
+    @Test
+    public void testSchemaWithNullURI() throws Exception {
+        ObjectMapper mapper = new ObjectMapper();
+        ObjectNode node = (ObjectNode) mapper
+                .readTree("{\"type\":\"integer\",\"maximum\":9}");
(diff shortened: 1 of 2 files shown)


andreaTP:
any update on this fix?

joelittlejohn:
Thanks for pinging @andreaTP. I saw the CI job failed and didn't get around to investigating that. I've kicked this off again and will merge if this passes.

MikeEdgar:
I think the issue with CI may be this: https://github.blog/changelog/2022-08-09-github-actions-the-ubuntu-18-04-actions-runner-image-is-being-deprecated-and-will-be-removed-by-12-1-22/

unkish:
Would it be possible to provide also test cases as outlined in https://github.com/joelittlejohn/jsonschema2pojo/blob/master/CONTRIBUTING.md

MikeEdgar:
Unit test added. Between the new test and the existing integration tests, both null/non-null schema Ids are tested.

joelittlejohn:
@MikeEdgar is it not possible to recreate this case via the integration tests?

MikeEdgar:
@joelittlejohn , the integration tests all seem to be using a particular approach where the schemas are loaded from JSON (which always sets the id) using the Maven plugin Mojo. If that's the preferred location for new tests, it will likely need to be an alternate process.

andreaTP:
Ping @joelittlejohn and @unkish , anything that should be done here to get the fix merged and released?

Thanks in advance 🙏

jsonschema2pojo-core/src/test/java/org/jsonschema2pojo/rules/AdditionalPropertiesRuleTest.java
```diff
@@ -0,0 +1,61 @@
...
+import com.sun.codemodel.JType;
+
+public class AdditionalPropertiesRuleTest {
+
+    static final ObjectMapper mapper = new ObjectMapper();
+    RuleFactory factory;
+    AdditionalPropertiesRule rule;
+
+    @Before
+    public void setUp() throws JClassAlreadyExistsException {
```
unkish, line 41: Nit:

- `static final ObjectMapper mapper = new ObjectMapper();` doesn't need to be `static`
- doesn't look like `throws JClassAlreadyExistsException` is needed
- `rule` could be declared within method ie;
```java
    @Test
    public void testSchemaWithNullURI() throws Exception {
        final AdditionalPropertiesRule rule = new AdditionalPropertiesRule(new RuleFactory());
```

unkish:
> anything that should be done here to get the fix merged and released?

It's up to maintainer (might be dependent on available free time/resources/...)

MikeEdgar:
@joelittlejohn , @unkish - please have another look and let me know if any other changes are necessary.

unkish:
From the change itself perspective - it looks perfectly fine. There is also similar "guard" in `ArrayRule` https://github.com/joelittlejohn/jsonschema2pojo/blob/98c85ce4d4b9333bafae2c18eca5637ea6a7c5da/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/ArrayRule.java#L84

Latter was added with commit https://github.com/joelittlejohn/jsonschema2pojo/commit/8e97ddb03a82bd366250392c3e79450c6420bd6a

\
Additionally verified that NPE would also be thrown via public API "usage" in `SchemaMapper` and current change would fix it:

```java
public class SchemaMapperTest {
    ...

    @Test
    public void generateCreatesTypeFromSchemaWithAdditionalProperties() throws Exception {
        String schema = "{\"type\":\"object\",\"additionalProperties\":{\"type\":\"integer\",\"maximum\":9}}";
        JType result = new SchemaMapper().generate(new JCodeModel(), "ExampleClass", "", schema);
        JType returnType = ((JDefinedClass) result).getMethod("getAdditionalProperties", new JType[0]).type();

        assertEquals("Map<String,Integer>", returnType.name());
    }

}

```

PR #1520, opened 2023-05-10
Switch CI job runner to 'ubuntu-latest'

unkish:
Switch CI job runner to 'ubuntu-latest'

* Use identical runner value across all workflows


unkish:
https://github.com/actions/runner-images/issues/6002

PR #1124, opened 2020-04-12
Features/issue 1123 process schema definitions

dewthefifth:
This pull request contains the implementation of schema definition processing, however it breaks an existing unit test as well as backwards compatibility. Do not merge yet, I'm creating it for suggestions/input.

PR #1523, opened 2023-05-22
Generate objects from subschemas defined in 'definitions'/'$defs' section

### LINKED ISSUE #1123: Please create objects in the definition section to be used
I would have thought the following would work:

```
{
  "$schema": "http://json-schema.org/draft-07/schema",
  "type": "object",
  "definitions": {
    "MessageAttribute": {
      "type": "object",
      "additionalProperties": true,
      "properties": {
        "Type": {
          "type": "string"
        },
        "Value": {
          "type": "string"
        }
      }
    }
  },
  "properties": {
    "MessageAttributes": {
      "existingJavaType": "java.util.Map<String, MessageAttribute>",
      "type": "object",
      "additionalProperties": { "$ref": "#/definitions/MessageAttribute" }
    }
  }
}
```

But the MessageAttribute class is not created so the existingJavaType fails.  I can work around this problem today in an ugly way but it would be nice for the above syntax to work.  The ugly way is to move everything in the definitions section to another file as below:

`message-attribute.json`
```
{
  "$schema": "http://json-schema.org/draft-07/schema",
  "type": "object",
  "additionalProperties": true,
  "properties": {
    "Type": {
      "type": "string"
    },
    "Value": {
      "type": "string"
    }
  }
}
```

`message.json`
```
{
  "$schema": "http://json-schema.org/draft-07/schema",
  "type": "object",
  "properties": {
    "MessageAttributes": {
      "existingJavaType": "java.util.Map<String, MessageAttribute>",
      "type": "object",
      "additionalProperties": { "$ref": "message-attribute.yaml" }
    }
  }
}
```

unkish:
Closes #1123


unkish:
A fairly naïve implementation inspired by https://github.com/joelittlejohn/jsonschema2pojo/pull/1124.
All props and kudos to Duane Zamrok.

Not supported: defaulting to using the dialect of the parent schema when figuring out whether to use/prefer `definitions` over `$defs`.


Feedback/critique is welcomed.

lejtemxviw:
It looks like OAS3 would expect to include the schemas for the spec in `components/schemas`, rather than `definitions` - could we modify this change so that the path to look for was customizable, somehow?  Or maybe leave this PR as is, but create a new ticket/PR to add that as an improvement to this PR?

unkish:
> It looks like OAS3 would expect to include the schemas for the spec in components/schemas, rather than definitions - could we modify this change so that the path to look for was customizable, somehow

I guess one option could be changing `private String getDefinitionsNodeName` to `protected String getDefinitionsNodeName` which would then allow overriding given method in custom `SchemaRule`.

Please note that handing/processing of `oneOf`/`allOf`/`anyOf` nodes is not in the scope of given request (if OAS3 contains these - they'll be skipped)

joelittlejohn:
I guess if we make this configurable, we need to ensure that you can specify a _path_ to the schemas rather than just a property name. It sounds like oas3 as the node in `/components/schemas`. I like the idea of making this configurable as `$defs` was only recently standardised.

unkish:
> I like the idea of making this configurable as $defs was only recently standardised

@joelittlejohn what is your vision on having definitions path being configurable ?
Would it be via 

- configuration property
- extension property
- both ?

Shall that be addressed within given issue ?

joelittlejohn:
For now I was imagining a configuration property, like definitionsPath (default is as per the JSON schema spec). I think having different definitions paths per document would be a pretty confusing way to do things.

Given that the current behaviour has been in place for so long, I also think we need a new config option to turn this on: generateDefinitions, default false.

unkish:
> For now I was imagining a configuration property, like definitionsPath (default is as per the JSON schema spec)

1. Would that be per [current latest JSON schema spec](https://json-schema.org/draft/2020-12/json-schema-core#name-schema-re-use-with-defs) ? Ie. `$defs`
2. Which "format" it would expected to be defined
    1. `$defs`
    2. `/$defs`
    3. `#/$defs`
    4. ...

> I think having different definitions paths per document would be a pretty confusing way to do things.

Whilst I do agree that it might be confusing, on the other hand it might as well be not that uncommon (plugin does not need to support all possible permutations though).

May I assume that it would mean there will be no need for fallback logic as the behavior will be straight-forward:
if _generateDefinitions_ is `true` and `definitionsPath` exists - process it's definitions.

Blu2k19:
@joelittlejohn 
Any news? what can we do to help get this merged?

tinpa-de:
Hey, we would be interested in this feature! 😄 
Can we somehow support to get it done?

ctrimble:
@unkish It looks like this PR has stalled out.  Are you still interested in getting it merged?

unkish:
> @unkish It looks like this PR has stalled out.

@ctrimble by "stalled out" you meant that there were conflicts ?

> Are you still interested in getting it merged?

A little backstory: it's not that I'm personally interested in getting it merged - rather there seemed to be a decent demand for given feature and Duane Zamrok already did a lot of "heavy lifting" so I've picked up his PR (https://github.com/joelittlejohn/jsonschema2pojo/pull/1124) and enhanced it a bit.

If there’s consensus among the author and maintainers/collaborators that this feature is worth including in the upcoming release, please feel free to merge it.
If not - I'm fine with either leaving it open or be closed.

xmourgues:
Hello,
My team derived this plugin to have this PR included for our developments. 
So, on a personal note, I think this would be a good idea to officially support this.

PR #1590, opened 2024-03-02
feat: support declared type as array for jsonSchema

yzaouzaou:
When using Jsr303 annotation, to get our objects fileds checked for constraints, we have to annotate them with `@Valid`, this annotation is added on fileds that aren't premetives (or objects représenting premitives like : Integer) to recursively check constraints. 
Until now, to annotate a field with this annotation, we check for it's type weither it's `"object"` or `"array"`. Since the specification of jsonSchema seems to accept a string or an array of strings, when we use an array of strings the annotation `@Valid` is not added to the field, thus, the field is not checked for constraints.

for example, in the current version, if we use this jsonShema (which is valid according to specification of jsonschema)
```
{
  "type": "object",
  "properties": {
    "myFiled": {
      "type": ["object", "null"],
      "properties": {
        "childprimitivefield": {
          "type": "string",
          "maxLength": 5
        }
      }
    }
  }
}

```
The filed `myField` will not have the annotation `@Valid` and it's objects won't be validated.

## Some links : 

- see [beanvalidation.org](https://beanvalidation.org) for more information about jsr303
- see [json-schema.org](https://json-schema.org/specification) for more information about jsonschema

## PS :

For tests, i have just changed existing tests, please let me know if it's better to duplicate existing tests to cover this new feature even if it's the same as existing one 😃.


eirnym:
where in schema provided you specified that type may be array?

* `root` must be an object (e.g. Map or POJO)
* `myFiled` must be an object (e.g. Map or POJO) or `null`
* `childprimitivefield` must be a `string`

yzaouzaou:
Hi @eirnym 

The issue does not lie with the generated model itself, but rather with the required annotations. The class `PropertyRule` assumes that for a field to be considered as an object or an array, its type in the JSON schema must be specified as "object" or "array." However, the JSON schema specification indicates that the type can be expressed either as a simpleType (i.e., a string with a literal type) or as an array of simple types. (see [jsonschema draft 2020-12 meta/validation](https://[json-schema.org/draft/2020-12/meta/validation](https://json-schema.org/draft/2020-12/meta/validation))). 

```json
...
        "type": {
            "anyOf": [
                { "$ref": "#/$defs/simpleTypes" },
                {
                    "type": "array",
                    "items": { "$ref": "#/$defs/simpleTypes" },
                    "minItems": 1,
                    "uniqueItems": true
                }
            ]
        }
...
        "simpleTypes": {
            "enum": [
                "array",
                "boolean",
                "integer",
                "null",
                "number",
                "object",
                "string"
            ]
        }
```
Therefore, in the example I provided, the type of `myFiled` (apologies for the earlier typo; it was supposed to be `myField`) is expressed as an array of simplTypes ("object" and null value). When `PropertyRule` checks its type and determines if it is an object, the answer is no because `node.path("type").asText()` returns `"[ \"object\", null ]"` and this is not equal to "object". However, according to the specification mentioned above, this should also be considered as an object. As a result, it should be annotated with the `@Valid` annotation, which will allow Jakarta Bean Validation to validate its fields.

I hope my description is clearer. I am available to address any additional comments or feedback.

thank you @eirnym  😄

PR #1592, opened 2024-03-14
Feature/added required nonnull check

jiri-meluzin:
When a field in json schema is required and includeJsr303Annotations then compiler complains about missing checks of nullability of the fields.

I have introduced new config option includeRequireNonNullOnRequiredFields which added required checks so no nullability compiler warnings


jiri-meluzin:
@joelittlejohn could you please review the PR

PR #1598, opened 2024-04-02
Upgrade codemodel from com.sun 2.6 to org.glassfish 4.0.5

juherr:


juherr:
@joelittlejohn What do you think if the support of Java8 will be dropped?

joelittlejohn:
I'm open to that, but it's there a specific benefit to moving to this version? Does 3.x support java 8?

juherr:
Good idea, I will start to check if it is possible.
But notice that lot of dependencies dropped java8 support and it will be difficult to be up to date.

juherr:
I worked with 3.x. To be honest, I expected more changes that could use some new java features but it looks like it is not a very active project.
The changelog is available at https://github.com/eclipse-ee4j/jaxb-ri/releases

For the record, it looks https://github.com/phax/jcodemodel is a bit more active but the latest release is old too and the merge with upstream is compromised (see https://github.com/phax/jcodemodel/issues/73 and https://github.com/eclipse-ee4j/jaxb-ri/issues/1353)

joelittlejohn:
I do wonder if there is something similar to codemodel that has kept up with all the language additions that Java has had recently. I guess probably not as XML has become less popular and so has jaxb.

juherr:
@ctrimble Fyi

ctrimble:
@juherr This is on my radar.

It may have to wait until I get a few  more PRs onto master, since this library change will likely break other changes.  I will get a release plan together.

juherr:
@ctrimble I don't think the current pull request is breaking anything. But right, the next major version of the library is dropping the support of many java versions ...

ctrimble:
I was thinking of existing **open** PRs containing import statements.

ctrimble:
@juherr, I am starting to get a release plan together.  Since we are changing the import of a core library here, this feels like it must come in on at least a minor version bump.

I am going to shoot for this to be in 1.3.0.  I will try to merge it as late as possible in that release, so that other PRs do not get jammed up on the change to import statements.

juherr:
@ctrimble Ok, I see your point, and you’re right, it will deprecate most of the current pull requests.
Feel free to update this pull request with any changes. My only goal was to make sure we’re able to make the move.

juherr:
@ctrimble Could you create a `release/1.3.x` branch so we can set it as the target for related pull requests and start merging them right away?

ctrimble:
@juherr my plan was to get 1.2.3 ready, get a release done, then start on the 1.3.0 items.  I have been working with Joe to make sure he is on board with cutting that release.  I don't think it will take long.

I don't think that Joe is interested in making the branches super complex.  So lets just keep things on master and chip through stuff.

joelittlejohn:
Closing this one for now. I've done significant work with the phax library now (including adding some major new features there) but it's not the right way forward for us due to it's lack of compatibility. We'll likely move over to eclipse-ee4j/jaxb-ri eventually.

PR #1614, opened 2024-05-17
Fixing problems with remote URL when using Maven Plugin

gr0l:
When using remote URL's like an HTTP-Url as sourcePaths input. The URL's have been normalized to a file like notation. Afterwards the type generator was unable to resolve the remote file.

### DIFF
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/util/URLUtil.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/util/URLUtil.java
@@ -21,11 +21,27 @@ import java.net.MalformedURLException;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
+import java.util.Arrays;
+import java.util.Set;
+import java.util.stream.Collectors;
 
 import org.apache.commons.lang.StringUtils;
 import org.jsonschema2pojo.URLProtocol;
 
 public class URLUtil {
+
+    private static final Set<URLProtocol> LOCAL_PROTOCOL = Arrays.stream(new URLProtocol[]{URLProtocol.NO_PROTOCOL, URLProtocol.FILE}).collect(Collectors.toSet());
+
+    private URLUtil() {
+    }
+
+    public static boolean isLocalUrl(String input) {
+        return LOCAL_PROTOCOL.contains(parseProtocol(input));
+    }
+
+    public static boolean isRemoteUrl(String input) {
+        return !isLocalUrl(input);
+    }
 
     public static URLProtocol parseProtocol(String input) {
         return URLProtocol.fromString(StringUtils.substringBefore(input, ":"));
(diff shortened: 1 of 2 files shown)


jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/util/URLUtil.java
```diff
@@ -21,12 +21,28 @@
...
+import java.util.Arrays;
+import java.util.Set;
+import java.util.stream.Collectors;
 
 import org.apache.commons.lang.StringUtils;
 import org.jsonschema2pojo.URLProtocol;
 
 public class URLUtil {
 
+    private static final Set<URLProtocol> LOCAL_PROTOCOL = Arrays.stream(new URLProtocol[]{URLProtocol.NO_PROTOCOL, URLProtocol.FILE}).collect(Collectors.toSet());
```
joelittlejohn, line 33: Can this just be replaced by `Set.of`?

unkish: I'm afraid it can't since `Set::of` was introduced in Java 9 and current compiler configuration is:
https://github.com/joelittlejohn/jsonschema2pojo/blob/34cc69306fad2f5e27caa4c74cba9e16d8a0234a/pom.xml#L213-L223

That being said it could be changed to `= EnumSet.of(URLProtocol.NO_PROTOCOL, URLProtocol.FILE)` with a small caveat - resulting set would be mutable (however that shouldn't be a problem since it's _private_ and there are no methods returning it's reference.

gr0l: I changed the corresponding line. I am now Using an EnumSet wrapped in an unmuttable set.

gr0l: @joelittlejohn: Can it be merged now?

ctrimble:
@gr0l I know this PR has been open for some time, but I am actively trying to get a 1.2.3 release put together.  This seems like a bug.  I will give it a review and see if we can get it merged.

ctrimble:
@gr0l Hello!  Is there any chance you can get this on top of master and squash it into a single commit?

gr0l:
@ctrimble I merged the current master into my fork. Does github not support squashing, when the merge is applied?

ctrimble:
@gr0l I am running off of the existing convention, which is to bring in PRs with a merge commit.  I was hoping you could just `git rebase -i master` this, drop all of the merge commits and squash your second real commit onto the first, then force push your branch.  The way this is, there will be all kinds of strange merge commits in the history.  Not the end of the world, but it does make sorting through history difficult.

gr0l:
@ctrimble Done

ctrimble:
@gr0l Thanks for the contribution.  I am going to work up an integration test for this.  Once I finish that, this will get merged.

ctrimble:
@gr0l I got a test put together for this.  It fails without your code and passes with it, so everything looks in order.

Thanks again for your PR and your willingness to restructure history.  I appreciate it.

PR #1619, opened 2024-06-16
Add GenerationConfig option to exclude Object Header from toString()

ajaysagar:
Hi, thanks for the very useful library! We are using jsonschema2pojo to perform semantic diffing between multiple JSON payloads, for which we are using toString in certain scenarios to assist comparing objects. Here we would like the option to exclude the object header (object identity hash) so that multiple objects in memory that are logically the same would generate the same output for toString. Thanks!


ajaysagar:
@joelittlejohn anything I can do to move this PR forward? Thanks!

PR #1670, opened 2025-03-05
Fix #1669: JSR 303 annotations on array types

### LINKED ISSUE #1669: BUG: JSR 303 annotations not generated for array types when minLength and/or maxLength are present
I am using the following JSON Schema:

```json
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "title": "BinaryImageAttributes",
  "type": "object",
  "properties": {
    "bytes": {
      "type": "string",
      "media": {
        "binaryEncoding": "base64"
      },
      "description": "Base64-encoded binary image data.",
      "minLength": 1,
      "maxLength": 102400
    }
  },
  "required": ["bytes"]
}
```

This is generating the following Java Bean:

```java

package io.humangraphics.backend.lambda.image.analyze.api.model.request;

import java.util.Arrays;
import javax.annotation.processing.Generated;
import com.fasterxml.jackson.annotation.JsonInclude;
import com.fasterxml.jackson.annotation.JsonProperty;
import com.fasterxml.jackson.annotation.JsonPropertyDescription;
import com.fasterxml.jackson.annotation.JsonPropertyOrder;
import jakarta.validation.constraints.NotNull;


/**
 * BinaryImageAttributes
 * <p>
 * 
 * 
 */
@JsonPropertyOrder({
    "bytes"
})
@Generated("jsonschema2pojo")
public class BinaryImageAttributes {

    /**
     * Base64-encoded binary image data.
     * (Required)
     * 
     */
    @JsonProperty("bytes")
    @JsonPropertyDescription("Base64-encoded binary image data.")
    @NotNull
    private byte[] bytes;

    /**
     * Base64-encoded binary image data.
     * (Required)
     * 
     */
    @JsonProperty("bytes")
    public byte[] getBytes() {
        return bytes;
    }

    /**
     * Base64-encoded binary image data.
     * (Required)
     * 
     */
    @JsonProperty("bytes")
    public void setBytes(byte[] bytes) {
        this.bytes = bytes;
    }

    public BinaryImageAttributes withBytes(byte[] bytes) {
        this.bytes = bytes;
        return this;
    }

    @Override
    public String toString() {
        StringBuilder sb = new StringBuilder();
        sb.append(BinaryImageAttributes.class.getName()).append('@').append(Integer.toHexString(System.identityHashCode(this))).append('[');
        sb.append("bytes");
        sb.append('=');
        sb.append(((this.bytes == null)?"<null>":Arrays.toString(this.bytes).replace('[', '{').replace(']', '}').replace(", ", ",")));
        sb.append(',');
        if (sb.charAt((sb.length()- 1)) == ',') {
            sb.setCharAt((sb.length()- 1), ']');
        } else {
            sb.append(']');
        }
        return sb.toString();
    }

    @Override
    public int hashCode() {
        int result = 1;
        result = ((result* 31)+ Arrays.hashCode(this.bytes));
        return result;
    }

    @Override
    public boolean equals(Object other) {
        if (other == this) {
            return true;
        }
        if ((other instanceof BinaryImageAttributes) == false) {
            return false;
        }
        BinaryImageAttributes rhs = ((BinaryImageAttributes) other);
        return Arrays.equals(this.bytes, rhs.bytes);
    }

}
```

I expect the `private byte[] bytes` field to have the following annotations:

```java
    /**
     * Base64-encoded binary image data.
     * (Required)
     * 
     */
    @JsonProperty("bytes")
    @JsonPropertyDescription("Base64-encoded binary image data.")
    @NotNull
    @Size(minLength=1, maxLength=102400)
    private byte[] bytes;
```

It is currently missing the `@Size` annotation.

sigpwned:
This is a (relatively) straightforward fix of the issue where JSR 303 annotations are not generated for array types.

For example, the schema...

```json
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "title": "BinaryImageAttributes",
  "type": "object",
  "properties": {
    "bytes": {
      "type": "string",
      "media": {
        "binaryEncoding": "base64"
      },
      "description": "Base64-encoded binary image data.",
      "minLength": 1,
      "maxLength": 102400
    }
  },
  "required": ["bytes"]
}
```

...should generate a `@Size` annotation on the `bytes` field due to the presence of `minLength` and `maxLength`, but it does not.

The core issue was that the code (quite reasonably) assumed that `JType#fullName()` would always return values that could subsequently be loaded by `Class#forName()`, modulo any type parameters. However, for arrays (e.g., byte arrays), `JType#fullName()` would return `byte[]`, whereas `Class#forName()` expects `[B`. Simply checking for array types using `JType#isArray()` before inspecting names fixes the issue.

Fix includes tests.

Hopefully this all makes sense! Please send all questions, comments, and thoughts my way, and I will address ASAP.

Thank you for maintaining `jsonschema2pojo`!

### DIFF
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/MinItemsMaxItemsRule.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/MinItemsMaxItemsRule.java
@@ -63,6 +63,14 @@ public class MinItemsMaxItemsRule implements Rule<JFieldVar, JFieldVar> {
     }
 
     private boolean isApplicableType(JFieldVar field) {
+        // Per https://github.com/joelittlejohn/jsonschema2pojo/issues/1669
+        // Class.forName() loads array classes for names like "[B" (for primitives) and
+        // "[Ljava.lang.String;" (for objects). However, JType.fullName() returns the name as
+        // "byte[]" and "java.lang.String[]" respectively. So, we need to check if the type is an
+        // array type before inspecting names, and if so, return true.
+        if (field.type().isArray())
+            return true;
+
         try {
             String typeName = field.type().boxify().fullName();
             // For collections, the full name will be something like 'java.util.List<String>' and we
@@ -76,8 +84,7 @@ public class MinItemsMaxItemsRule implements Rule<JFieldVar, JFieldVar> {
             return String.class.isAssignableFrom(fieldClass)
                     || Collection.class.isAssignableFrom(fieldClass)
                     || Map.class.isAssignableFrom(fieldClass)
-                    || Array.class.isAssignableFrom(fieldClass)
-                    || field.type().isArray();
+                    || Array.class.isAssignableFrom(fieldClass);
         } catch (ClassNotFoundException ignore) {
             return false;
         }
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/MinLengthMaxLengthRule.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/MinLengthMaxLengthRule.java
@@ -63,6 +63,14 @@ public class MinLengthMaxLengthRule implements Rule<JFieldVar, JFieldVar> {
     }
 
     private boolean isApplicableType(JFieldVar field) {
+        // Per https://github.com/joelittlejohn/jsonschema2pojo/issues/1669
+        // Class.forName() loads array classes for names like "[B" (for primitives) and
+        // "[Ljava.lang.String;" (for objects). However, JType.fullName() returns the name as
+        // "byte[]" and "java.lang.String[]" respectively. So, we need to check if the type is an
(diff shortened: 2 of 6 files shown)


ctrimble:
@sigpwned I know this PR has been sitting, but I am trying to get some releases together this month.  I agree that this looks like a bug and not a feature.  I will give this a review and try to get it into the 1.2.3 release.

joelittlejohn:
Thanks again for finding this and providing the fix Andy! Migration to Junit 5 also hit us here and made the changes out of date.

I have made a few tweaks to the solution but kept your logic and the integration test and example schema you provided.  I saw there's also some pre-existing redundant logic here where CodeModel can already return the 'erasure' of a type but this was being done manually.

I also took the decision to remove the unit tests, so we don't need to worry about updating those now. We shouldn't be creating tests that mock out codemodel, as they are very fragile and not very valuable. Code generation features should be verified with integration tests.

I think this one is ready to merge and release in 1.3.3.

PR #1673, opened 2025-04-10
Avoid duplicate "is" prefix for boolean getters

NiklasvonM:
Boolean property names that start with the prefix "is" currently receive another "is" prefix, generating, for example, the getter `isIsValid` for the property name "isValid".

This merge request aims to avoid these duplicate prefixes.

### DIFF
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/util/NameHelper.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/util/NameHelper.java
@@ -219,13 +219,24 @@ public class NameHelper {
     public String getGetterName(String propertyName, JType type, JsonNode node) {
         propertyName = getPropertyNameForAccessor(propertyName, node);
 
-        String prefix = type.equals(type.owner()._ref(boolean.class)) ? "is" : "get";
+        boolean isBoolean = type.equals(type.owner()._ref(boolean.class));
+
+        boolean nameAlreadyHasIsPrefix = isBoolean &&
+                                        propertyName.startsWith("is") &&
+                                        propertyName.length() > 2 &&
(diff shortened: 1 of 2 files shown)


Juphex (approved):
lgtm

ctrimble:
Hello.  I think this needs a feature flag, since it could change output for others.  The flag should be clear about what the feature is and it should be disabled by default.  We will also need tests showing the old behavior without the feature and the new behavior with the feature active.  I am happy to support getting this into the 1.3.0 release.

NiklasvonM:
Thank you for your comment and support!

Implementing the feature flag and integration tests, I noticed that this new feature may lead to duplicate method names. For example, both properties in the following schema would generate the `isValid` getter:
```json
{
  "type": "object",
  "properties": {
    "isValid": {
      "type": "boolean"
    },
    "valid": {
      "type": "boolean"
    }
  }
}
```
Do you have any preferences how to deal with this? I see three options:

- Fall back to the current behaviour if the new one would create conflicting method names.
- Add suffixes similar to `MakeUniqueClassName.makeUnique()`.
- Throw an `IllegalArgumentException`.

joelittlejohn:
@NiklasvonM Thanks for sharing this code for others, but I don't think we will merge this PR.

The way that java properties and their accessor methods are named in Java follows strict rules. All Java libraries use these rules to interpret values correctly and it's a very powerful convention. If you have a property called isValid, then the accessor for this is isIsValid. isValid is an accessor for the a property called valid. See also setIsValid. You probably don't have control over this, hence the PR, but a better option would be calling your property 'valid'.

Adding another config option to produce accessors in a way that breaks these standard accessor rules is just not something we want to maintain. As a general rule, this plugin can't support all possible ways of generating Java, as there are so many different preferences and styles out there, so generally I won't merge PRs of this kind.

Issue #1461, opened 2023-01-05
Maven plugin's `includeGeneratedAnnotation` does not support `jakarta.annotation.Generated`


anno1985:
When using both `includeJsr303Annotations` and `includeGeneratedAnnotation`, it seems one still gets `javax.annotation.(processing.)Generated` instead of [`jakarta.annotaiton.Generated`](https://jakarta.ee/specifications/annotations/2.0/annotations-spec-2.0.html#jakarta-annotation-generated).

joelittlejohn:
So I guess javax has become jakarta, and now we have another Generated annotation to throw into the mix :)

unkish:
Current annotations are part of the JRE, thus shouldn't require additional dependencies.
Would throwing in `jakarta.annotation` in the mix mean yet another flag, as both:
`includeJsr303Annotations` and `useJakartaValidation` are about _validation api_:
https://github.com/joelittlejohn/jsonschema2pojo/blob/8e772064776c51376d63c35674e6dd5a0d5e123c/pom.xml#L373-L382

Whereas [jakarta.annotaiton.Generated](https://jakarta.ee/specifications/annotations/2.0/annotations-spec-2.0.html#jakarta-annotation-generated) would be about _annotations_:
```
<dependency>
    <groupId>jakarta.annotation</groupId>
    <artifactId>jakarta.annotation-api</artifactId>
    <version>2.1.1</version>
</dependency>
```

joelittlejohn:
:weary:

joelittlejohn:
It's funny how an annotation is such a trivial thing, just a tag with no behaviour at all, but it's bounced around so many different locations over the years. Why can't they just include this trivial annotation in Java's core API!

eirnym:
I think it's mostly political reasons between Java Core and J2EE.

darkmastermindz:
Hi any updates on this? @joelittlejohn This is blocking JDK17+ compatibility which is required for CVE fixes.

GregDThomas:
FWIW, https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator-gradle-plugin/README.adoc has the same problem (i.e. javax vs jakarta) - and overcomes it with a single flag "useJakartaEe= true/false" to determine if it uses javax.* or jakarta.* style annotations.

GregDThomas:
In the meantime, I've overcome this issue with the following bit of Gradle;

```
[...]

PR #1684, opened 2025-06-13
Add jakarta.annotation.Generated if useJakartaValidation is true

akuhtz:
Add `jakarta.annotation.Generated` if `useJakartaValidation` is `true` instead of `javax.annotation.processing.Generated`.


unkish:
@ctrimble I see you've added `1.2.3` milestone to this PR, however I didn't see neither any decision on #1461 (_validation-api_ vs. _annotations-api_ - people tend to confuse those) nor any argumentation as to which approach if any will be taken and why.

ctrimble:
@unkish I just classified things as bugs or features.  I hadn't dug into correctness yet.

ctrimble:
@akuhtz Thanks for this PR.  Unfortunately, it will not be merged to master.  There has been significant discussion about this flag and the Generated annotation on #1461 .  The consensus is that the `useJakartaValidation` is not a general flag for Jakarta annotations.  Although lack of support for this annotation is an issue, this is not the path forward for adding support.

PR #1689, opened 2025-07-09
Migrate from junit4 to junit5

unkish:
Another attempt at migrating to junit5, while attempting to:

* make as little changes as possible
* preserve logic of `Jsonschema2PojoRule` (with respect to generated/compiled output structure) as much as possible  eg.
    `target/jsonschema2pojo/IncludeJsr303AnnotationsIT/_1_true/jsr303AnnotationsValidatedForAdditionalProperties`
    `target/jsonschema2pojo/IncludeJsr303AnnotationsIT/_2_false/jsr303AnnotationsValidatedForAdditionalProperties`

### DIFF
--- a/.github/workflows/android-ci.yml
+++ b/.github/workflows/android-ci.yml
@@ -18,7 +18,7 @@ jobs:
         java_version: [8]
     steps:
       - name: Setup Maven Action
-        uses: s4u/setup-maven-action@v1.14.0
+        uses: s4u/setup-maven-action@v1.18.0
         with:
           checkout-fetch-depth: 0
           java-version: ${{ matrix.java_version }}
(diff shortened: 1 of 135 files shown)


joelittlejohn:
I'm keen to merge this immediately @unkish if you are happy. We need to go ahead with this and move forward. It's a huge task but I regret moving from a basic static helper to a Rule, so maybe we could simplify that again some day.

Do you think this can be merged now or is there anything outstanding that you feel needs to be resolved?

unkish:
> It's a huge task but I regret moving from a basic static helper to a Rule, so maybe we could simplify that again some day.

Oh, I didn't know that. Will dig around in git history, thank you for pointing it out.

> Do you think this can be merged now or is there anything outstanding that you feel needs to be resolved?

From my perspective - it could be merged, I didn't disable or remove tests that would not have compiled.
Please feel free to merge as long as you are also satisfied with how it turned out and don't see any issues/improvements that might need to be addressed.

joelittlejohn:
> Oh, I didn't know that. Will dig around in git history, thank you for pointing it out.

It's a big job to change, so no need to do anything, but something we can think on for the future.

PR #1712, opened 2025-09-04
Use spotless to enforce style and apply

juherr:


juherr:
@ctrimble I’m not sure applying the style all at once is a good idea, but I wanted to show the impact and highlight whether the `jsonschema2pojo-eclipse-formatter.xml` should be updated to reduce the diff.
Once you’ve confirmed the file is correct, I can drop the commit with the changes and replace it with [a gradual check](
https://github.com/diffplug/spotless/tree/main/plugin-maven#how-can-i-enforce-formatting-gradually-aka-ratchet).

ctrimble:
@juherr Thanks for submitting this. but it is a distraction right now.  I will take a look when the backlog is more in order.

juherr:
Sure, it is not urgent

eirnym:
I'll add my 2 groszy to the discussion. 

Back in the day I've resigned from spotless plugin both at work and in personal projects because it's annoying: you need to hack a solution to test if it works and… you have to format. 

Additionally, it take too much time to run, so no rapid development either. 

We at work and I in my personal projects usually on commit or even on push (depending on team and project), so pre-commit (or any similar project) checked both locally and on CI works the best.

juherr:
@eirnym I didn’t bind the plugin to any Maven phase and added a CI check as you suggested.
I can also add a few words to explain how to run the plugin on a Git event.

That’s exactly what you had in mind, right?

eirnym:
I haven't read Spotless for a while now, but AFAIK it binds itself to the build phase like it was back in the day. It's easy to check: just make code less pretty and build the project

juherr:
I’ve been using Spotless on many of my projects and haven't run into the issue you mentioned. ;)

eirnym:
I've checked documentation and it's not automatic as you've said. But it's a separate non-automated execution which is still not ideal. It's easy to forget run `<build> spotless` before commit. 

It's possible to write a manual pre-commit hook, but it's still not automated.

juherr:
@eirnym Did you see https://github.com/diffplug/spotless/tree/main/plugin-maven#git-hook ?

eirnym:
Yes, I did. It's not so easy to manage (keep, install, update) and it's not extensible as pre-commit is. 

It's may be ok for a very small project which won't add any other pre commit hooks ever. In teams and in project I work, this tool was replaced with almost direct run of formatters spotless calls.

eirnym:
as of specific hooks to use for format, I recommend to look at this one: https://github.com/macisamuele/language-formatters-pre-commit-hooks

ctrimble:
As an alternative location to bind this, I have a format lifecycle published.  I haven't used it in a while, but we could add an`mvn format` lifecycle and then bind this plugin to that.  That may be a better location for the license header as well.  That plugin is always causing me issues with conflicts in my IDE.

juherr:
@ctrimble What do you mean exactly?  
Are you suggesting creating a custom `maven-format-plugin` that supports `<plugins>` in its configuration,  
or overriding the default Maven `jar` lifecycle by adding a new step to it?

ctrimble:
@juherr We could add https://github.com/ctrimble/maven-format-lifecycle to the project.  Then we would have `mvn format` from the command line.  It just creates a maven lifecycle that has nothing to do with the build.

I haven't messed with that in a few years, but I am not aware of any breaking changes in maven that have broken lifecycle plugins.

PR #1599, opened 2024-04-03
Test on all Java LTS

juherr:
**Summary:**  
- Added CI tests for all Java LTS versions (starting from 8).  
- Integrated examples into the CI workflow to improve test coverage. _https://github.com/joelittlejohn/jsonschema2pojo/pull/1735_
- Fixed various issues to ensure all tests pass successfully.

PR #1714, opened 2025-09-07
Rework CI: run all examples

juherr:
CI changes from #1599, which should not include these updates.
As expected (and already addressed in #1599), the Gradle Android example does not work with Java 17/21.

### DIFF
--- a/jsonschema2pojo-maven-plugin/example/pom.xml
+++ b/jsonschema2pojo-maven-plugin/example/pom.xml
@@ -2,8 +2,8 @@
     xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
     <groupId>org.jsonschema2pojo</groupId>
-    <artifactId>jsonschema2pojo-test</artifactId>
-    <version>0.0.1-SNAPSHOT</version>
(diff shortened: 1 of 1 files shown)


joelittlejohn:
@juherr Thanks so much for all the work you have been doing on this recently! Please do try to keep changes focused on one improvement at a time, as it makes reviewing much easier and allows things to be merged much more quickly. Splitting out these changes was a useful step. Please resist the urge to keep expanding the scope of what you're doing to fix other things at the same time - we can always land more PRs.

Am I right in thinking that the path forward now is to merge this PR first (and accept that those two examples will fail) then rebase #1599 and merge that afterwards?

juherr:
@joelittlejohn These changes originated as part of the experiments for #1599 and have since been refined into the version you see here.  
This PR is required for #1599, and the Android Gradle example currently fails on Java 17 and 21 but works correctly with #1599 as demonstrated there.

joelittlejohn:
@juherr So we merge this one first, right?

juherr:
@joelittlejohn Yes. The failling jobs should be fixed by the next pull request.

ctrimble:
@juherr I am closing this PR, as the same changes are coming from #1599 and I have already started to break that down into smaller chunks.  I will continue to peal history off of that PR, until we get it into a position where the actual JDK update can be reviewed.

Issue #831, opened 2018-02-05
Fesability of using oneOf, anyOf or allOf in schema for jsonschema2pojo.


lvimalchandran:
Hi, 
Can anyone please let me know whether using oneOf, anyOf or allOf in schema for jsonschema2pojo will work or not. Or is there any other option to handle the scenario.

ehrmann:
Have you seen https://github.com/joelittlejohn/jsonschema2pojo/wiki/Proposal-for-allOf,-anyOf-and-oneOf?

ctrimble:
@lvimalchandran These features of JSON Schema have a lot of problems when it comes to static binding.  They are really geared to validation.  In general, it seems like any potential binding solution would require a validator during type selection.  That is blocking support for these keywords.

ilyako87:
Hi,
I read the information here https://github.com/joelittlejohn/jsonschema2pojo/wiki/Proposal-for-allOf,-anyOf-and-oneOf 
But I do not understand, if the jsonschema2pojo allows the use of simple OnOff?
Or the use of these keywords does not work at all?

tidanilocarvalho:
Hi, I am working on a project which used to have jsnoschema2pojo Gradle plugin 0.5.1

the spec has the 
attr: {
     description: some description,
     myannoation1: value1
    myannoation2: value2
      allOf {
          $ref: anotherType
      }
}

In version 0.5.1 this used to work once I did an upgrade to version 1.0.1 it stopped works.

I use a class to create my own annotations, because without this allOf, as this example

attr: {
     description: some description,
     myannoation1: value1
    myannoation2: value2
     $ref: anotherType
}

All that is out of $ref is not used, then my own annotations are not generated

Does anybody have any suggestions?

Is it related to this issue?

AceHack:
Any update on using oneOf

andrevocat:
same question as AceHack: Any update on using oneOf's?

dewthefifth:
I'm curious what the expected Java class would look like in a "oneOf" scenario, since Java doesn't really have support for such a notion. Could one of you post an example of what you'd expect the output to be? Is the problem that you just want the schema generator to generate a class with all candidates of `oneOf` included in the class?

For example are you expecting this schema 
```
{
  "allOf": [
    {
      "oneOf": [
        { "ref": "a.json" },
        { "ref": "b.json" }
      ],
      "oneOf": [
        { "ref": "c.json" },
        { "ref": "d.json" }
[...]

PR #1719, opened 2025-10-02
allOf, anyOf, oneOf and not as rules with warnings.

ctrimble:
This PR shows a potential way we could get `allOf`, `anyOf`, `oneOf` and `not` worked into the project as rules.  Its purpose is to try some implementation ideas.

### DIFF
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/RuleFactory.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/RuleFactory.java
@@ -458,4 +458,8 @@ public class RuleFactory {
         return new UnimplementedTypeRule("oneOf", this);
     }
 
+    public Rule<JClassContainer, JType> getNotRule() {
+        return new UnimplementedTypeRule("not", this);
+    }
+
 }
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/RuleFactory.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/RuleFactory.java
@@ -446,4 +446,16 @@ public class RuleFactory {
         return new JavaNameRule();
     }
 
+    public Rule<JClassContainer, JType> getAllOfRule() {
+        return new UnimplementedTypeRule("allOf", this);
+    }
+
+    public Rule<JClassContainer, JType> getAnyOfRule() {
+        return new UnimplementedTypeRule("anyOf", this);
+    }
+
+    public Rule<JClassContainer, JType> getOneOfRule() {
+        return new UnimplementedTypeRule("oneOf", this);
+    }
+
 }
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/SchemaRule.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/SchemaRule.java
@@ -84,7 +84,17 @@ public class SchemaRule implements Rule<JClassContainer, JType> {
         JType javaType;
         if (schemaNode.has("enum")) {
             javaType = ruleFactory.getEnumRule().apply(nodeName, schemaNode, parent, generatableType, schema);
-        } else {
+        } 
+        else if (schemaNode.has("allOf")) {
+            javaType = ruleFactory.getAllOfRule().apply(nodeName, schemaNode, parent, generatableType, schema);
+        }
+        else if (schemaNode.has("anyOf")) {
+            javaType = ruleFactory.getAnyOfRule().apply(nodeName, schemaNode, parent, generatableType, schema);
+        }
+        else if (schemaNode.has("oneOf")) {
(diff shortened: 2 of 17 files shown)


unkish:
This one seems to be breaking change (no longer creating a class where it used to do so, even though it was an empty one) as it breaks on schema that used to work fine previously:

```json
{
  "type": "object",
  "properties": {
    "storage": {
      "type": "object",
      "oneOf": [
        { "type": "string" },
        { "type": "null" }
      ]
    }
  }
}
```

Now, from my observations (not claiming to be 100% accurate) in most of the cases folks tend to have schemas without
```json
{
  "type": "object",
  "properties": {
...
```
having instead something like:
```{
  "type": "object",
  "$defs": {
    "storage": {
      "type": "object",
      "properties": {
        "storageType": {
          "type": "object",
          "oneOf": [
            { "type": "string" },
            { "type": "null" }
          ]
        }
      }
    }
  }
}
...
```
Since `definitions`, `$defs` (https://github.com/joelittlejohn/jsonschema2pojo/pull/1523) are not being processed - then no class is generated for `storage` / `storageType`.

1. given the above it feels like `ObjectRule` should check for presence of `allOf` / `oneOf` / `anyOf`.
2. a separate question raised in #831 - why do `allOf` / `oneOf` / `anyOf` keywords have to be treated specially and emit warnbing whereas other keywords shouldn't.
3. `... is not implemented.` may leave a false impression of implementation being developed whereas it might not always be the case. Wouldn't `... is not supported` or similar be better way to signal absence of processor/processing ability ?

ctrimble:
I will take a look at the definition of these and see if this is even considered valid.

I like the warnings for several reasons:

1. It makes it clear that we hit a condition that should alter the output, but it didn't.
2. It provides a rule to implement with custom logic.  The rules in this project are closed to extension, so lots of code needs to be copied when adding hooks like this.
3. It gives the project a starting point to handle these keywords.

ctrimble:
I took a look at some of the more recent specifications they have posted at [json-schema.org](https://json-schema.org/specification).  A couple of interesting things I noticed.  First, they have buried draft 04.  Also, they have created terminology for keywords and these are now considered `applicator` keywords.

Looking at the more current specs, these keywords can not only stand alone, but they can be combined at almost any level and with each other.  I am not sure the fundamental design of this library can really support these.  We would almost need to do two passes over the schemas to compute the resulting types, kind of like a gather and scatter type algorithm.  Recursive decent type algorithms just aren't going to be practical, at least where we are accumulating information in CodeModel.

A little more investigation on this may be worth while, but adding a preprocessor/transformer for the schemas might be the best way to give users support without a complete redesign.  This could be done with a `Function<Schema, Schema>`,  allowing users to insert something like [JSLT](https://github.com/schibsted/jslt) or custom code to handle their specific edge cases.

unkish:
> I like the warnings for several reasons:
...

I'm not against them.
What I'm asking is why treating `*Of` somehow specially and emit warnings whilst be silent upon encountering any other _unknown_ keywords ?

P.S. Emitting warning(s) doesn't necessarily has to imply that an empty class will not be created.

joelittlejohn:
I think this will be too noisy @ctrimble as it will introduce potentially huge numbers of warnings that were not present before. I think @unkish is also right that if you set expectations that the library warns on unknown schema keywords, but you don't warn on other ignored things, you are creating potential confusion.

> Looking at the more current specs, these keywords can not only stand alone, but they can be combined at almost any level and with each other. 

There are actually a lot of fundamental reasons why these rules are not directly comparable to language types like we have in Java. This poses major difficulties in supporting them universally. I've written long comments about this in various issues. If you are validating then you can just validate all the rules (does the data satisfy 'all of' these rules? and does the data satisfy 'any of' these rules?), and there's really no problem in highly complex validations. Type systems are more limited though, and Java's isn't even a particularly powerful type system. Inferring valid structures, and then valid types, from these becomes very difficult. These rules can't actually appear at _any_ level, only in a schema, but yes it's hard/impossible to support any combination of these. It's why support for these keywords has not been an easy thing to add. This before you even get started on simply limitations like the fact that there's no useful union type for e.g. Integer and String in Java, just Object (or something arbitrary like Comparable).

I think there is at least a little light at the end of the tunnel. I do think we can have a useful strategy for each of these keyqords, albeit with some limitations. The recursive nature of this - they can appear in any schema - is not so much of a problem.

ctrimble:
@unkish @joelittlejohn thanks for the input.  I think this is enough to show that just adding some rules is not going to be a first step toward getting this to work.  Here are the learnings and then I will close this PR out.

1. If we want to add warnings, we would need to add a validator for what we support or add some other mechanism that can detect keywords that we didn't consume.
2. If we do produce warnings for unknown keywords, we need to do something to dedup them, so that warnings are not super noisy.
3. Preserving the "subtypes" that were combined is really not a good idea.  JsonSchema's nature as a validator means that these types will not make much sense.

I have the beginnings of a transformation to get this supported.  I am going to create a branch in this repo for that one and open another PR for it.

PR #1720, opened 2025-10-04
Handle allOf, anyOf, oneOf, and not by repeated merging of properties.

ctrimble:
This PR shows how we could scatter properties from keywords like `allOf` into the containing schema.  The algorithm will work in parallel and will handle cyclic references that may exist in the schemas.  The idea is to place something like this between the schema store and the existing rules, converting these keywords into schemas with all of their properties expanded.

The current implementation contains a `maxSteps` property, but we could convert this into a `maxSeconds`, if we want to bound the algorithm by time as opposed to complexity of the schemas.

### DIFF
--- a/jsonschema2pojo-core/pom.xml
+++ b/jsonschema2pojo-core/pom.xml
@@ -97,6 +97,33 @@
             <version>4.5.14</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>net.karneim</groupId>
+            <artifactId>pojobuilder</artifactId>
+            <version>4.3.0</version>
+            <scope>provided</scope>
+        </dependency>
     </dependencies>
+
+    <build>
+        <plugins>
+            <!-- 
+              We can remove this before merge.  Included to make iteration faster.
+            -->
+            <plugin>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                <annotationProcessorPaths>
+                    <path>
+                        <groupId>net.karneim</groupId>
+                        <artifactId>pojobuilder</artifactId>
+                        <version>4.3.0</version>
+                    </path>
+                </annotationProcessorPaths>
+                <annotationProcessors>net.karneim.pojobuilder.processor.AnnotationProcessor</annotationProcessors>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
 
 </project>
--- a/jsonschema2pojo-core/src/test/java/org/jsonschema2pojo/transform/SchemaReducer.java
+++ b/jsonschema2pojo-core/src/test/java/org/jsonschema2pojo/transform/SchemaReducer.java
@@ -1,0 +1,165 @@
+/**
+ * Copyright © 2010-2020 Nokia
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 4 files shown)


ctrimble:
@joelittlejohn @unkish this PR is still just a test demo, but we could put something like this into the project.  For now, please keep any feedback to issues with viability.  I have cut a ton of corners (added builder generator, using Commons Lang Tuples, etc.).  We can change those thing later.  The purpose, for now, is just to get something concrete to react to.

I am thinking we can turn this into a wrapper around the SchemaStore, so that the rest of the project doesn't need to understand this change.  We could flip this feature on and off by installing an identity transform, instead of a transform that expands properties, so it could live in the project for a while as an experimental feature.

This would change the behavior of the project a little, since document loading would move to the transform, instead of when a rule encountered a `$ref`, but we could probably hide that detail from the rule implementations.

ctrimble:
I have been working on a set of transforms that we can use to handle `allOf`, `anyOf`, `oneOf` and `not` as a transform that is applied before the existing rules and I have identified a case where this will definitely not work.  Consider a schema where a field can be both either a `date` and a `date-time` string:

```
{
  "type": "object",
    "anyOf": [
      { "properties": { "temporal": { "type": "string", "format": "date" } } }
      { "properties": { "temporal": { "type": "string", "format": "date-time" } } }
    ]
  }
  ```
  
  There is no good way to erase this `anyOf` and have the rest of the rules work, since this project uses `format` to determine the type of the field for a given property.  I think the best we could do here is:
  
  ```
  {
    "type": "object",
    "properties": {
      "temporal": {
        "type": "string",
        "anyOf": [ { "format": "date" }, { "format": "date-time" } ]
       }
     },
    "anyOf": [
      { "properties": { "temporal": { "type": "string", "format": "date" } } }
      { "properties": { "temporal": { "type": "string", "format": "date-time" } } }
    ]
  }
  ```
  
   This would result in a Java object being generated with a `String` field.  Conversion to the `Date` or `DateTime` would be left to the user.
   
   Personally, if a bunch of time is going to be invested into getting these keywords to work, I would like it to make my life extremely easy when using the project's output.  I think that would include creating a set of `JsonSerialzier`s and `JsonDeserializer`s for situations like these and generate fields of the closest super type or interface of all the alternatives.  Even better would be generating some kind of wrapper type that represents the alternative types, like a `BiAlternative<Date, DateTime>` where the type mapping is all handled in generated code.

ctrimble:
I am going to get the code I have cleaned up and pushed.  Then I am going to open a competing draft where these keywords are implemented in the existing rules.  My gut is telling me that getting an easy implementation of these rules will require adding some kind of extra model that is a little closer to the Java we are trying to generate, but not as concrete as CodeModel.

ctrimble:
I pushed some transform examples and an example of a TransformingSchemaStore.  Since it appears that inlining `$ref`s will not work, doing multiple passes of transformation is most likely not needed and this code could be simplified.  I am not a huge fan of this approach, so I am going to leave this for now and look into supporting this in the rules.

joelittlejohn:
@ctrimble I'm going to close this one in favour of #1775. I want to take an unobtrusive approach, even if there are some compromises. I've proposed a route forward there which should satisfy most use-cases.

PR #1724, opened 2025-10-17
Update examples

ctrimble:
This PR was extracted from work done by @juherr on #1599 . It updates the example projects, making them more consistent and able to build with more JDKs.  Here is what has changed:

- The Maven based example was relocated under `jsonschema2pojo-maven-plugin/example`, to be consistent with the gradle examples.
- `mvnw` was added to the maven example.
- `gradlew` was updated in both the gradle examples.
- dependabot configuration was added for the examples.

**Note:** The gradle example now requires JDK 17 when using `gradlew`.  Gradle, in general, has made it difficult to upgrade Gradle and still use JDK 8.

### DIFF
--- a/.github/workflows/android-ci.yml
+++ b/.github/workflows/android-ci.yml
@@ -11,21 +11,27 @@ on:
 
 jobs:
   buildTest:
-    name: Build JDK ${{ matrix.java_version }}
+    name: JDK ${{ matrix.version.java }} | Example Gradle (Android)
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        java_version: [8]
+        version:
+          - { java: 8,  gradle: '7.6.4' }  # AGP Java 8 => Gradle 7.6.4 (MavenPlugin legacy)
     steps:
       - name: Setup Maven Action
         uses: s4u/setup-maven-action@v1.18.0
         with:
           checkout-fetch-depth: 0
-          java-version: ${{ matrix.java_version }}
+          java-version: ${{ matrix.version.java }}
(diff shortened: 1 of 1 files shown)


juherr:
I'm not sure it's a good idea to keep the wrappers inside the example projects because Dependabot can't upgrade them automatically.  
Also, as you know, the Gradle version isn't fixed and depends on the JDK in use.  

I'd strongly recommend managing the Maven and Gradle versions used by examples externally, in the CI configuration, to keep things consistent and maintainable across environments.

ctrimble:
@juherr I understand your position on these wrappers.  Taking them out of the automated builds is necessary.

However, the examples are there for people to use.  I don't like the idea of taking a regression because some service cannot do upgrades for us or we cannot run a matrix build against it.  It can live there for the humans and the automation can do something else.

ctrimble:
I made some minor adjustments to the Andorid Sampe CI, just as a temporary measure to keep the CI working.

jsonschema2pojo-maven-plugin/example/pom.xml
```diff
@@ -17,7 +17,7 @@
             <plugin>
                 <groupId>org.jsonschema2pojo</groupId>
                 <artifactId>jsonschema2pojo-maven-plugin</artifactId>
-                <version>${project.version}</version>
+                <version>1.2.1</version>
```
juherr: By fixing the version here, the examples no longer test the current snapshot during integration. Was that the intended behavior?
At least, use a property that could be overridden from the command line.

PR #1734, opened 2025-10-17
Adds an integration test for network sourcePaths.

ctrimble:
This PR adds a test to verify that #1614 fixes issues with network based sourcePaths.


ctrimble:
This test failed without #1614 and passed with it.

PR #1742, opened 2025-12-10
Set scope of 'org.apache.groovy:groovy' dependency to 'provided'

unkish:


unkish:
Change is related to jdk 17 update.
Seems like setting scope for `org.apache.groovy:groovy` to `provided` in `jsonschema2pojo-gradle-plugin`  slipped right trough my eyes.

Previously `org.apache.groovy:groovy` was resolved as transitive dependency of `dev.gradleplugins:gradle-api` (scope `provided`) however as newer version of `org.apache.groovy:groovy` was needed to compile using JDK17 both `org.apache.groovy:groovy` and `org.apache.groovy:groovy-all` transitive dependencies had to be excluded from `dev.gradleplugins:gradle-api` and `org.apache.groovy:groovy` had to be added explicitly.

PR #1743, opened 2025-12-10
Remove Ant task (no longer supported)

unkish:


unkish:
Following places would need some additional work/changes removing mentioning of ANT:
* wiki pages
* reference/link to "Ant task" on https://www.jsonschema2pojo.org/

joelittlejohn:
## For anyone that needs to use jsonschema2pojo via Ant

Ant now makes up such a small proportion of users we can no longer justify maintaining the Ant task. It is not cost free to continue having to update and maintain the task, and ensure that new features are added there too. Ant itself has not had a minor version release in the last 10 years, with only patch releases in that time.

You can continue to use the Ant task from jsonschema2pojo 1.2.2 or earlier. From jsonschema2pojoo 1.3.0 onwards, there is no Ant task.

If you're looking for advice on how to use the Ant task, check out [this historical version of the Getting Started guide](https://github.com/joelittlejohn/jsonschema2pojo/wiki/Getting-Started/c497258361fc50d63708d41569864c3911aa12b4#the-ant-task).

PR #1746, opened 2026-01-14
Handle null when generating from array example

lalmeras:
When schema is generated from an existing JSON array, null values are used to override existing field. If the item of the json array contains null field, generated field model use Object type. If json array contains null field for a given field, any previously discovered schema is lost.

This fix modifies mergeObjectNodes to ignore any null field when updating field schema. It allows to correctly merge all example schemas from array, and get rid of erroneous Object fallback when last example use a null value.

Example:

```json
[
{"field": {"subfield": "test"}},
{"field": null}
]
```

Before:

```java
class Type {
private Object field;
}
```

After:

```java
class Type {
private Field field;
}

class Field {
private String subfield;
}
```

### DIFF
--- a/jsonschema2pojo-core/src/test/java/org/jsonschema2pojo/SchemaMapperTest.java
+++ b/jsonschema2pojo-core/src/test/java/org/jsonschema2pojo/SchemaMapperTest.java
@@ -16,9 +16,15 @@
 
 package org.jsonschema2pojo;
 
-import static org.hamcrest.MatcherAssert.*;
-import static org.hamcrest.Matchers.*;
-import static org.mockito.Mockito.*;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.notNullValue;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.eq;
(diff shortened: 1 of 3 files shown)


joelittlejohn:
@lalmeras Nice change, thank you. Would it be possible to add a test for this?

lalmeras:
I updated (amend and rebase) the PR with a test-case.

jsonschema2pojo-core/src/test/java/org/jsonschema2pojo/SchemaMapperTest.java
```diff
@@ -37,6 +43,7 @@
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.sun.codemodel.JCodeModel;
 import com.sun.codemodel.JPackage;
+import com.sun.codemodel.JType;
```
unkish: Unused import

lalmeras:
@unkish fixed

joelittlejohn:
Closing as a duplicate of #1772. 

@lalmeras I removed the test from SchemaMapperTest and added a new integration test instead. It's how we test all the Java generation features of this library, we have an example schema (or json) like the one you provided, and we have an integration test that generates some Java classes. We compile them, and make sure they work correctly.

Thanks for raising this one and providing the implementation!

PR #1750, opened 2026-01-31
Improve speed of integration tests by safely reusing compiler and filemanager

joelittlejohn:
The file manager reads a large number of jars from the classpath so it's very useful to cache this. We use ThreadLocals to ensure that these tests could be parallelised. As part of this I have removed the ability to use a custom compiler in the tests. I plan to remove the Dynamics Properties feature soon.

Parallelising these tests reduces runtime further.

Total changes here cut the IT test run time by roughly 65%.

### DIFF
--- a/jsonschema2pojo-integration-tests/pom.xml
+++ b/jsonschema2pojo-integration-tests/pom.xml
@@ -29,6 +29,15 @@
             </plugin>
             <plugin>
                 <artifactId>maven-failsafe-plugin</artifactId>
+                <configuration>
+                    <properties>
+                        <configurationParameters>
+                            junit.jupiter.execution.parallel.enabled = true
+                            junit.jupiter.execution.parallel.mode.default = concurrent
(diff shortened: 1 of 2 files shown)


unkish:
My apologies for joining late.
There are few things that could be cleaned up further:

* `org.jsonschema2pojo.integration.util.Compiler::systemJavaCompiler`
* `org.jsonschema2pojo.integration.util.Compiler::eclipseCompiler`
* `org.eclipse.jdt:ecj` dependency from both `jsonschema2pojo-integration-tests` and project pom
* `ecj.version` from within `properties` section of main `pom.xml`

P.S. It also seems that `CompilerWarningIT` was the only one (besides `Jsonschema2PojoRuleTest`) invoking `Jsonschema2PojoRule::getDiagnostics`, which in theory means that functionality ralated to 'captureDiagnostics' could also be removed.

PR #1768, opened 2026-02-01
Add code quality tools with dedicated GitHub Actions

joelittlejohn:
Add four code quality tools, each in its own Maven profile and GitHub Action:
- SpotBugs (-Pspotbugs): Static bug detection
- PMD (-Ppmd): Code style and best practices
- Error Prone (-Perrorprone): Compile-time error checking
- JaCoCo (-Pjacoco): Code coverage reporting with aggregate report

These tools only run in their dedicated workflows, not during normal CI builds.

Also fixes:
- FormatRule: Make static date format fields final (SpotBugs high priority)
- JsonPointerUtils: Replace double-brace initialization (Error Prone)

This is just an experiment for now. I'm interested to see what it's like to have these separate actions run and examine the code.

### DIFF
--- a/.github/workflows/infer.yml
+++ b/.github/workflows/infer.yml
@@ -1,0 +1,73 @@
+name: Infer
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+  workflow_dispatch:
+
+jobs:
+  infer:
+    runs-on: ubuntu-latest
+    name: Facebook Infer Static Analysis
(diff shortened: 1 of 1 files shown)


joelittlejohn:
I feel bad for the amount of GitHub's compute I'm using :) but are these useful? Some more work needed to create useful reports I think, but I like how the dedicated GitHub actions mean that the checks on GitHub show exactly which tool is complaining.

PR #1775, opened 2026-02-08
Implement allOf/anyOf/oneOf

joelittlejohn:
This is a vision for allOf/anyOf/oneOf that I've been thinking about for many years. The approach is simple but will hopefully support a very large percentage of use-cases for these keywords.

* **allOf:** All schemas are merged together to create a superset of schema rules.
* **anyOf / oneOf:** We generate each individual subtype, then create a 'marker' interface which all types implement. When working with types, users may have to rely on type checks and casts, but since Java has no union type (and adding intervening types is ugly) I think this is reasonable.

To achieve polymorphic deserialisation, there's a new 'discriminator' keyword, similar to OpenAPI's discriminator keyword, to specify the JSON property that will indicate the type name.

This PR is not complete, but it shows where I think we'll go.

### DIFF
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/AbstractAnnotator.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/AbstractAnnotator.java
@@ -101,6 +101,7 @@ public abstract class AbstractAnnotator implements Annotator {
         return false;
     }
 
+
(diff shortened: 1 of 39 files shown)


tamademicheli:
@joelittlejohn  i created as well a pull request, i didn't see you had already one, well good, how can i help to push a bit this one, shall i checkout and fix the android part ?

PR #1778, opened 2026-02-09
Fix JLS 9.7.4 type-use annotation placement when using validation annotations

### LINKED ISSUE #1776: Breaking Change in 1.3.3: Type-use @Valid on qualified/parameterized types causes "type annotation not expected here"
## Summary

After upgrading to 1.3.3 (which includes PR #1771 – support for type-use annotations for array item validation), compilation fails when generated POJOs have array/list properties whose item type is a **qualified** or **parameterized** type (e.g. from `existingJavaType`).

The compiler reports:

```
type annotation @jakarta.validation.Valid is not expected here
(to annotate a qualified type, write java.lang.String)
```

## Root cause

PR #1771 changed array item validation from:

```
@Valid
List<Item> items
```

To:
```
List<@Valid Item> items
```

That is correct and valid when the element type is a simple name (Item). However, when the schema uses existingJavaType (e.g. String) or the generated type is a qualified name, the plugin still emits @Valid in front of the full type, e.g.: List<@Valid String>. 

Per JLS (type annotations), a type annotation on a qualified type must appear immediately before the last identifier, not before the whole qualified type. So the generated code is invalid.

joelittlejohn:
Place type-use annotations directly before the simple type name rather than before the fully-qualified name. This produces correct code like `com.example.@Valid MyItem` instead of `@Valid com.example.MyItem` when the type cannot be imported due to a name collision.

When the type is imported, simple names are used: `@Valid String`.

From the language spec:

> For example, assume an annotation type TA which is meta-annotated with
> just https://github.com/target(ElementType.TYPE_USE). The terms
> `@TA java.lang.Object` and `java.@TA lang.Object` are illegal because
> the simple name to which `@TA` is closest is classified as a package name.
> On the other hand, `java.lang.@TA Object` is legal.

Simplify JAnnotatedClass by removing 2 of 3 reflection points:
- Store annotation JClass refs directly instead of reflectively constructing JAnnotationUse instances
- Return this from substituteParams (only concrete types are wrapped)
- Keep importedClasses reflection (unavoidable with codemodel 2.6)

Closes #1776

### DIFF
--- a/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/model/JAnnotatedClass.java
+++ b/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/model/JAnnotatedClass.java
@@ -18,13 +18,13 @@ package org.jsonschema2pojo.model;
 
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
 import java.lang.reflect.InvocationTargetException;
-import java.lang.reflect.Method;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
-import java.util.Objects;
 
 import com.sun.codemodel.JAnnotationUse;
 import com.sun.codemodel.JClass;
@@ -50,11 +50,27 @@ import com.sun.codemodel.JTypeVar;
  * <li>{@link JAnnotatedClass#annotated(Class)} creates a new <em>type</em> with embedded annotations.
  * Example: {@code private List<@NotNull String> items;}</li>
  * </ul>
- * The implementation of this feature relies on reflection and hooking into the JAnnotationUse to access a
- * package-private constructor. This helps us to avoid a breaking change and forking codemodel, but may be a fragile
- * approach.
+ * This implementation uses reflection to access {@code JFormatter}'s private {@code importedClasses} field,
+ * which is needed to determine correct annotation placement per JLS §9.7.4. This is unavoidable since
+ * codemodel 2.6 provides no public API for querying import state.
  */
 public class JAnnotatedClass extends JClass {
+
+    /** Matches codemodel's package-private JFormatter.CLOSE_TYPE_ARGS */
+    private static final char CLOSE_TYPE_ARGS = '\uFFFF';
+
+    private static final Field IMPORTED_CLASSES_FIELD;
+    private static final Constructor<JAnnotationUse> ANNOTATION_USE_CONSTRUCTOR;
+    static {
+        try {
+            IMPORTED_CLASSES_FIELD = JFormatter.class.getDeclaredField("importedClasses");
+            IMPORTED_CLASSES_FIELD.setAccessible(true);
+            ANNOTATION_USE_CONSTRUCTOR = JAnnotationUse.class.getDeclaredConstructor(JClass.class);
+            ANNOTATION_USE_CONSTRUCTOR.setAccessible(true);
+        } catch (NoSuchFieldException | NoSuchMethodException e) {
+            throw new ExceptionInInitializerError(e);
+        }
+    }
 
     private final JClass basis;
     private final List<JAnnotationUse> annotations;
@@ -82,10 +98,8 @@ public class JAnnotatedClass extends JClass {
      */
     private static JAnnotationUse createAnnotationUse(JClass annotationClass) {
         try {
-            Constructor<JAnnotationUse> c = JAnnotationUse.class.getDeclaredConstructor(JClass.class);
-            c.setAccessible(true);
-            return c.newInstance(annotationClass);
-        } catch (IllegalAccessException | NoSuchMethodException | InvocationTargetException | InstantiationException e) {
+            return ANNOTATION_USE_CONSTRUCTOR.newInstance(annotationClass);
+        } catch (IllegalAccessException | InvocationTargetException | InstantiationException e) {
             throw new RuntimeException("Failed to create JAnnotationUse", e);
         }
     }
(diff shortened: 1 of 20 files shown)


joelittlejohn:
Okay, that was a wild ride. A lot more change needed here than I was expected, to take over rendering of the type and make sure the annotations are inserted in the right place, but also to ensure that we respect imported classes and don't use fully-qualified class names where we shouldn't.

unkish:
There seems to be an issue ~~somewhere, but unfortunately I did not have time yet to dive into where exactly the problem is~~.
Consider following reproducer case:

`foo.json`:
```json
{
  "type": "object",
  "additionalProperties": {
    "type": "array",
    "items": {
      "type": "string"
    }
  }
}
```

Code that is generated for `Foo` has issue, note that definition part is missing second type parameter:
```java
    @JsonIgnore
    private Map<@Valid String> additionalProperties = new LinkedHashMap<String, Set<Bar>>();
```

<br>

Update: in given case it seems that both `ValidRule::isArray` and `ValidRule::isMap` pass, however since `isArray` check is done first, then logic for annotating array item is applied.
I'm looking whether switching places for `if (isArray(node))` and `if (isMap(field.type())` would break anything

joelittlejohn:
Ah yes, this is specific to additionalProperties, where the Map is implied but the schema relates to the map values. Great catch!

joelittlejohn:
This is likely not just about swapping the order. We need to think about how additionalProperties works in this case and make sure it is treated slightly differently.

unkish:
That approach might be even "cleaner".

I've noticed aslo that in case of following schema:
```json
{
  "type": "object",
  "additionalProperties": {
    "type": "array",
    "uniqueItems": true,
    "items": {
      "type": "string"
    }
  }
}
```

Following java is generated:
```java
import java.util.Set;
import javax.annotation.processing.Generated;
import com.fasterxml.jackson.annotation.JsonAnyGetter;
import com.fasterxml.jackson.annotation.JsonAnySetter;
...

@JsonPropertyOrder({

})
@Generated("jsonschema2pojo")
public class Foo {

    @JsonIgnore
    private Map<String, java.util.@Valid Set<String>> additionalProperties = new LinkedHashMap<String, Set<String>>();

    @JsonAnyGetter
    public Map<String, Set<String>> getAdditionalProperties() {
        return this.additionalProperties;
    }
...
```

Note how there is `import java.util.Set;` present, getter has return type `Map<String, Set<String>>` field definition however has fqcn in a type parameter `Map<String, java.util.@Valid Set<String>>`
This does not affect compilation output, just catched my eye.

unkish:
Question not related to this specific commit:
Should `ValidRule` have a "guard" against something like
```json
{
  "type": "object",
  "properties": {
    "bar": {
      "type": "object",
      "existingJavaType": "java.util.Map"
    },
    "foo": {
      "type": "array",
      "items": {
        "type": "string"
      },
      "existingJavaType": "java.util.List"
    }
  }
}
```

Upd: I guess not necessarily as `DefaultRule::getDefaultList` would get `java.lang.IndexOutOfBoundsException: Index: 0`

joelittlejohn:
Yes, you've definitely got it! I'm working on this.

joelittlejohn:
I'm going to put some extra guards in too. There's too many ways to easily break this.

joelittlejohn:
Okay. so I've added more tests and guards here. 

There's one last edge case I'm taking a look at:

```json
{
    "type": "object",
    "additionalProperties": {
        "type": "array",
        "items": {
            "type": "array",
            "items" : {
                "type": "array",
                "items" : {
                    "type": "array",
                    "items" : {
                        "type": "string"
                    }
                }
            }
        }
    }
}
```

I think that the Valid annotations should be applied recursively. Not sure if this is fixable but let's see.

unkish:
I can check the latest changes tomorrow, if that is alright.

joelittlejohn:
Yes, please don't look at this today. Let's mull it over. I also want some more time to consider whether to change this more fundamentally.

Exploring this recursive situation, I think it has revealed to me that fundamentally the ValidRule now needs to be applied to types not fields. This might actually simplify the entire thing.

unkish:
I had a look at changes today and spotted following:
1. If additional properties is not overridden/specialized
```json
{
  "type": "object",
  "additionalProperties": true
}
```
valid code is generated, however there will be no `@Valid` annotation:
```java
    @JsonIgnore
    private Map<String, Object> additionalProperties = new LinkedHashMap<String, Object>();
```
if additional properties are overriden
```json
{
  "type": "object",
  "additionalProperties": {
    "existingJavaType": "java.lang.Object"
  }
}
```
then `@Valid` would be present:
```java
    @JsonIgnore
    private Map<String, @Valid Object> additionalProperties = new LinkedHashMap<String, @Valid Object>();
```

2.  There seems to be an issue with `PropertiesRule::addOverrideBuilder(JDefinedClass, JMethod, JVar)` which can cause generation of 2 "with" methods, consider following example:
`bar.json`
```json
{
  "type": "object",
  "javaType": "com.example.Bar",
  "properties": {
    "version": {
      "type": "integer",
      "required": true
    }
  }
}
```
`foo.json`
```json
{
  "type": "object",
  "javaType": "com.example.Foo",
  "extendsJavaClass": "com.example.Bar",
  "properties": {
    "version": {
      "type": "integer"
    }
  }
}
```
with `includeJsr303Annotations = true`  and `generateBuilders = true`, produces:
```java
    public Foo withVersion(@Valid Integer version) {
        this.version = version;
        return this;
    }

    @Override
    public Foo withVersion(@Valid Integer version) {
        super.withVersion(version);
        return this;
    }
```

method with `@Override` is generated since `parentParam.type()` returns `JReferencedClass(Integer)` whilst `thisJDefinedClass` has `JMethod` with parameter being `JAnnotatedClass(Integer)`

3. There is no `@Valid` annotation on properties that are referenced in schema:
`foo.json`
```json
{
  "type": "object",
  "properties": {
    "baz": {
      "$ref": "baz.json"
    }
  }
}

```
`baz.json`
```json
{
  "type": "object"
}
```

joelittlejohn:
Thanks @unkish! I'm happier with the implementation now, moving the annotation step from 'when we add a field' to 'when we use a type' was a key insight and everything is now much simpler. Even the use of deeply nested classes works now, producing results like `List<List<List<List<@Valid MyItem>>>>`, which is what we want.

I'll look at these other things tonight.

joelittlejohn:
On these:

> 1. If additional properties is not overridden/specialized

Fixed.

> 2. There seems to be an issue with PropertiesRule::addOverrideBuilder

I actually can't replicate this. I added a test 'jsr303ValidAnnotationOnClassWithBuilders' but it seems to compile fine and I don't see the problem you mentioned when I look at the class. Maybe you can see why the test does not produce the problem?

> 3. There is no @Valid annotation on properties

Fixed.

I'm sure there are still more edge cases. I'm amazed this has proven to be so hard to get right, but I think we have a pretty good implementation now. Thanks so much for your comments and reviews @unkish!

unkish:
@joelittlejohn I've [tweaked](https://github.com/unkish/jsonschema2pojo/pull/6/changes/f29e841142956cd7bc803a27544a7b32e8addf61) a bit the test to showcase the issue with `generateBuilders`.
Test now fails with:

> target/jsonschema2pojo/IncludeJsr303AnnotationsIT/_1_true/jsr303ValidAnnotationOnClassWithBuilders/generate/com/example/Dog.java:45: error: method withAge(Integer) is already defined in class Dog
    public Dog withAge(`@Valid` Integer age) {
               ^
1 error



One way to fix it would be considering changing `JAnnotatedClass::equals` method to only take into account `basis` equality, leaving out `annotations` equality with reasoning that `JAnnotatedClass` represents a wrapper/proxy around `JClass`.

joelittlejohn:
Since there's no way to get the params for JAnnnotationUse, I decided to check the basis and the annotations for equality (but not their params).

I'm also concerned about adding `@Valid` to everything as I think people will not like this pointless extra noise. I added a check to avoid adding annotations to scalar types like java.lang.Integer, etc. Really hoping that it's the last change 🙃.

unkish:
Should `isContainer` also be made `protected` and/or `Optional` added to `isContainer` check ?

Besides `Optional` (and some other Optional specializations) being a container and possibility to have `isScalar` protected I haven't spotted anything more.

Great job 🎉

I really appreciate your patience tolerating my constant nudges! 😄

joelittlejohn:
A team effort!

unkish:
Before releasing I would like to double check if hibernate validator `@Valid Optional<SomeObject>` would produce warning (treating `Optional` as container).

Upd: validator emitted warnings.
The reason why I wanted to check it is that previously
```json
{
  "type": "object",
  "properties": {
    "optional": {
      "type": "object",
      "existingJavaType": "java.util.Optional<other.pkg.AComplexObject>"
    },
    "mapWithExistingType": {
      "type": "object",
      "existingJavaType": "java.util.Map<String, other.pkg.AComplexObject>"
    }
  }
}
```
previously generated:
```java
    @JsonProperty("optional")
    @Valid
    private Optional<other.pkg.AComplexObject> optional;
    @JsonProperty("mapWithExistingType")
    @Valid
    private Map<String, other.pkg.AComplexObject> mapWithExistingType;
```
Now we generate
```java
    @JsonProperty("optional")
    private Optional<other.pkg.AComplexObject> optional;
    @JsonProperty("mapWithExistingType")
    private Map<String, other.pkg.AComplexObject> mapWithExistingType;
```
there is no more `@Valid` neither on type itself which with new validator would cause warnings, yet unwrap and validate content, nor there is `@Valid` on type parameter.

PR #1779, opened 2026-02-12
Fix #926: Apply enum defaults for types already on the classpath

### LINKED ISSUE #926: default enum doesn't work when defining javaType
If I define the enum in a separate file, I the value set is always null.
Here is an example of my enum
```
{
  "javaType": "com.example.MyEnum",
  "type": "string",
  "enum": [ "A", "B", "C"  ]
}
```

Here is an example of an object using the enum
```
{
  "type": "object",
  "javaType": "com.example.MyClass",
  "properties": {
    "myenum": {
      "javaType": "com.example.MyEnum",
      "default": "A"
    }
  }
}
```

I am expecting the value should be set:
```
    @JsonProperty("myenum")
    private MyEnum myenum = MyEnum.fromValue("A");
```

The actual output is
```
    @JsonProperty("myenum")
    private MyEnum myenum = null;
```

manerow:
## Problem

When an enum type already exists on the plugin classpath (e.g. from a dependency JAR), `DefaultRule.getDefaultValue()` fails to generate the `fromValue(...)` initialiser.

The root cause is that `EnumRule` returns a `JDirectClass` (a reference to the pre-existing classpath class) instead of a `JDefinedClass` (a freshly generated class). `DefaultRule` only checks `fieldType instanceof JDefinedClass` when deciding whether to apply enum defaults, so classpath enum types fall through to `return JExpr._null()`.

This means any project that depends on a JAR containing pre-compiled enum classes while regenerating POJOs from the same schemas will silently lose all enum default values.

### Example

Given a schema with:
```json
"provider": {
  "javaType": "com.example.ProviderType",
  "type": "string",
  "enum": ["system", "user", "automation"],
  "default": "user"
}
```

**Expected** (and what you get when the enum is generated from scratch):
```java
private ProviderType provider = ProviderType.fromValue("user");
```

**Actual** (when `ProviderType.class` already exists on the classpath):
```java
private ProviderType provider = null;
```

## Fix

Add a fallback branch in `DefaultRule.getDefaultValue()` that detects classpath enum types using reflection:

1. `isClasspathEnum()` — checks if a non-`JDefinedClass` type is an enum on the classpath via `Class.isEnum()`
2. `getDefaultEnumFromClasspath()` — finds the `fromValue` method via reflection, determines the backing type from its parameter, and generates the correct `fromValue(...)` expression

This handles string-backed, integer-backed, and number-backed enums correctly by inspecting the `fromValue` method's parameter type.

## Test

Added an integration test (`DefaultIT.classpathEnumPropertyHasCorrectDefaultValue`) that:

1. Defines a pre-compiled enum class (`ExistingClasspathEnum`) in the test sources — this class is on the classpath during code generation, triggering the `ClassAlreadyExistsException` path in `EnumRule`
2. Uses a schema that references this enum with `"javaType"` and declares `"default": "beta"`
3. Generates and compiles the schema, then verifies the field is initialised to `ExistingClasspathEnum.BETA` instead of `null`

All 596 tests pass. The only pre-existing failure (`DynamicPropertiesIT.shouldSetStringFieldJava7`) targets Java 7 on JDK 17 and is unrelated.

Fixes #926

### DIFF
--- a/jsonschema2pojo-integration-tests/src/test/java/org/jsonschema2pojo/integration/DefaultIT.java
+++ b/jsonschema2pojo-integration-tests/src/test/java/org/jsonschema2pojo/integration/DefaultIT.java
@@ -425,4 +425,28 @@ public class DefaultIT {
 
     }
 
+    /**
+     * Verifies that enum default values are correctly applied when using
+     * {@code existingJavaType} to reference an enum on the classpath (issue #926).
+     *
+     * @see <a href="https://github.com/joelittlejohn/jsonschema2pojo/issues/926">issue #926</a>
+     */
+    @Test
+    public void existingJavaTypeEnumPropertyHasCorrectDefaultValue() throws Exception {
+
+        ClassLoader resultsClassLoader = schemaRule.generateAndCompile(
+                "/schema/default/defaultWithExistingClasspathEnum.json", "com.example");
+
+        Class<?> generatedClass = resultsClassLoader.loadClass("com.example.DefaultWithExistingClasspathEnum");
+        Object instance = generatedClass.getDeclaredConstructor().newInstance();
+
+        Method getter = generatedClass.getMethod("getExistingEnumWithDefault");
+        Object defaultValue = getter.invoke(instance);
+
+        assertThat(defaultValue, is(notNullValue()));
+        assertThat(defaultValue, is(instanceOf(ExistingClasspathEnum.class)));
+        assertThat(defaultValue, is(equalTo(ExistingClasspathEnum.BETA)));
+
+    }
+
 }
--- a/jsonschema2pojo-integration-tests/src/test/resources/schema/default/defaultWithExistingClasspathEnum.json
+++ b/jsonschema2pojo-integration-tests/src/test/resources/schema/default/defaultWithExistingClasspathEnum.json
@@ -1,0 +1,11 @@
+{
+    "type" : "object",
+    "properties" : {
+        "existingEnumWithDefault" : {
+            "existingJavaType" : "org.jsonschema2pojo.integration.classpath.ExistingClasspathEnum",
+            "type" : "string",
+            "enum" : ["alpha", "beta", "gamma"],
+            "default" : "beta"
+        }
+    }
+}


joelittlejohn:
I haven't checked yet, but I'm slightly confused on this one. You should be using 'existingJavaType' to refer to something from the classpath. 'javaType' is used when you are generating a new type, but you want to specify the fully-qualified type name.  No?

manerow:
Thanks for the quick look @joelittlejohn!

You're right that `existingJavaType` is the intended way to reference a classpath type. However, the library already supports the case where `javaType` points to a type that happens to exist on the classpath: `EnumRule` catches `ClassAlreadyExistsException` and gracefully reuses the existing class. This works correctly for type resolution. The only gap is that `DefaultRule` doesn't recognize the resulting type as an enum (it checks `instanceof JDefinedClass`, but the reused class is a `JDirectClass`), so the default value is silently dropped to `null`.

This matters in multi-module projects where the same schemas are processed in different build contexts. The upstream module generates everything from scratch (`javaType` is correct there). A downstream module has its own schemas that `$ref` upstream definitions, so the upstream schema files must be present in the source directory for reference resolution. The plugin then processes all schemas in that directory, encountering enum types that already exist on the classpath from the upstream JAR. The downstream module doesn't control the upstream schemas and can't change `javaType` to `existingJavaType`.

This PR just completes what `ClassAlreadyExistsException` handling already does: it extends classpath-enum support to default values.

For the test, I used `javaType` because that's the real-world scenario. `existingJavaType` goes through `TypeRule` and bypasses enum generation entirely, so it wouldn't exercise this code path.

joelittlejohn:
Fair enough. Hyrum's Law in action I suppose!

I think this is a hangover from before we separated javaType and existingJavaType. I think we should fix this for existingJavaType because it gives you a way to refer to an existing enum and choose a default value, without having to use the javaType trick.

manerow:
Good point @joelittlejohn. The fix already covers `existingJavaType`. Both paths produce a `JClass` that reaches `DefaultRule.getDefaultValue()` the same way, and `isClasspathEnum()` handles both.

I've added an integration test (`existingJavaTypeEnumPropertyHasCorrectDefaultValue`) that uses `existingJavaType` with a default value and verifies it resolves correctly. All tests pass.

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/DefaultRule.java
```diff
@@ -255,6 +260,50 @@ private static JExpression getDefaultEnum(JType fieldType, String value) {
...
     }
 
+    /**
+     * Checks whether the given type is an enum that already exists on the
+     * classpath (i.e. it is not a {@link JDefinedClass} generated during this
+     * run, but a reference to a pre-compiled class).
+     *
+     * @see <a href="https://github.com/joelittlejohn/jsonschema2pojo/issues/926">issue #926</a>
+     */
+    private static boolean isClasspathEnum(JType fieldType) {
```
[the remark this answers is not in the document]
manerow: @unkish gentle bump on this thread 😄

jsonschema2pojo-integration-tests/src/test/java/org/jsonschema2pojo/integration/classpath/ExistingClasspathEnum.java
```diff
@@ -0,0 +1,69 @@
...
+import com.fasterxml.jackson.annotation.JsonValue;
+
+/**
+ * A pre-compiled enum that simulates the pattern generated by jsonschema2pojo.
+ * Used by integration tests to verify that default values are correctly applied
+ * when an enum type already exists on the classpath.
+ *
+ * @see <a href="https://github.com/joelittlejohn/jsonschema2pojo/issues/926">issue #926</a>
+ */
+public enum ExistingClasspathEnum {
```
unkish, line 32: I would rather move that file away from `org.jsonschema2pojo.integration`, maybe into `com.example.classpath` keeping former package name for integration tests.

jsonschema2pojo-integration-tests/src/test/java/org/jsonschema2pojo/integration/DefaultIT.java
```diff
@@ -400,4 +401,52 @@ public void uniqueArrayPropertyWithoutDefaultIsEmptySet() throws NoSuchMethodExc
...
     }
 
+    /**
+     * Verifies that enum default values are correctly applied when the enum
+     * type already exists on the classpath (issue #926).
+     *
+     * @see <a href="https://github.com/joelittlejohn/jsonschema2pojo/issues/926">issue #926</a>
+     */
+    @Test
+    public void classpathEnumPropertyHasCorrectDefaultValue() throws Exception {
```
unkish, line 411: I'm not quite sure as to what to think of this test - this will just cause `EnumRule` to return reference to existing class which should be equivalent to test that has `existingJavaType`.
It does no harm here, but should it be a part of `DefaultIT` or `EnumIT`, maybe someone else has also an opinion on that.

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/DefaultRule.java
```diff
@@ -157,6 +158,10 @@ static JExpression getDefaultValue(JType fieldType, String value) {
 
             return getDefaultEnum(fieldType, value);
 
+        } else if (fieldType instanceof JClass && isClasspathEnum(fieldType)) {
```
unkish, line 161: Nit: `instanceOf` operator could be used here, eliminating the need for downstream casting
```java
        } else if (fieldType instanceof JClass jClass && isClasspathEnum(jClass)) {

            return getDefaultEnumFromClasspath(jClass, value);
```

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/DefaultRule.java
```diff
@@ -255,6 +260,50 @@ private static JExpression getDefaultEnum(JType fieldType, String value) {
...
+     *
+     * @see <a href="https://github.com/joelittlejohn/jsonschema2pojo/issues/926">issue #926</a>
+     */
+    private static JExpression getDefaultEnumFromClasspath(JClass fieldType, String value) {
+        try {
+            Class<?> enumClass = Thread.currentThread().getContextClassLoader()
+                    .loadClass(fieldType.fullName());
+            for (Method m : enumClass.getMethods()) {
+                if ("fromValue".equals(m.getName()) && m.getParameterCount() == 1) {
+                    JType backingType = fieldType.owner().ref(m.getParameterTypes()[0]).unboxify();
```
unkish, line 295: `unboxify` is redundant as
* it will be done downstream in `getDefaultValue`
* `.ref` would throw an `IllegalArgumentException` for primitive type

jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/DefaultRule.java
```diff
@@ -255,6 +260,50 @@ private static JExpression getDefaultEnum(JType fieldType, String value) {
...
+     * parameter type so the correct literal can be generated.
+     *
+     * @see <a href="https://github.com/joelittlejohn/jsonschema2pojo/issues/926">issue #926</a>
+     */
+    private static JExpression getDefaultEnumFromClasspath(JClass fieldType, String value) {
+        try {
+            Class<?> enumClass = Thread.currentThread().getContextClassLoader()
+                    .loadClass(fieldType.fullName());
+            for (Method m : enumClass.getMethods()) {
+                if ("fromValue".equals(m.getName()) && m.getParameterCount() == 1) {
```
unkish, line 294: Nit: to me both the method name and the PR/issue title are a bit misleading as although sounding generalized in reality there seems to be an assumption that "enum" is an enum generated by js2p.

unkish, line 294: Should:
* return type also be checked ?
* static modifier be checked ?

jsonschema2pojo-integration-tests/src/test/resources/schema/default/defaultWithExistingClasspathEnum.json
```diff
@@ -0,0 +1,11 @@
+{
+    "type" : "object",
+    "properties" : {
+        "existingEnumWithDefault" : {
```
unkish, line 4: I wonder if it'd be possible to have a test with a "non-standard" enum value type eg. one that is backed by double or bigdecimal

PR #1781, opened 2026-02-14
Apply '@Valid' to properties with "existingJavaType"

unkish:


unkish:
Idea of applying `@Valid` also to properties with `"existingJavaType"` was in the air during https://github.com/joelittlejohn/jsonschema2pojo/pull/1778 development.

This little POC is for the case if a non elegant approach could be considered for properties having `"existingJavaType"`.

joelittlejohn:
Looks good to me. I think we should merge this in 1.3.4 to avoid losing @Valid on fields that use existingJavaType.

jsonschema2pojo-integration-tests/src/test/java/org/jsonschema2pojo/integration/config/IncludeJsr303AnnotationsIT.java
```diff
@@ -700,6 +704,37 @@ public void jsr303ValidAnnotationNotAppliedToScalarTypes() throws Exception {
                 source, not(containsString("@Valid")));
     }
 
+    @Test
+    void jsr303AnnotationsValidatedForExistingJavaType() throws ReflectiveOperationException {
```
joelittlejohn: I understand this is no longer necessary/recommended with JUnit 5, but since every other test is public I think we best keep new tests public too (until we update all of them).

jsonschema2pojo-integration-tests/src/test/resources/schema/jsr303/minAndMaxItemsExistingType.json
```diff
@@ -5,7 +5,7 @@
             "type" : "array",
             "minItems" : 2,
             "maxItems" : 4,
-            "existingJavaType" : "java.util.LinkedList<String>"
+            "existingJavaType" : "java.util.LinkedList<java.lang.String>"
```
joelittlejohn: Is this needed for java.lang types btw? Other example schemas suggest it is not 🤔

jsonschema2pojo-integration-tests/src/test/resources/schema/jsr303/existingJavaTypeProperties.json
```diff
@@ -0,0 +1,46 @@
...
+    },
+    "rawOptional": {
+      "type": "object",
+      "existingJavaType": "java.util.Optional"
+    },
+    "rawMap": {
+      "type": "object",
+      "existingJavaType": "java.util.Map"
+    }
+  }
```
joelittlejohn, line 67: Could you add a couple of tests here for `"existingJavaType": "long"` and `"existingJavaType": "String[]"`?

jsonschema2pojo-integration-tests/src/test/resources/schema/jsr303/minAndMaxItemsExistingType.json
unkish: It is not required, however without fully qualified class name, generated code will have `LinkedList<@Valid String>`.
I will remove it (to reduce the noise).

jsonschema2pojo-integration-tests/src/test/java/org/jsonschema2pojo/integration/config/IncludeJsr303AnnotationsIT.java
unkish: Sure, I will change it

jsonschema2pojo-integration-tests/src/test/resources/schema/jsr303/existingJavaTypeProperties.json
unkish: Yes.
That is a very good remark. Thank You!

unkish: I have decided to leave out `"existingJavaType": "String[]"` since as it stands currently (checked also with `1.2.2`) this would generate incorrect code:
```java
    @JsonProperty("stringArray")
    @Valid
    private String stringArray;
```

This can be fixed by adjusting `TypeUtil::resolveType` to pass correct `arrayCount` value to `buildClass`:
```java
            FieldDeclaration fieldDeclaration = (FieldDeclaration) JavaParser.parseBodyDeclaration(typeDefinition + " foo;");
            final var fieldDeclarationType = (ReferenceType) fieldDeclaration.getType();
            ClassOrInterfaceType c = (ClassOrInterfaceType) fieldDeclarationType.getType();

            return buildClass(_package, c, fieldDeclarationType.getArrayCount());
```

However this will only work if both `includeHashcodeAndEquals` and `includeToString` will be set to `false` as otherwise guard blocks in `ObjectRule::addToString`, `ObjectRule::addHashCode`, `ObjectRule::addEquals` would trigger and abort code generation.
```java
if (!fieldVar.type().elementType().isPrimitive()) {
    throw new UnsupportedOperationException("Only primitive arrays are supported");
}
```

Please let me know what are your thoughts on that.

unkish:
Unrelated to this commit:
`org.jsonschema2pojo.cli.Arguments` https://github.com/joelittlejohn/jsonschema2pojo/blob/1c6b7664e453c62265c5e60a630f52f1fd47452a/jsonschema2pojo-cli/src/main/java/org/jsonschema2pojo/cli/Arguments.java#L197-L198 has default value `1.6` - perhaps that should be changed to `8`

jsonschema2pojo-integration-tests/src/test/resources/schema/jsr303/existingJavaTypeProperties.json
joelittlejohn: Can you think for a reason to keep this guard and the UnsupportedOperationException? It doesn't seem necessary to me and I think we should remove it. The code uses Arrays.hashCode/Arrays.equals/Arrays.toString which should work find for Object arrays, right?

joelittlejohn: Best to add a separate PR for that change though, because I think it's worth including in the release notes.

joelittlejohn: We can merge this with a failing test if that's easiest. Feel free to organise it however you find convenient.

unkish: > Can you think for a reason to keep this guard and the UnsupportedOperationException? It doesn't seem necessary to me and I think we should remove it. The code uses Arrays.hashCode/Arrays.equals/Arrays.toString which should work find for Object arrays, right?

For single dimensional arrays they would work.
Multidimensional arrays would require `deepEquals` / `deepToString`, `deepHashCode`.
It also looks like `@Valid` has no effect on multidimensional arrays, however that is rather implementation detail of validator

unkish:
It looks like there's one cornercase.
```json
{
  "type": "object",
  "javaType": "com.example.ExistingJavaTypeProperties",
  "properties": {
    "itemOfExistingClassFqcn": {
      "type": "object",
      "existingJavaType": "com.example.ClassWithSizeAnnotation"
    },
    "itemOfExistingClass": {
      "type": "object",
      "existingJavaType": "ClassWithSizeAnnotation"
    }
  }
}
```
Generates a code that is compiling (note the placement of `@Valid` on `itemOfExistingClassFqcn` field)
```java
    @JsonProperty("itemOfExistingClassFqcn")
    private com.example.@Valid ClassWithSizeAnnotation itemOfExistingClassFqcn;
    @JsonProperty("itemOfExistingClass")
    private @Valid ClassWithSizeAnnotation itemOfExistingClass;
```
but fails validation:
```java
        instance = createInstanceWithPropertyValue(clazz, "itemOfExistingClassFqcn", new ClassWithSizeAnnotation());
        assertNumberOfConstraintViolationsOn(instance, is(1));
```


Prefix is added in `JAnnotatedClass`
```java
        if (!isImported(f, rawType) && !prefix.isEmpty()) {
            f.p(prefix);
        }
```

Since generated type and `existingJavaType` both belong to same package, then `!isImported(f, rawType)` is `true` and since `existingJavaType` had fqcn then `!prefix.isEmpty()` also resolves to `true`


I have not yet figured out how to fix this.

joelittlejohn:
Is the problem just aesthetic @unkish? i.e. `com.example` is printed for ClassWithSizeAnnotation even though it is not needed. Or is the validation not working?

jsonschema2pojo-integration-tests/src/test/resources/schema/jsr303/existingJavaTypeProperties.json
joelittlejohn: Let's ignore multi-dimensional arrays for now, at leat we will support unidimensional arrays that contain objects (which we did not in the past).

unkish:
> Is the problem just aesthetic @unkish? i.e. `com.example` is printed for ClassWithSizeAnnotation even though it is not needed. Or is the validation not working?

Validation is not working as `@Valid` is absent on field when queried through reflection.
Issue can be observed if schema file would contain a mix of non fully qualified class names with fullyqualified class names (`JCodeModel::ref(Class)` produce fully qualified class name references)

unkish:
@joelittlejohn I haven't been able to think of any better solution than either
* declaring that a mix of fqcn and non-fqcn would not be supported at least for `@Valid` annotation
or
* [roll back a bit changes to `TypeRule`](https://github.com/unkish/jsonschema2pojo/tree/issues/apply_valid_to_existingJavaType_properties.v2) and let `ValidRule` either annotate the field itself or add annotation inside container😞

joelittlejohn:
@unkish Sorry, I'm being dense here. Could you add a failing test?

I don't see why this is a problem:

```java
    @JsonProperty("itemOfExistingClassFqcn")
    private com.example.@Valid ClassWithSizeAnnotation itemOfExistingClassFqcn;
    @JsonProperty("itemOfExistingClass")
    private @Valid ClassWithSizeAnnotation itemOfExistingClass;
```

and I don't understand why it leads to this:

> Validation is not working as @Valid is absent on field when queried through reflection.

The annotation placement on itemOfExistingClassFqcn looks correct to me.

Yes, it's using a fully qualified name for the class even though it does not have to, but the annotation is in the right place, isn't it?

If you add a failing test here I will see if I can think of a way to fix it.

joelittlejohn:
@unkish Never mind, I have created a failing test now based on the schema you provided. Now to fixing...

unkish:
My understanding is that in case of `private com.example.@Valid ClassWithSizeAnnotation itemOfExistingClassFqcn;` hibernate skips validation as `@Valid` is a type and not a filed annotation.
As annotations are applied in `TypeRule` then either `JAnnotatedClass::generate` would somehow need to know/track where it is being invoked to print annotations correctly, or `JAnnotatedClass` would need to be applied only on container's types.

unkish:
Two options on my table atm:
1. [remove potential "type-use context" before creating field](https://github.com/unkish/jsonschema2pojo/commit/c262658c5cbb42675e7049fd6707fcf515c6f10d)
2. [apply `ValidRule` on `JFieldVar` instead of `JType`](https://github.com/unkish/jsonschema2pojo/commit/d538bc1e58af141bbea41a9a1d3506058fda74a3)

@joelittlejohn when you have some spare time, I would appreciate if you could have a look and share your thoughts on proposed solutions or any alternative approaches that might solve the issue.

